PR #35, opened 2019-04-04
Made simple agents suitable for challenge submissions.

mathfac:
## Motivation and Context
Made simple agents suitable for challenge submissions:
- switch from running benchmark to challenge
- get rid of config for simple agents as only one parameter needed
- removed redundant act method

## How Has This Been Tested

`pytest`
`black . --line-length 79;mypy . --ignore-missing-imports;pycodestyle;flake8`

## Types of changes

- Bug fix (non-breaking change which fixes an issue)/improvement

## Checklist


baselines/agents/simple_agents.py
```diff
@@ -58,9 +53,8 @@ def act(self, observations):
 
 
 class RandomForwardAgent(RandomAgent):
-    def __init__(self, config):
-        super(RandomForwardAgent, self).__init__(config)
-        self.dist_threshold_to_stop = config.TASK.SUCCESS_DISTANCE
+    def __init__(self, success_distance):
+        super(RandomForwardAgent, self).__init__(success_distance)
```
abhiskk: You can do `super().__init__(success_distance)` (Python3 syntax)

baselines/agents/simple_agents.py
```diff
@@ -81,8 +75,8 @@ def act(self, observations):
 
 
 class GoalFollower(RandomAgent):
-    def __init__(self, config):
-        super(GoalFollower, self).__init__(config)
+    def __init__(self, success_distance):
+        super(GoalFollower, self).__init__(success_distance)
```
abhiskk: `super().__init__(success_distance)`

baselines/agents/simple_agents.py
```diff
@@ -135,20 +129,15 @@ def get_agent_cls(agent_class_name):
 
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument(
-        "--task-config", type=str, default="tasks/pointnav.yaml"
-    )
+    parser.add_argument("--success_distance", type=float, default=0.2)
```
abhiskk: "--success_distance" -> `--success-distance`. Note that this does not require changes in python code, python argpase automatically maps `--success-distance` to `args.success_distance`.

baselines/agents/simple_agents.py
```diff
@@ -135,20 +129,15 @@ def get_agent_cls(agent_class_name):
...
     agent = get_agent_cls(args.agent_class)(
-        habitat.get_config(args.task_config)
+        success_distance=args.success_distance
     )
-    benchmark = habitat.Benchmark(args.task_config)
-    metrics = benchmark.evaluate(agent)
-
-    for k, v in metrics.items():
-        habitat.logger.info("{}: {:.3f}".format(k, v))
+    challenge = habitat.Challenge()
```
abhiskk: I suggest we retain the old `benchmark` usage here instead of `challenge`. This code is orthogonal to challenge codebase I think, also if someone is testing locally `benchmark` fits in better semantically compared to `challenge`

baselines/agents/simple_agents.py
```diff
@@ -135,20 +129,15 @@ def get_agent_cls(agent_class_name):
 
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument(
-        "--task-config", type=str, default="tasks/pointnav.yaml"
-    )
+    parser.add_argument("--success_distance", type=float, default=0.2)
     parser.add_argument("--agent_class", type=str, default="GoalFollower")
```
abhiskk: Same as above `--agent-class`

baselines/agents/simple_agents.py
mathfac: @abhiskk can you give more context why `success-distance` is better than `success_distance`?

abhiskk: It is not necessarily better, but generally I have seen `success-distance` type cmd arguments being used in python, I looked around though and I have found both types of examples actually so don't have a preference here. We should be consistent, the only other place I think command line args are being used in the project are for PPO: https://github.com/facebookresearch/habitat-api/blob/master/baselines/rl/ppo/utils.py#L275  and they are in the format of `--success-distance`. I am happy with either usage, but we should be consistent.

erikwijmans: I tend to think that `--success-distance` looks cleaner, although it makes it a little harder to search for the definition of the command line arg for a certain variable.

mathfac:
Addressed the comments.

erikwijmans:
Looks like a `.swp` file got added with this PR :(

PR #39, opened 2019-04-05
Slambased agents

ducha-aiki:
Added ORBSLAM2-based agents: Blind, Classic RGBD, Classic RGB-monoD agents from "Benchmarking 
Classic and Learned Navigation in Complex 3D Environments" https://github.com/ducha-aiki/navigation-benchmark


mathfac:
Dmytro, thank you for all the hard work and follow ups. It should be good to go!

haoliangjiang:
Thank you for your work @ducha-aiki . Could you please share your config file for running the code? I checked the "pointnav_rgbd.yaml" in the slam_agnets.py. I think it misses some important keys.

PR #44, opened 2019-04-05
Greedy follower

danielgordon10:
## Motivation and Context

Adds the greedy follower class for querying the next greedy action based on geodesic distance. Not sure about file placement, but this seemed like a reasonable spot for it.

## How Has This Been Tested

Tested via ICCV code.

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #46, opened 2019-04-06
Compatibility with habitat-sim#6

erikwijmans:
## Motivation and Context

Updates HAPI to be compatible with https://github.com/facebookresearch/habitat-sim/pull/6

This also gets rid of the `SIM_NAME_TO_ACTION` and `SIM_ACTION_TO_NAME` maps as HSIM will now happily take integers as the keys into the agents action space, because python.

## How Has This Been Tested

Currently just with `pytest`.  Trained the PPO baseline as a burn-in test to make sure I didn't mess-up any memory management or scene swapping logic.

## Types of changes


erikwijmans:
@mathfac @abhiskk Burn in testing looks good (shuffled list of episodes so that the scene is swapped every 10 episodes, so there have been lots of swaps), ready for review.

test/test_trajectory_sim.py
```diff
@@ -50,7 +50,7 @@ def test_sim_trajectory():
                     np.array(
                         test_trajectory["rotations"][i], dtype=np.float32
                     ),
-                    state.rotation,
+                    np.array([*state.rotation.imag, state.rotation.real]),
```
abhiskk: Maybe a typo in `imag`?

test/test_habitat_env.py
```diff
@@ -255,14 +251,13 @@ def test_rl_env():
...
-    observation, reward, done, info = env.step(
-        SIM_NAME_TO_ACTION[SimulatorActions.STOP.value]
-    )
+    observation, reward, done, info = env.step(SimulatorActions.STOP.value)
     assert done is True, "done should be true after STOP action"
 
     env.close()
 
 
+@pytest.mark.skip
```
abhiskk: Why has this skip been added, is it to account for the fact that this test is slow or is it because this PR Is breaking this test?

habitat/sims/habitat_simulator.py
```diff
@@ -208,18 +197,23 @@ def create_sim_config(
         agent_config.sensor_specifications = sensor_specifications
         agent_config.action_space = {
             SimulatorActions.LEFT.value: habitat_sim.ActionSpec(
-                "lookLeft", {"amount": self.config.TURN_ANGLE}
+                "turn_left",
+                habitat_sim.ActuationSpec(amount=self.config.TURN_ANGLE),
```
abhiskk: What exactly is the use for `ActuationSpec`, is it to factor in noise?

habitat/sims/habitat_simulator.py
```diff
@@ -321,41 +316,9 @@ def action_space_shortest_path(
             the source. For the last item in the returned list the action
             will be None.
         """
-        assert agent_id == 0, "No support of multi agent in {} yet.".format(
-            self.__class__.__name__
+        raise NotImplementedError(
```
abhiskk: What about leaving this implementation in and having two versions of shortest path:

1. Greedy shortest path which runs fast but not necessarily gives you the exact shortest path. 
2. Action space shortest path which gives you an accurate shortest path.

test/test_trajectory_sim.py
erikwijmans: No typo.  It's short for `imaginary` and returns the coefficients of the `i, j, k` components (the imaginary part) of the quaternion.

test/test_habitat_env.py
erikwijmans: This PR breaks this test.  The action space shortest path is both slow and unreliable (it gives an accurate path when it finds one, but it is very unreliable in its ability to find a path).  The PR on the sim side breaks that code and due to the unreliability of the action space shortest path code, I don't think it's worth the effort to un-break it.  The greedy follower form #44 is a better option.  

I have a version of that written in c++ with a nicer interface since it is closer to the simulator that will replace the action space shortest path code, so this test will come back.  The greedy follower on the sim side just relies in habitat-sim#6, and thus can't be landed yet.

habitat/sims/habitat_simulator.py
erikwijmans: Current the ActuationSpec just specifies how much the action moves the agent.  Be it the number of degrees to turn or the number of meters to move forward.  I made it a separate class that way actions that have more things to configure (noisy actions) will fit into the same paradigm, but with their own subclass of `ActuationSpec`.

habitat/sims/habitat_simulator.py
erikwijmans: See comment on `@pytest.mark.skip` for context.

test/test_baseline_agents.py
```diff
@@ -4,14 +4,25 @@
...
+except ImportError:
+    has_torch = False
+
+if has_torch:
+    from baselines.agents import ppo_agents
 
 CFG_TEST = "test/habitat_all_sensors_test.yaml"
 
 
+@pytest.mark.skipif(not has_torch, reason="Test needs torch")
```
mathfac: Thank you for adding skip condition.

habitat/sims/habitat_simulator.py
```diff
@@ -34,10 +34,10 @@ def check_sim_obs(obs, sensor):
...
 
 class SimulatorActions(Enum):
-    LEFT = "look_left"
-    RIGHT = "look_right"
-    FORWARD = "move_forward"
-    STOP = "stop"
+    FORWARD = 0
+    LEFT = 1
+    RIGHT = 2
+    STOP = 3
```
mathfac: Do we want to use this moment and make STOP=0?

erikwijmans: That should stay in #53 IMO.  Clearer commit history

test/test_habitat_env.py
abhiskk: Great, can you add a TODO here so that we don't forget to bring this test back in.

PR #49, opened 2019-04-08
Dataset utility functions

danielgordon10:
## Motivation and Context

New util methods for creating different dataset splits for later use by vector_env.

## How Has This Been Tested

Tested extensively during ICCV process.

## Types of changes

## Checklist


habitat/core/dataset.py
```diff
@@ -101,3 +103,115 @@ def default(self, object):
...
+                    new_dataset.episodes.append(
+                        self.episodes[rand_items[ep_ind]]
+                    )
+                    ep_ind += 1
+                new_episodes.extend(new_dataset.episodes)
+        if remove_unused_episodes:
+            self.episodes = new_episodes
+        return new_datasets
+
+    def get_uneven_splits(self, num_splits):
```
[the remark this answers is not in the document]
danielgordon10: I use it for making sure that all episodes from an original test set are in the splits but that none are duplicated.

habitat/core/dataset.py
```diff
@@ -67,7 +69,7 @@ def scene_ids(self) -> List[str]:
         Returns:
             unique scene ids present in the dataset
         """
-        return list({episode.scene_id for episode in self.episodes})
+        return sorted(list({episode.scene_id for episode in self.episodes}))
```
abhiskk: Just for completeness I think we should add a `set` constructor also here to ensure that the `scene_id` are unique.

habitat/core/dataset.py
```diff
@@ -101,3 +103,124 @@ def default(self, object):
...
+                new_datasets.append(new_dataset)
+            new_episodes = self.episodes
+        else:
+            rand_items = np.random.choice(
+                len(self.episodes),
+                num_splits * max_episodes_per_split,
+                replace=False,
+            )
+            # Sort it so episode scene doesn't switch as often.
+            rand_items.sort()
```
abhiskk: Maybe this responsibility should be delegated to the user for the dataset rather than us sorting and returning the data.

habitat/core/dataset.py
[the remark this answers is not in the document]
abhiskk: I think we need a better name for the method here.

habitat/core/dataset.py
danielgordon10: The curly brackets is a set construction.

habitat/core/dataset.py
danielgordon10: I'll add a parameter for collating from the same scene_id.

habitat/core/dataset.py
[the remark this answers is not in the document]
danielgordon10: What don't you like about the name? Can you suggest an alternative?

habitat/core/dataset.py
```diff
@@ -101,3 +105,140 @@ def default(self, object):
...
+            len(self.episodes) >= num_splits
+        ), "Not enough episodes to create this many splits."
+
+        new_datasets = []
+        if max_episodes_per_split is None:
+            max_episodes_per_split = len(self.episodes) // num_splits
+        max_episodes_per_split = min(
+            max_episodes_per_split, (len(self.episodes) // num_splits)
+        )
+        if num_splits * max_episodes_per_split == len(self.episodes):
```
mathfac: Why do we have this special case? Does it implements same logic as non exact size match regarding randomization, collating scenes?

habitat/core/dataset.py
```diff
@@ -101,3 +105,140 @@ def default(self, object):
...
+        duplicated.
+        Args:
+            num_splits: The number of splits to create.
+        Returns:
+            A list of new datasets, each with their own subset of episodes.
+        """
+        assert (
+            len(self.episodes) >= num_splits
+        ), "Not enough episodes to create this many splits."
+        num_splits = min(num_splits, len(self.episodes))
```
mathfac: Based on assert above `num_splits == num_splits`.

habitat/core/dataset.py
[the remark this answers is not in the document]
mathfac: @danielgordon10 this method can be combined with `get_splits` and add new param as `equal_splits` to control that behavior.

habitat/core/dataset.py
danielgordon10: It is faster to use slices rather than make new lists, but the else branch can work for exact splits as well. I can remove it if you want.

habitat/core/dataset.py
danielgordon10: Yep. I was indecisive about which behavior I wanted. I can take the min thing out.

PR #50, opened 2019-04-08
Vector env utility functions

danielgordon10:
## Motivation and Context

Extra vector env functions that I found useful for debugging and speeding up eval.

## How Has This Been Tested

Tested extensively during ICCV experiments.

## Checklist


abhiskk:
I think this is not rebased on master, @danielgordon10 can you rebase this? Thanks.

danielgordon10:
It should be now.

mathfac:
@danielgordon10 I like the idea to make VecEnv more general and unlock option to call any function. Can you add tests to make sure we have this functionality working when other code get updated?

mathfac:
Maybe, we can take old test that use `step/reset` actions and run through `call`?

danielgordon10:
I will write some tests for this. It shouldn't be too hard. I'll just add some to the habitat_env tests.

danielgordon10:
I think it's nice to keep the same `step/reset` functions as they align with other environments like gym and even the single Env (not VectorEnv).

PR #51, opened 2019-04-08
Greedy Shortest Path Follower

danielgordon10:
## Motivation and Context

Adds the greedy follower class for querying the next greedy action based on geodesic distance. Not sure about file placement, but this seemed like a reasonable spot for it.

## How Has This Been Tested

Tested via ICCV code.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,7 +76,7 @@ target/
 .ipynb_checkpoints/
 
 # exclude data from source control by default
-/data/
+*/data/
 
 # Mac OS-specific storage files
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -76,7 +76,7 @@ target/
 .ipynb_checkpoints/
 
 # exclude data from source control by default
-*/data/
+data
 
 # Mac OS-specific storage files
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+examples/images
 
 # Translations
 *.mo
--- a/examples/shortest_path_follower_example.py
+++ b/examples/shortest_path_follower_example.py
@@ -6,7 +6,6 @@
 
 import os
 import shutil
-import sys
 
 import imageio
 
--- a/examples/shortest_path_follower_example.py
+++ b/examples/shortest_path_follower_example.py
@@ -18,22 +18,23 @@ if not os.path.exists(IMAGE_DIR):
     os.makedirs(IMAGE_DIR)
 
 
-def shortest_path_example():
-    if os.path.exists(os.path.join(IMAGE_DIR, "shortest_path_example")):
-        shutil.rmtree(os.path.join(IMAGE_DIR, "shortest_path_example"))
+def shortest_path_example(mode):
     config = habitat.get_config(config_file="tasks/pointnav.yaml")
     env = habitat.Env(config=config)
     goal_radius = env.episodes[0].goals[0].radius
     if goal_radius is None:
         goal_radius = config.SIMULATOR.FORWARD_STEP_SIZE
     follower = ShortestPathFollower(env.sim, goal_radius, False)
+    follower.mode = mode
 
     print("Environment creation successful")
     for episode in range(3):
         observations = env.reset()
         dirname = os.path.join(
-            IMAGE_DIR, "shortest_path_example", "%02d" % episode
+            IMAGE_DIR, "shortest_path_example", mode, "%02d" % episode
         )
+        if os.path.exists(dirname):
+            shutil.rmtree(dirname)
         os.makedirs(dirname)
         print("Agent stepping around inside environment.")
         count_steps = 0
@@ -48,5 +49,10 @@ def shortest_path_example():
         print("Episode finished after {} steps.".format(count_steps))
 
 
+def main():
+    shortest_path_example("geodesic_path")
+    shortest_path_example("greedy")
+
+
 if __name__ == "__main__":
-    shortest_path_example()
+    main()
--- a/examples/shortest_path_follower_example.py
+++ b/examples/shortest_path_follower_example.py
@@ -4,16 +4,24 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
+import shutil
 import sys
 
-import cv2
+import imageio
 
 import habitat
 from habitat.sims.habitat_simulator import SIM_NAME_TO_ACTION
 from habitat.tasks.nav.shortest_path_follower import ShortestPathFollower
 
+IMAGE_DIR = os.path.join("examples", "images")
+if not os.path.exists(IMAGE_DIR):
+    os.makedirs(IMAGE_DIR)
+
 
 def shortest_path_example():
+    if os.path.exists(os.path.join(IMAGE_DIR, "shortest_path_example")):
+        shutil.rmtree(os.path.join(IMAGE_DIR, "shortest_path_example"))
     config = habitat.get_config(config_file="tasks/pointnav.yaml")
     env = habitat.Env(config=config)
     goal_radius = env.episodes[0].goals[0].radius
@@ -24,6 +32,10 @@ def shortest_path_example():
     print("Environment creation successful")
     for episode in range(3):
         observations = env.reset()
+        dirname = os.path.join(
+            IMAGE_DIR, "shortest_path_example", "%02d" % episode
+        )
+        os.makedirs(dirname)
         print("Agent stepping around inside environment.")
         count_steps = 0
         while not env.episode_over:
@@ -32,18 +44,8 @@ def shortest_path_example():
             )
             observations = env.step(SIM_NAME_TO_ACTION[best_action.value])
             count_steps += 1
-            if "pytest" not in sys.modules:
-                im = observations["rgb"][:, :, ::-1].copy()
-                cv2.putText(
-                    im,
-                    "Action: " + best_action.value,
-                    (0, 20),
-                    cv2.FONT_HERSHEY_SIMPLEX,
-                    0.5,
-                    (0, 0, 255),
-                )
-                cv2.imshow("im", im)
-                cv2.waitKey(1)
+            im = observations["rgb"]
+            imageio.imsave(os.path.join(dirname, "%03d.jpg" % count_steps), im)
         print("Episode finished after {} steps.".format(count_steps))
 
 
--- a/examples/shortest_path_follower_example.py
+++ b/examples/shortest_path_follower_example.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import cv2
+
+import habitat
+from habitat.sims.habitat_simulator import SIM_NAME_TO_ACTION
+from habitat.tasks.nav.shortest_path_follower import ShortestPathFollower
+
+
+def shortest_path_example():
+    config = habitat.get_config(config_file="tasks/pointnav.yaml")
+    env = habitat.Env(config=config)
+    goal_radius = env.episodes[0].goals[0].radius
+    if goal_radius is None:
+        goal_radius = config.SIMULATOR.FORWARD_STEP_SIZE
+    follower = ShortestPathFollower(env.sim, goal_radius, False)
+
+    print("Environment creation successful")
+    for episode in range(10):
+        observations = env.reset()
+        print("Agent stepping around inside environment.")
+        count_steps = 0
+        while not env.episode_over:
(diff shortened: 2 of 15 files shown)


mathfac:
@danielgordon10 review comments are on old PR https://github.com/facebookresearch/habitat-api/pull/44#pullrequestreview-223969037.

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
...
+
+        return max_grad_dir
+
+    @property
+    def mode(self):
+        return self._mode
+
+    @mode.setter
+    def mode(self, new_mode):
+        assert new_mode in {"spath", "grad"}
```
mathfac: That would be great to add more description what "spath", "grad" stands here.

habitat/sims/habitat_simulator.py
```diff
@@ -451,3 +454,18 @@ def distance_to_closest_obstacle(self, position, max_search_radius=2.0):
...
+
+    @property
+    def UP(self):
+        return np.array([0.0, 1.0, 0.0])
+
+    @property
+    def FORWARD(self):
+        return -np.array([0.0, 0.0, 1.0])
+
+    def straight_spath_points(self, position_a, position_b):
```
mathfac: Why we can't use geodesic distance based shortest path here, without creating new method?

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
+from typing import Union
```
mathfac: Regarding `greedy_follower.py` name, it can be misleading as it doesn't state shortest path anywhere and we have agents that are followers (Goal Follower) as well., maybe something like `shortest_path_follower.py`

habitat/sims/habitat_simulator.py
```diff
@@ -451,3 +454,18 @@ def distance_to_closest_obstacle(self, position, max_search_radius=2.0):
         return self._sim.pathfinder.distance_to_closest_obstacle(
             position, max_search_radius
         )
+
+    @property
+    def UP(self):
```
mathfac: Let's add this to Simulator API and name it with lower case and name it like up_vector, forward_vector. Otherwise can be misleading with actions.

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                return self._get_return_value(best_turn)
+
+    def _reset_agent_state(self, state: habitat_sim.AgentState) -> None:
+        self._sim.set_agent_state(
+            state.position, state.rotation, reset_sensors=False
+        )
+
+    @staticmethod
+    def _angle_between_quats(q1: np.quaternion, q2: np.quaternion) -> float:
```
mathfac: if this functions `_angle_between_quats ` and `_quat_from_two_vectors ` is general enough, it can be used to utils where we have angle utils.

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
...
+        dq = quaternion.as_float_array(q1_inv * q2)
+
+        return 2 * np.arctan2(np.linalg.norm(dq[1:]), np.abs(dq[0]))
+
+    @staticmethod
+    def _quat_from_two_vectors(v0: np.array, v1: np.array) -> np.quaternion:
+        v0 = v0 / np.linalg.norm(v0)
+        v1 = v1 / np.linalg.norm(v1)
+        c = v0.dot(v1)
+        if c < (-1 + 1e-8):
```
mathfac: That would be great to name magic constants like `1e-8`

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+            # NB: Add a little offset as things get weird if
+            # points[1] - points[0] is anti-parallel with forward
+            if len(points) < 2:
+                return None
+            max_grad_dir = self._quat_from_two_vectors(
+                self._sim.FORWARD,
+                points[1]
+                - points[0]
+                + 1e-3 * np.cross(self._sim.UP, self._sim.FORWARD),
```
mathfac: 1e-3 here.

habitat/tasks/nav/greedy_follower.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+        else:
+            current_rotation = self._sim.get_agent_state().rotation
+            current_dist = self._geo_dist(goal_pos)
+
+            best_delta = -1e10
```
mathfac: If we can change naming to give more meaning to delta variable, that can help to maintain this code.

habitat/tasks/nav/greedy_follower.py
danielgordon10: Done.

habitat/tasks/nav/greedy_follower.py
danielgordon10: Done

habitat/sims/habitat_simulator.py
danielgordon10: Where do you want these moved to? Do you mean move them to habitat-sim somewhere? Or am I missing what you're saying?

habitat/tasks/nav/greedy_follower.py
danielgordon10: Done

habitat/tasks/nav/greedy_follower.py
danielgordon10: I'm not sure which names you don't like.

habitat/tasks/nav/greedy_follower.py
danielgordon10: Done

habitat/tasks/nav/greedy_follower.py
danielgordon10: Done

habitat/sims/habitat_simulator.py
danielgordon10: If you're asking why can't we do greedy search of best action using only geodesic distance/why do we need to call pathfinder to get path points the answer is that's slower, but also an option that the follower falls back on if it can't find this method.

danielgordon10:
@erikwijmans should look at this change as well

habitat/sims/habitat_simulator.py
mathfac: @danielgordon10 I thought this method `straight_spath_points ` returns points, not actions.

danielgordon10: It does return points. Then the shortest_path_finder uses those points to find actions.

.gitignore
```diff
@@ -76,7 +76,7 @@ target/
 .ipynb_checkpoints/
 
 # exclude data from source control by default
-/data/
+data
```
mathfac: That is excluding all code that have /data/ in path: for example habitat/internal/data, test/data. If baselines will have data preparation code and put it in data folder inside of baselines it will be ignored. It looks too general to me and we shouldn't spread data around code base if it's not checked in or not in root data folder.

examples/shortest_path_follower_example.py
```diff
@@ -0,0 +1,48 @@
...
+            cv2.putText(
+                im,
+                "Action: " + best_action.value,
+                (0, 20),
+                cv2.FONT_HERSHEY_SIMPLEX,
+                0.5,
+                (0, 0, 255),
+            )
+            cv2.imshow("im", im)
+            cv2.waitKey(1)
```
mathfac: Will it play well with tests?

habitat/tasks/nav/shortest_path_follower.py
```diff
@@ -0,0 +1,184 @@
...
+EPSILON = 1e-6
+
+
+def action_to_one_hot(action: int) -> np.array:
+    one_hot = np.zeros(len(SIM_NAME_TO_ACTION), dtype=np.float32)
+    one_hot[action] = 1
+    return one_hot
+
+
+class ShortestPathFollower:
```
mathfac: Let's add docstring for class as well.

examples/shortest_path_follower_example.py
```diff
@@ -0,0 +1,48 @@
...
+            im = observations["rgb"][:, :, ::-1].copy()
+            cv2.putText(
+                im,
+                "Action: " + best_action.value,
+                (0, 20),
+                cv2.FONT_HERSHEY_SIMPLEX,
+                0.5,
+                (0, 0, 255),
+            )
+            cv2.imshow("im", im)
```
mathfac: Does this work in headless mode?
We can use condition
```
if "pytest" not in sys.modules:
 ...
```
to guard unexpected behavior during testing.

erikwijmans:
Changes to magic constants is fine by me

danielgordon10:
@erikwijmans I just meant in general make sure I didn't mess something up and that my comments match what you think they should be.

.gitignore
danielgordon10: My problem here is this does not work for symlinks. I looked online and it seems the only way to ignore symlinks is to not use slashes. It's generally bad practice to have your dataset inside your project folder anyways (causes you to need multiple copies). I don't know what the right solution is here.

examples/shortest_path_follower_example.py
danielgordon10: I know habitat has some viewer to show the current image if you built it, but I don't know how to use it. This was a quick workaround for sanity checking that actually helped me catch a bug, but I still think it's useful to have. Ideally calling env.render() would pop up a screen like it does in openai-gym, but that is not the current behavior.

habitat/tasks/nav/shortest_path_follower.py
danielgordon10: You ask so much of me

examples/shortest_path_follower_example.py
mathfac: maybe let's use save image option as in previous examples?

.gitignore
mathfac: Ok, let's remove slashes.

.gitignore
```diff
@@ -42,6 +42,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+examples/images
```
mathfac: Thank you, I felt that pain!

habitat/sims/habitat_simulator.py
```diff
@@ -451,3 +454,18 @@ def distance_to_closest_obstacle(self, position, max_search_radius=2.0):
         return self._sim.pathfinder.distance_to_closest_obstacle(
             position, max_search_radius
         )
+
+    @property
+    def up_vector(self):
```
mathfac: Let's add this method to Simulator API `core/simulator.py`.

habitat/sims/habitat_simulator.py
```diff
@@ -451,3 +454,18 @@ def distance_to_closest_obstacle(self, position, max_search_radius=2.0):
...
+        return np.array([0.0, 1.0, 0.0])
+
+    @property
+    def forward_vector(self):
+        return -np.array([0.0, 0.0, 1.0])
+
+    def get_shortest_path_points(self, position_a, position_b):
```
mathfac: Let's create method similar to this: https://github.com/facebookresearch/habitat-api/blob/master/habitat/core/simulator.py#L288 and call it `shortest_path` that returns
ShortestPathPoints.

erikwijmans: We should name it `straight_shortest_path`.  If the shortest path is calculated on an occupancy grid via A*, the greedy follower will definitely not work as expected and will generate very strange paths.  You'd get much better paths from the code version that to estimates the geodesic gradient and follows that.

habitat/sims/habitat_simulator.py
danielgordon10: I see what you mean now. shortest_path_follower expects a HabitatSim object rather than an arbitrary Simulator object, so I don't know if this makes sense to add to the Simulator API. It is possible that different simulators will need different algorithms for shortest paths.

habitat/sims/habitat_simulator.py
danielgordon10: I see what you mean now. shortest_path_follower expects a HabitatSim object rather than an arbitrary Simulator object, so I don't know if this makes sense to add to the Simulator API. It is possible that different simulators will need different algorithms for shortest paths.

habitat/sims/habitat_simulator.py
danielgordon10: I changed the name and updated the docs. Maybe instead of straight it should be geodesic though?

mathfac: @erikwijmans, can you elaborate more what `straight` stands for here and how it's different from path used for geodesic distance calculation?

habitat/sims/habitat_simulator.py
mathfac: @danielgordon10, yes we are overfitting to Habitat Sim here, but we will revisit the design choices when new simulators will need to be integrated.

PR #68, opened 2019-04-23
Properly set agent position

erikwijmans:
## Motivation and Context

The agent's state contains the absolute position of the sensors.  When we call `set_agent_state`, we just provide the position of the agent's body and expect the sensors to follow.  This PR simply doesn't provide any sensor states and things fall-back to behaving as they previously did.

## How Has This Been Tested

Moving some episode visualization code to be on master and caught this.

## Types of changes

### DIFF
--- a/habitat/sims/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator.py
@@ -421,6 +421,10 @@ class HabitatSim(habitat.Simulator):
         state.position = position
         state.rotation = rotation
 
+        # NB: The agent state also contains the sensor states in _absolute_ coordinates.
+        # In order to set the agent's body to a specific location and have the sensors follow,
+        # we must not provide any state for the sensors.
+        # This will cause them to follow the agent's body
         state.sensor_states = dict()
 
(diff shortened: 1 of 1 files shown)


habitat/sims/habitat_simulator.py
```diff
@@ -420,6 +420,9 @@ def set_agent_state(
         state = self.get_agent_state(agent_id)
         state.position = position
         state.rotation = rotation
+
```
mathfac: Do we want to add comment here, to make sure it won't be removed mistakenly.

erikwijmans: Done

erikwijmans:
> Is this connected with switching to new simulator API

Yeah, I added the sensor states to the agent's state in https://github.com/facebookresearch/habitat-sim/pull/6 and didn't think to update the set_agent_state logic here

mathfac:
@erikwijmans, curious why any of the sensors didn't catch this? I believe eqa tests should fail as it checks RGB mean value for frames.

PR #69, opened 2019-04-23
Adds an ipython notebook illustrating simple use cases

bhavanajain:
## Types of changes

## Checklist


configs/tasks/pointnav_mp3d.yaml
```diff
@@ -24,6 +27,6 @@ TASK:
     SUCCESS_DISTANCE: 0.2
 DATASET:
   TYPE: PointNav-v1
-  SPLIT: train
+  SPLIT: val 
   POINTNAVV1:
-    DATA_PATH: data/datasets/pointnav/mp3d/v1/{split}/{split}.json.gz
+    DATA_PATH: /private/home/bhavanajain/habitat-notebooks/habitat-api/data/datasets/pointnav/mp3d/v1/{split}/{split}.json.gz
```
mathfac: it should be changed to previous path.

bhavanajain:
> @bhavanajain thank you for sending first notebook.
> That would be great if we can include it as unit test to our testing suite. So we keep it up to date. Maybe something like this can help load notebook code and run it as test:
> 
> ```
> class NotebookLoader(object):
>     """Module Loader for Jupyter Notebooks"""
>     def __init__(self, path=None):
>         self.shell = InteractiveShell.instance()
>         self.path = path
> 
>     def load_module(self, fullname):
>         """import a notebook as a module"""
>         path = find_notebook(fullname, self.path)
> 
>         print ("importing Jupyter notebook from %s" % path)
> 
>         # load the notebook object
>         with io.open(path, 'r', encoding='utf-8') as f:
>             nb = read(f, 4)
> 
> 
>         # create the module and add it to sys.modules
>         # if name in sys.modules:
>         #    return sys.modules[name]
>         mod = types.ModuleType(fullname)
>         mod.__file__ = path
>         mod.__loader__ = self
>         mod.__dict__['get_ipython'] = get_ipython
>         sys.modules[fullname] = mod
> 
>         # extra work to ensure that magics that would affect the user_ns
>         # actually affect the notebook module's ns
>         save_user_ns = self.shell.user_ns
>         self.shell.user_ns = mod.__dict__
> 
>         try:
>           for cell in nb.cells:
>             if cell.cell_type == 'code':
>                 # transform the input to executable Python
>                 code = self.shell.input_transformer_manager.transform_cell(cell.source)
>                 # run the code in themodule
>                 exec(code, mod.__dict__)
>         finally:
>             self.shell.user_ns = save_user_ns
>         return mod
> ```
How about using nbval to test jupyter notebooks?
https://github.com/computationalmodelling/nbval

mathfac:
You can update `configs/tasks/pointnav_mp3d.yaml` in the notebook, so we can leave it unchanged.

PR #71, opened 2019-04-24
Fixed render function in vector environment

### LINKED ISSUE #70: Incorrect render() method in VectorEnv class.
https://github.com/facebookresearch/habitat-api/blob/8c5329ce9f777395b45c2daac135377e629ec148/habitat/core/vector_env.py#L313-L331
In this function you send tuple 
```python
(args, {"mode": "rgb_array", **kwargs})
``` 
However, it does not contain command index that you want to be executed by worker. Furthermore, your environments do not have ```rgb_array``` mode of rendering, they use ```rgb``` instead. This makes ```render()``` incorrect.
I think you have to replace line with ```write_fn``` call with:
```python
write_fn( (RENDER_COMMAND, (args, {"mode": "rgb", **kwargs})) )
```

titardrew:
## Motivation and Context

`render()` function in `VectorEnv` sends wrong message to workers.
Fixes #70 

## How Has This Been Tested

Added test case for `VectorEnv` rendering with `rgb_array` mode.

## Types of changes

- Bug fix (#70)

## Checklist

### DIFF
--- a/test/test_habitat_env.py
+++ b/test/test_habitat_env.py
@@ -211,6 +211,7 @@ def test_rl_vectorized_envs():
         assert len(rewards) == num_envs
         assert len(dones) == num_envs
         assert len(infos) == num_envs
+        assert envs.render(mode='rgb_array')
         if (i + 1) % configs[0].ENVIRONMENT.MAX_EPISODE_STEPS == 0:
             assert all(dones), "dones should be true after max_episode steps"
 
--- a/test/test_habitat_env.py
+++ b/test/test_habitat_env.py
@@ -211,7 +211,7 @@ def test_rl_vectorized_envs():
         assert len(rewards) == num_envs
         assert len(dones) == num_envs
         assert len(infos) == num_envs
-        assert envs.render(mode='rgb_array')
+        assert envs.render(mode="rgb_array")
         if (i + 1) % configs[0].ENVIRONMENT.MAX_EPISODE_STEPS == 0:
             assert all(dones), "dones should be true after max_episode steps"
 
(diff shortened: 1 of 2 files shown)


test/test_habitat_env.py
```diff
@@ -211,6 +211,7 @@ def test_rl_vectorized_envs():
         assert len(rewards) == num_envs
         assert len(dones) == num_envs
         assert len(infos) == num_envs
+        assert envs.render(mode="rgb_array")
```
abhiskk: Should this be `"rgb"`, instead of `"rgb_array"`. 

Also on a note of testing this, I don't think adding an assert check here is a good idea because the render method will display a window with tiled images. Just testing it locally and posting a screenshot would be useful for testing this functionality.

abhiskk:
Thanks @titardrew , I added comments on the PR.

test/test_habitat_env.py
titardrew: Thanks for commenting on this!
1) It has to be `"rgb_array"` because render function looks like:
https://github.com/facebookresearch/habitat-api/blob/37fc4de0216fd3287cb08f9f7b6e7349c1f286b6/habitat/core/vector_env.py#L406-L424
So it does not know anything about `"rgb"` mode. The thing is, that you have to pass `"rgb"` mode only to worker, not to VectorEnv's render function. If you want consistency at this point, then we have to change render function by replacing all mentions of `"rgb_array"` by `"rgb"`.
2) Well, no it won't display a window. It does that when mode is set to 'human'. In the case of `"rgb_array"` mode it will just render this image and return it as an array, without opening a window. Anyway, I can save image returned by this function and post it here if it is needed.

abhiskk: Thanks, sorry I had forgotten the exact working of the render method, can you add a message along with the assert check, `assert envs.render(mode="rgb_array"), "vector env render is broken"` and just add an image to the PR, it will be good to go then.

titardrew: Sure

PR #76, opened 2019-04-26
Small changes to the topdown map interface to make plotting the agent easier

danielgordon10:
Turns out we were missing a few things from the previous code.


danielgordon10:
image
All tests pass. Black is happy. I added to the shortest path example so it now produces top down maps as well and uses the `image_to_video` function. Sample image pasted above.

mathfac:
Looks great, thank you!

PR #80, opened 2019-05-06
Adds get_observation function

danielgordon10:
## Motivation and Context
Utility method for getting the current observation. Sometimes you want the observation without having to make a step/reset call for whatever reason.

## Types of changes

## Checklist


danielgordon10:
I was using this after calling set_agent_state which doesn't return an observation. Maybe instead set_agent_state should return an observation?

erikwijmans:
There is utility in being able to get observations at a specific location outside of the normal episodic way with `reset`.  One primary use-case being using habitat to generate a dataset for other computer vision tasks on the fly -- tasks such as depth from single image, semantic segmentation, object detection, etc.

I do think that this is a use case we should support.  Maybe a cleaner API would be a `get_observations_at` function that takes a position and heading, or a full agent state including sensors positions?  We'd currently just implement that with `set_agent_state` and the `get_observation` code in this PR, but wouldn't tie ourselves to implementing it like that going forward.

dhruvbatra:
>  One primary use-case being using habitat to generate a dataset for other computer vision tasks on the fly -- tasks such as depth from single image, semantic segmentation, object detection, etc.

> I do think that this is a use case we should support. Maybe a cleaner API would be a get_observations_at function that takes a position and heading, or a full agent state including sensors positions? 

Completely agree! We have been asked about this in past issues and by our collaborators in person.

mathfac:
@danielgordon10, lets implement `get_observations_at` function with `set_agent_state` and the `get_observation`. Actually, I needed that functionality today as well to check IOU from particular agent position.

danielgordon10:
I don't understand the point of `get_observation_at` What is the use case? Do we want to move the agent there, get the observation, then move the agent back?

erikwijmans:
The primary use-case in my mind would be for generating datasets on the fly for various different SL tasks (depth from single image, semantic segmentation, etc.).  I imagine a loop like this:

```
for batch in range(batch_size):
    pos, rotation = sim.sample_random_state()
    obs = sim.get_observation_at(pos, rotation)
```

So yes, we'd currently move the agent to the location, get observation, and move it back, but we'd be allowing ourselves to implement it differently in the future.

dhruvbatra:
Yup, this is an important use case to generate photo-realistic egocentric CV datasets, and we need to support it. In addition to the for loop with random viewpoints that @erikwijmans mentioned, we should (eventually) consider supporting generating egocentric videos from realistic generated trajectories. Like this video: https://youtu.be/z8uJh_xUq7A?t=92

danielgordon10:
I made a few fairly significant changes to the interface to try and make the results more interpretable in terms of what happens in the case of failures. I also added a flag for returning back to the original location.

habitat/sims/habitat_simulator.py
```diff
@@ -392,51 +392,90 @@ def _get_agent_config(self, agent_id: Optional[int] = None) -> Any:
...
+    def get_observations_at(
+        self,
+        position: List[float],
+        rotation: List[float],
+        keep_agent_at_new_pose: bool = False,
+    ) -> Optional[Observations]:
 
-        self._check_agent_position(position, agent_id)
+        # TODO do we need to support multiple agents for this?
+        agent_id = 0
```
mathfac: `agent_id = 0` can be removed, we have default value for that, just `self.get_agent_state()`

danielgordon10: Done.

danielgordon10:
Added tests.

PR #89, opened 2019-05-21
Modify partitioning logic to handle left-over scenes due to integer d…

abhiskk:
## Motivation and Context

If `len(scenes)` is not divisible by `args.num_processes` then there will be some scenes left over which will not be assigned to any subprocess. This PR handles such cases.

## Types of changes

## Checklist


erikwijmans:
Given 7 scenes and 4 processes, this will end up with 3 processes having 1 scene and 1 process having 4 scenes.  It would instead make more sense to give 3 process 2 scenes each and 1 process 1 scene.

This can be done with something like
```
scenes_for_procs = [[] for _ in range(args.num_processes)]
next_proc_id = 0
for s in scenes:
    scenes_for_procs[next_proc_id].append(s)
    next_proc_id = (next_proc_id  + 1) % args.num_processes

...

config_env.DATASET.POINTNAVV1.CONTENT_SCENES = scenes_for_procs[i]
```

mathfac:
Yes, that should work, `len(scenes)` is usually < 300.

PR #92, opened 2019-05-29
isort to sort imports alphabetically

erikwijmans:
## Motivation and Context

We like pretty code, isort makes code prettier, thus we use isort.  I also added isort to the CI to keep things pretty.

`isort -rc .`
`black --line-length 79 .`

## How Has This Been Tested

Just re-orders imports! 

Although I choose the isort configs to make isort and black play nice, so I may have overfit to my test-set :)


setup.py
```diff
@@ -4,14 +4,22 @@
...
 import os.path
 import sys
 
 import setuptools
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "habitat"))
+
 from version import VERSION  # noqa
 
+
```
abhiskk: Is BLACK okay with this, this seems like a PEP violiation

habitat_baselines/agents/slam_agents.py
```diff
@@ -1,39 +1,38 @@
+# https://sumit-ghosh.com/articles/python-download-progress-bar/
```
abhiskk: Is this getting used somewhere?

Also I just noticed that this file is missing the facebook header, I can add it in another PR.

setup.py
erikwijmans: Seems like it didn't run on setup.py for some reason....

habitat_baselines/agents/slam_agents.py
erikwijmans: Looks like that comment is referencing this function: https://github.com/facebookresearch/habitat-api/blob/master/habitat_baselines/agents/slam_agents.py#L39

Moved it to there

erikwijmans:
I'm currently trying to figure out why isort is failing with travis.  It works locally.....

erikwijmans:
It's different versions of isort :/

erikwijmans:
@mathfac I believe this was done with #97?

mathfac:
@erikwijmans, right.

erikwijmans:
Closing then!

PR #94, opened 2019-05-30
Enable optional habitat_baselines installation

JasonJiazhiZhang:
## Motivation and Context
  * Enable the optional installation of habitat_baselines package and its necessary dependencies, in addition to the core habitat package and its dependencies. Implemented through `--user-option` and setuptools command overriding. 
  *  Exclude `test/`, `examples/` by default when finding package during installation, and further exclude `habitat_baselines/` if only installing habitat core.
  * Create separate requirement.txt file for each type of baseline approach
To install only habitat core (with pip editable mode):
```
pip install -e .
```
To install both habitat and habitat_baselines (along with dependencies of both):
```
python setup.py develop  --all                  

# pip install -e .  --install-option="--all"     will NOT work 
# as --install-option is inconsistent between editable mode and standard mode
```
Standard installation (non-develop mode, future proofing):
```
# habitat core only:
pip install .

# core + baseline
pip install . --install-option="--all"
```

## How Has This Been Tested
This setup.py file has been tested through a clean install under the following environment:
```
Platform: linux Ubuntu 18.04.1 LTS 
Python 3.6.8
pytest-4.5.0, py-1.8.0, pluggy-0.11.0
cuda/9.0
anaconda3/5.0.1
cudnn/v7.0-cuda.9.0
gcc/7.1.0
cmake/3.10.1/gcc.5.4.0
```
Both    `pip install -e .` and   `python setup.py develop  --all` have been tested, with `python setup.py test` passing  (35 passed 6 skipped) and `python examples/benchmark.py` runnable.

## Types of changes

## Checklist


JasonJiazhiZhang:
The main README.md is not updated yet.

mathfac:
As we discussed functionality we are looking for:
```
pip install habitat-api -> can import habitat core, no habitat_baseline
```
```
pip install habitat-api['all'] -> can import habitat_baseline and all dependencies are installed

```
cc @abhiskk

JasonJiazhiZhang:
This PR has been revamped to not only enable optional dependency installation, but also optional package inclusion/exclusion of habitat_baselines.
Avoiding the hacky approach of using `sys.argv` directly in setup.py, this implementation utilizes `--install-option` argument of pip, which passes additional args to `setuptools.setup`. Command method overriding is required to properly use the additional arguments passed in. Discussion and suggestions about this approach and potentially better approaches are welcomed.  

Two caveats regarding this implementation:
  * Ideally we want to use pip for all 4 types of installation (develop core, develop core+baseline, install core, install core+baseline), unfortunately `pip -e` has inconsistency when passing `--install-option` and we are forced to directly use python setup.py develop for the "develop core+baseline" scenario.
  * The overridden command (install/develop) modifies arguments passed in setuptools.setup() through a very poorly documented interface. Although it's not likely to be problematic since this interface originates from distutils of Python core. 

@mathfac  @abhiskk @apsdehal

habitat_baselines/rl/requirements.txt
```diff
@@ -0,0 +1 @@
+torch>=1.0.0
```
apsdehal: It will be better if you fix the torch version. The inconsistencies and breaking changes in PyTorch version has broken pythia more than once.

setup.py
```diff
@@ -6,8 +6,10 @@
 
 import os.path
 import sys
-
+import glob
 import setuptools
+from setuptools.command.develop import develop
```
apsdehal: Is this blacked yet?

setup.py
```diff
@@ -26,13 +28,55 @@
 LONG_DESCRIPTION = readme
 AUTHOR = "Facebook AI Research"
 LICENSE = license
-REQUIREMENTS = (reqs.strip().split("\n"),)
+REQUIREMENTS = reqs.strip().split("\n")
+DEFAULT_EXCLUSION = ["test", "examples"]
+
```
apsdehal: NIT: Extra new line.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+FULL_REQUIREMENTS = []
```
apsdehal: This should technically be a set.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+# collect requirements.txt file in all subdirectories
+for file_name in glob.glob("**/*requirements.txt", recursive=True):
```
apsdehal: What did we finalize on moving this out of the package as a separate package?

apsdehal: Standard file name is `requirements.txt`. Why are you looking for `*requirements.txt`. It would be good to follow standard for this.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+    with open(file_name) as f:
+        reqs = f.read()
+        FULL_REQUIREMENTS.extend(reqs.strip().split("\n"))
+
+
+# overriding install and develop
+class OptionedCommand(object):
```
apsdehal: In python3, you don't need to inherit object.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+    user_options = [("all", None, "include habitat_baselines in installation")]
+
+    def initialize_options(self):
+        super().initialize_options()
```
apsdehal: This is bad practice. if someone inherited `OptionedCommand` without inheriting relevant module, this will crash, let's inherit proper class here.

apsdehal: Discussed with Jiazhi. For now, since there is no way to go around this, we can keep this. Most of the code that I looked at which does similar kind of thing, actually does their extra stuff in common functions.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+        self.all = None
+
+    def run(self):
+        if not self.all:
+            DEFAULT_EXCLUSION.extend(
```
mathfac: Is there an option to specify instead `INCLUSION_ONLY` option? Most probably we don't want install anything else except `habitat` folder and don't want to update this lined with each new folder added.

setup.py
```diff
@@ -26,13 +28,55 @@
...
+                ["habitat_baselines", "habitat_baselines.*"]
```
mathfac: Let's create a constant and have closer to the begging of the file.

setup.py
mathfac: Maybe make defaultness explicit as `install` and `develop` are too ambiguous:
```suggestion
from setuptools.command.develop import develop as DefaultDevelopCommand
```

setup.py
mathfac: Agree with @apsdehal, let's add a verbose comment, regarding what's happening here.

setup.py
JasonJiazhiZhang: After some digging, I didn't find a more elegant way of doing this. `find_package` unfortunately does not have something similar to `INCLUSION ONLY`.

setup.py
```diff
@@ -6,8 +6,10 @@
 
 import os.path
 import sys
-
+import glob
```
mathfac: run `isort` tool please

JasonJiazhiZhang: isort will not only move up `import glob` but also swap the two lines below:
```
sys.path.insert(0, os.path.join(os.path.dirname(__file__), "habitat"))
from version import VERSION  # noqa
```
Which causes installation failure and causes failure in CI test. 
Ignoring the swap right now.

mathfac:
We can send changes if we have more comments/concerns regarding the approach. Merging.

PR #98, opened 2019-06-03
Action space changes, added look_up and look_down actions

danielgordon10:
## Motivation and Context
THIS CHANGE IS NOT BACKWARDS COMPATIBLE!!!

Changing action space names to say what they actually do, adding the look up and look down actions

## How Has This Been Tested

Tested before ICCV experiment push, not positive that Look Up and Look Down still work. You will need a new build of habitat-sim to support the look action changes.

## Types of changes

## Checklist


danielgordon10:
An important change I made was that the `test_trajectory_sim.py` test now expects stuff like `MOVE_FORWARD` rather than 0. I think this is more obvious and ties the code less to the numbers and more to them as constants which is how I think they should be thought of. But idk how this json was generated, so there may be another change that I need to add that changes how trajectories are recorded.

mathfac:
Daniel, thank you for the follow up. 
1. Does this PR change default action space?
2. I have a feeling that it will break [current pretrained checkpoints](https://github.com/facebookresearch/habitat-challenge#starter-code-and-training) based on[ Agent API.](https://github.com/facebookresearch/habitat-api/blob/master/habitat/core/agent.py#L24)

erikwijmans:
It will break current checkpoints, although they are fairly easy to fix as we can just re-order the weights in the final FC layer (or add a new layer that re-order the logits).

erikwijmans:
With that said, maybe the correct thing to do is have a registry for action spaces.  This will be important as people want to build new actions and integrate them into habitat-api.  We can then change the default action space to be `[stop, move_forward, turn_left, turn_right]` with registered action spaces for the old `[move_forward, turn_left, turn_right, stop]` and another registered on for  `[stop, move_forward, turn_left, turn_right, look_up, look_down]`

danielgordon10:
@mathfac It will change the default action space and break the checkpoints, but it should be possible to permute the weights in the checkpoints to fix this. I can help with this if necessary.

mathfac:
> With that said, maybe the correct thing to do is have a registry for action spaces. This will be important as people want to build new actions and integrate them into habitat-api. We can then change the default action space to be `[stop, move_forward, turn_left, turn_right]` with registered action spaces for the old `[move_forward, turn_left, turn_right, stop]` and another registered one for `[stop, move_forward, turn_left, turn_right, look_up, look_down]`
Yes, I think that's right direction to go, while @danielgordon10 PR is still step forward.

@danielgordon10, can you run and make sure no tests are failing:
```
pytest
black . --line-length 79
```

danielgordon10:
@mathfac All tests and black pass. Doesn't the Travis check those things?
@mathfac @erikwijmans I personally find the registry pattern you guys have very hard to follow code-wise since it sort of seems like you need an instance of a thing to actually look at the values, and then you have to step through all this registry garbage when debugging. But I agree that it would be nice to specify in a config-sort-of-way. I may not be the right person to make the registry change, I'm not exactly sure how it works in the first place. But for now I can change the order back to stop being 4th action, but keep in the name changes and the `look_up` and `look_down` actions.

erikwijmans:
@danielgordon10 I also agree that the current registry is confusing, would love your thoughts on https://github.com/facebookresearch/habitat-api/pull/99 :)

PR #99, opened 2019-06-04
Fixes the registry

erikwijmans:
## Motivation and Context

Our registry pattern is far from idea and thus we haven't used it consistent.  This PR aims to resolve this by copying the registry pattern used in [Pythia](https://github.com/facebookresearch/pythia) (as it is much nicer) and then registering all current measures and sensors and using the register to query for those instead of `getattr` as this is a more clearly extendable pattern (as compared to monkey patching).

The simulator registry is somewhat weird as we want to be able to register simulators only if they are installed.  I have a work-around for this, but maybe what we currently have for the simulators is better.

## How Has This Been Tested

`pytest`

## Types of changes


danielgordon10:
I like that this gets rid of the weird non-blank __init__.py files. Those already seem like magic to me, so this change is appreciated. I also think it's clearer what you need to add/change and where when extending the framework. However, it could use some dummy example in the documentation if you had time.

erikwijmans:
Added an example :)

examples/register_new_sensors_and_measures.py
```diff
@@ -0,0 +1,104 @@
...
+    def reset_metric(self, episode):
+        # Our measure always contains all the attributes of the episode
+        self._metric = vars(episode).copy()
+        # But only on reset, it has an additional field of my_value
+        self._metric["my_value"] = self._config.VALUE
+
+    # This is called whenver an action is taken in the environment
+    def update_metric(self, episode, action):
+        # Now the measure will just have all the attributes of the episode
+        self._metric = vars(episode).copy()
```
mathfac: Not sure, why you need to update `self._metric` each step, value doesn't change.

examples/register_new_sensors_and_measures.py
```diff
@@ -0,0 +1,104 @@
...
+    # Defines the name of the measure in the measurements dictionary
+    def _get_uuid(self, *args: Any, **kwargs: Any):
+        return "episode_info"
+
+    # This is called whenver the environment is reset
+    def reset_metric(self, episode):
+        # Our measure always contains all the attributes of the episode
+        self._metric = vars(episode).copy()
+        # But only on reset, it has an additional field of my_value
+        self._metric["my_value"] = self._config.VALUE
```
mathfac: Not sure it's clear why `my_value` can be needed here. We can call it `EPSIODE_STARTED` or have a measure that counts different actions taken, etc.

examples/register_new_sensors_and_measures.py
erikwijmans: It's a demo.  I am just showing how to add a new metric and what the difference between reset and update by setting it at every time-step and doing the `my_value` thing.

examples/register_new_sensors_and_measures.py
erikwijmans: The reason to have the value is so that I can have something that is settable via the config system to show how adding things to the config shows up in the measure.

PR #100, opened 2019-06-04
Add single-episode config for baseline training

JasonJiazhiZhang:
## Motivation and Context
Add single-episode training config. This option will signal the dataset generation process to generate dataset with a single episode. This could be useful as a sanity check when developing new algorithms, since any  effective algorithm should overfit on single episode fairly quickly. Such testing strategy will be incorporated later in baseline algorithm tests. 

  * Add config option NUM_EPISODE_SAMPLE, with default value -1 that
indicates use all episodes. When set value to non-negative integer,
pointna_dataset will sample the specified number of episodes.
  * To use this option, either add the `single_episode.yaml` config file in the training bash command, e.g.:
```bash
python -u habitat_baselines/train_ppo.py\
         ... 
         ...
   --task-config "configs/tasks/pointnav.yaml,configs/datasets/single_episode.yaml"
```
  * or do:
```bash
python -u habitat_baselines/train_ppo.py\
         ... 
         ...
         DATASET.NUM_EPISODE_SAMPLE 1
```

  * Fix the functionality of adding configs at the end of command line. Now config will be correctly added (e.g. code block above)

## How Has This Been Tested
All established test and benchmarks run successfully.

Script with the following params observes a much more rapid increase in value-loss when using the single-episode option.
```
python -u habitat_baselines/train_ppo.py\
     --use-gae     --sim-gpu-id 0     --pth-gpu-id 0\
     --lr 2.5e-5     --clip-param 0.1     --value-loss-coef 0.5\
     --num-processes 1     --num-steps 16     --num-mini-batch 1\
     --num-updates 2000     --use-linear-lr-decay\
     --use-linear-clip-decay     --entropy-coef 0.01\
     --log-file "train.log"     --log-interval 5\
     --checkpoint-folder "data/checkpoints"\
     --checkpoint-interval 50\
     --task-config "configs/tasks/pointnav.yaml,configs/datasets/single_episode.yaml"\
     --seed 123
```

## Types of changes

## Checklist


configs/datasets/single_episode.yaml
```diff
@@ -0,0 +1,2 @@
+DATASET:
+  NUM_EPISODE_SAMPLE: 1
```
apsdehal: Ideally, updating this should be done via specifying `DATASET.NUM_EPISODE_SAMPLE 1`, which should be specified in documentation and then you don't even need this config. Even better, add a `test_run` argument which just loads this configuration automatically.

habitat/core/dataset.py
```diff
@@ -224,3 +224,12 @@ def get_splits(
         if remove_unused_episodes:
             self.episodes = new_episodes
         return new_datasets
+
+    def sample_episodes(self, num_episodes):
```
apsdehal: You are missing type annotations here.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -99,6 +99,9 @@ def __init__(self, config: Optional[Config] = None) -> None:
             with gzip.open(scene_filename, "rt") as f:
                 self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
 
+        if config.NUM_EPISODE_SAMPLE >= 0:
```
apsdehal: This can probably be optimized. Instead of loading all of the episodes and then sampling one from them, is there other way round where we only load the required episode?

JasonJiazhiZhang: Asked Oleksandr before, and will discuss again. My current understanding is that the single-episode option will likely be used with a small dataset to begin with, e.g. habitat_test_scenes, and we don't care that much about the cost of loading such a small dataset in its entirety.

apsdehal: Cool. From developer's perspective, a lot of times I will be testing it, so I need a faster alternative, but then again it depends on which direction the project is going in.

configs/datasets/single_episode.yaml
JasonJiazhiZhang: Agreed. Creating a separate config file (with only 2 lines) just for this functionality seems clunky. I think without any changes or any additional config file, this config should be able to be passed by adding `opts=[DATASET.NUM_EPISODE_SAMPLE, 1]` to the command line call (according to [this documentation](https://github.com/facebookresearch/habitat-api/blob/f1c6922df77a3224107c7f27bb0b40e34e9f253e/habitat/config/default.py#L175-L177)). However, I don't think it's working right now. Alternatively, I can add a `test-run` argument, which will either require the two-liner yaml file or require additional modification in training script. Will discuss with Oleksandr.

apsdehal: The way to do it in yacs is that you just add `DATASET.NUM_EPISODE_SAMPLE 1` to the end of your command.

JasonJiazhiZhang: Discussed with Amanpreet through chat. We are missing something like [this](https://github.com/facebookresearch/pythia/blob/fd2c9a52cbb983488fe62f1c4ec8719070bd07d0/pythia/utils/flags.py#L224-L229). Should work once we add this.

mathfac: Exactly `get_config` function has this option, but we need to pass arguments there. Makes sense to add such option to `train_ppo.py`.

habitat/datasets/pointnav/pointnav_dataset.py
mathfac: @apsdehal current dataset format is gzip json that can't be loaded partially. For large datasets we have separate file per scene. 1-10k (usual size for val split) episodes are really fast to load as file size is several kb.

mathfac: @JasonJiazhiZhang let's update EQADataset as well accordingly.

mathfac:
Please, rebase on master, we have additional CI test enabled.

habitat/datasets/eqa/mp3d_eqa_dataset.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, config: Config = None) -> None:
         ) as f:
             self.from_json(f.read())
 
+        if config.NUM_EPISODE_SAMPLE >= 0:
```
mathfac: Let's move it to `sample_episodes(self, num_episodes: int)` to have this check for all calls.

mathfac: @JasonJiazhiZhang this checks not needed anymore.

habitat/datasets/pointnav/pointnav_dataset.py
mathfac: @JasonJiazhiZhang this checks not needed anymore.

PR #110, opened 2019-06-06
Extendable action space

erikwijmans:
## Motivation and Context

We need to be able to extend actions in habitat-api!

Core changes:
* SimulatorActions is no longer an enum.  It is an extendable singleton.  It allows you to extend the action space but it is quite strict about it.  Actions cannot be put on different ints, actions cannot be remove.
* There is now a registry for action space configurations. 

This doesn't solve the issue of being able to change the action name to index mapping, but I don't think that is something that we should try to solve, we should just be very conservative about changing it (which I think is/was already the plan).

## How Has This Been Tested

pytest + new example :)

## Types of changes


habitat/core/simulator.py
```diff
@@ -366,3 +453,11 @@ def render(self, mode: str = "rgb") -> Any:
 
     def close(self) -> None:
         raise NotImplementedError
+
+    @property
+    def index_stop_action(self):
```
mathfac: Does this properties are forcing to have stop and forward actions?

habitat/sims/habitat_simulator/action_spaces.py
```diff
@@ -0,0 +1,41 @@
...
+import habitat_sim
+from habitat.core.registry import registry
+from habitat.core.simulator import (
+    ActionSpaceConfiguration,
+    Config,
+    SimulatorActions,
+)
+
+
+@registry.register_action_space_configuration(name="v1")
```
mathfac: Can we register as v1 the one without look up/look down to be compatible with Habitat whitepaper? Our checkpoints are not loading because of the number of actions as well.

habitat/core/simulator.py
erikwijmans: Yes, it does force us to have stop and forward (although those are currently also forced as they are always in `SimulatorActions`)

habitat/sims/habitat_simulator/action_spaces.py
erikwijmans: Done.  I called the whitepaper one v0

mathfac:
@erikwijmans test are constantly failing.

test/test_habitat_env.py
```diff
@@ -132,7 +134,9 @@ def test_threaded_vectorized_env():
     envs = habitat.ThreadedVectorEnv(env_fn_args=env_fn_args)
     envs.reset()
     non_stop_actions = [
-        v for v in range(len(SimulatorActions)) if v != SimulatorActions.STOP
+        act
```
mathfac: Looks like we could reduce copypasta using util function: `non_stop_actions(env)`

PR #112, opened 2019-06-07
Make train_ppo compatible with both PointNav goal sensors

JasonJiazhiZhang:
## Motivation and Context
Make  train_ppo.py compatible with both types of PointNav goal sensor (static_pointgoal/pointgoal). Remove hardcoded goal/task/dataset information from the script. Now train_ppo.py access those information from config files specified through `--task-config` option.  
Policy, Agent, Benchmark, and tests are modified accordingly.

In the future, would like to make PPO general for different tasks, not just for PointNav tasks. But this is beyond the scope of this PR. 

## How Has This Been Tested
test_baseline_agents.py has been modified for this change. Other tests pass as usual.
```
================================================================= test session starts =================================================================
platform linux -- Python 3.7.3, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
rootdir: /private/home/jiazhi/habitat-api, inifile: setup.cfg, testpaths: test
collected 44 items

test/test_baseline_agents.py ..                                                                                                                 [  4%]
test/test_config.py ..                                                                                                                          [  9%]
test/test_dataset.py ..........                                                                                                                 [ 31%]
test/test_habitat_env.py ..........s                                                                                                            [ 56%]
test/test_habitat_example.py ....                                                                                                               [ 65%]
test/test_install.py .                                                                                                                          [ 68%]
test/test_mp3d_eqa.py sss                                                                                                                       [ 75%]
test/test_pointnav_dataset.py .ss.                                                                                                              [ 84%]
test/test_sensors.py .....                                                                                                                      [ 95%]
test/test_trajectory_sim.py ..                                                                                                                  [100%]

=============================================================== short test summary info ===============================================================
SKIPPED [1] test/test_habitat_env.py:348: unconditional skip
SKIPPED [1] /private/home/jiazhi/habitat-api/test/test_mp3d_eqa.py:93: Please download Matterport3D EQA dataset to data folder.
SKIPPED [1] /private/home/jiazhi/habitat-api/test/test_mp3d_eqa.py:111: Please download Matterport3D EQA dataset to data folder.
SKIPPED [1] /private/home/jiazhi/habitat-api/test/test_mp3d_eqa.py:150: Please download Matterport3D EQA dataset to data folder.
SKIPPED [1] /private/home/jiazhi/habitat-api/test/test_pointnav_dataset.py:68: Test skipped as dataset files are missing.
SKIPPED [1] /private/home/jiazhi/habitat-api/test/test_pointnav_dataset.py:96: Test skipped as dataset files are missing.
======================================================= 38 passed, 6 skipped in 166.61 seconds ========================================================
```

## Types of changes

## Checklist


habitat_baselines/agents/ppo_agents.py
```diff
@@ -135,11 +139,14 @@ def main():
     )
     args = parser.parse_args()
 
-    config = get_defaut_config()
-    config.INPUT_TYPE = args.input_type
-    config.MODEL_PATH = args.model_path
+    task_config = get_config(args.task_config)
```
mathfac: Logically `task_config` is `config.TASK`. Let's use, just `config` here.

habitat/core/benchmark.py
```diff
@@ -22,13 +22,11 @@ class Benchmark:
...
     Args:
-        config_paths: file to be used for creating the environment.
-        config_dir: directory where config_paths is located.
+        config: environment config object for creating environment.
     """
 
-    def __init__(self, config_paths: Optional[str] = None) -> None:
-        config_env = get_config(config_paths)
-        self._env = Env(config=config_env)
+    def __init__(self, config: Optional[Config] = None) -> None:
```
mathfac: There is a reason why we specify path here, not config itself.

habitat/config/default.py
```diff
@@ -30,6 +30,7 @@
 _C.TASK.SUCCESS_DISTANCE = 0.2
 _C.TASK.SENSORS = []
 _C.TASK.MEASUREMENTS = []
+_C.TASK.GOAL_SENSOR_UUID = "pointgoal"
```
mathfac: Maybe both `_C.TASK.SUCCESS_DISTANCE` and `_C.TASK.GOAL_SENSOR_UUID` worth to move to `_C.TASK.NAV.GOAL_SENSOR_UUID/SUCCESS_DISTANCE` as it doesn't look general for all tasks types.

habitat_baselines/train_ppo.py
```diff
@@ -158,7 +158,7 @@ def construct_envs(args):
     return envs
 
 
-def main():
+def run_training():
```
mathfac:
```suggestion
def run_training(args):
```
Maybe, lets add `args` as params that will make possible to run this function from tests.

habitat/config/default.py
JasonJiazhiZhang: Didn't realize that both `TASK.SUCCESS_DISTANCE` and `TASK.GOAL_SENSOR_UUID` aren't general. Then it looks like there's much more work to be done to generalize agents

habitat_baselines/train_ppo.py
JasonJiazhiZhang: I'll leave this as is for now, will make changes like this in a separate PR about making train_ppo testable.

mathfac:
Checked the changes, looks good to me. Will leave to @abhiskk, @apsdehal, @erikwijmans to take a look. Can you test it with running PPO training with `static_pointgoal`? Thanks!

JasonJiazhiZhang:
> Checked the changes, looks good to me. Will leave to @abhiskk, @apsdehal, @erikwijmans to take a look. Can you test it with running PPO training with `static_pointgoal`? Thanks!

Refactored `train_ppo.py` runs `static_pointgoal` without issues. Just need to change SENSORS and GOAL_SENSOR_UUID in config file

habitat_baselines/train_ppo.py
```diff
@@ -111,8 +111,8 @@ def construct_envs(args):
     baseline_configs = []
 
     basic_config = cfg_env(config_paths=args.task_config, opts=args.opts)
-
-    scenes = PointNavDatasetV1.get_scenes_to_load(basic_config.DATASET)
+    dataset = make_dataset(basic_config.DATASET.TYPE)
+    scenes = dataset.__class__.get_scenes_to_load(basic_config.DATASET)
```
apsdehal: Is this the right way to do it? It should be a class method.

habitat_baselines/rl/ppo/policy.py
```diff
@@ -63,10 +71,12 @@ class Net(nn.Module):
     goal vector with CNN's output and passes that through RNN.
     """
 
-    def __init__(self, observation_space, hidden_size):
+    def __init__(self, observation_space, hidden_size, goal_sensor_uuid):
```
erikwijmans: What a "goal_sensor" is is very task specific.  Different task will have wildly different ideas of a goal sensor (i.e. EQA has questions).  The network here should be a default network that doesn't have an idea of a goal and then you make a PointNavNetwork by inheriting from the Net and implementing something like `get_goal_embedding`.  Now instead of `x =  observations[self.goal_sensor_uuid]` you have something like `x = self.get_goal_embedding(observations)`.  That would just return `observations['pointgoal']` for PointNav, but allows something like EQA or ObjectNav to be done with minimal changes.

habitat_baselines/rl/ppo/policy.py
```diff
@@ -12,12 +12,20 @@
 
 
 class Policy(nn.Module):
```
erikwijmans: The policy should either take which type of network it uses as an argument or this is done via inheritance (I am more partial to an argument).  See bellow for the reason for different networks.

habitat_baselines/rl/ppo/policy.py
mathfac: @erikwijmans, agree with all your points. Maybe, we can make it as a next step, as a the scope of this task was more narrow. @JasonJiazhiZhang, let's reflect it in PR title. Something, like `Made PPO less tangled with pointnav task."

abhiskk:
Sorry but I am slightly lost for context here, the aim of the PR is a good one but I don't have a concrete way of checking whether the generalization to different task works well. I see that the PointNav sensor has been abstracted away into a parameter but the codebase still feels very attached to the PointNav task. @JasonJiazhiZhang if you are planning to make it general for other tasks maybe a good way to test the generalization is to add a dummy task (like RoomNav) and then testing how you can modify the code for generalizing to that usecase.

habitat/config/default.py
```diff
@@ -158,7 +161,6 @@
 _C.DATASET.POINTNAVV1.DATA_PATH = (
```
mathfac: Can we also do this change:
```suggestion
_C.DATASET.DATA_PATH = (
```

habitat_baselines/train_ppo.py
mathfac: @apsdehal, what exactly do you mean?

erikwijmans: Class and static methods are also callable/work as expected when using an instance of the class.
i.e. `scenes = dataset.get_scenes_to_load(basic_config.DATASET)`

JasonJiazhiZhang: Fixed.

mathfac:
Can you re-merge with master as it has conflict. Are you user Mp3d_eqa tests are passing?
I would expect value of MP3D default path should be adde here: https://github.com/facebookresearch/habitat-api/blob/master/configs/test/habitat_mp3d_eqa_test.yaml

PR #116, opened 2019-06-07
Adding jupyter notebook illustrating the cameras

oawiles:
## Motivation and Context

Demonstrates how to extract the cameras in the habitat API and how these relate to
the given depth and new views.

## How Has This Been Tested

The jupyter notebook can be used to render random examples to verify that the
camera transformations are correct.

## Types of changes

## Checklist


mathfac:
@oawiles thank you for sending the PR and intention to make Habitat better. Can you elaborate more what exactly you call camera here and what the example should illustrate?

@msavva, @erikwijmans, maybe that make more sense to have this notebook on `habitat-sim` side as it illustrates HSIM related entities.
Regarding code itself:
- You can check [the existing notebook](https://github.com/facebookresearch/habitat-api/blob/master/notebooks/habitat-api-demo.ipynb) regarding code style: access of config fields, name of the notebook file, comments style etc.
- Looks like there is a redundant code in the notebook: we can get rid of SimpleRLEnv definition and goal radius and use `Env` class.

oawiles:
I was trying to use the habitat environment for doing view synthesis, e.g. given a viewpoint then project it into a new viewpoint. However, how to actually go from one view to another was not clear (e.g. the change of coordinate system, the intrinsic parameters, etc.) as this is not the main use of the code base. With some help from erikwijmans, I managed to get this to work. He/Manolis suggested that this could be useful for other people and that I should create a pull request to demonstrate how I did this? If this is not actually helpful no worries.

erikwijmans:
Yep, I definitely think this is helpful.  I think it makes sense to live in the API repo.  It would also make sense to live in the simulator repo, but no need to re-write/move it.

erikwijmans:
Couple suggestions:
1.  Let's name the notebook something a little more descriptive.  Like `relative_camera_views_demo`?  (@msavva @mathfac suggestions?)

2.

```python
    env._sim.set_agent_state(agent_id=0, position=translation, 
                         rotation=rotation)

    rgb = env.render(mode='rgb')
    depth = env.render(mode='depth')[:,:,0]

    depths += [depth]
    rgbs += [rgb]
```
should be
```python
    obs = env._sim.get_observations_at(position=translation, rotation=rotation)
    depths += [obs["depth"][...,0]]
    rgbs += [obs["rgb"]
```

erikwijmans:
The negotiations on the rotation and translation parts of the transformation matrix was confusing to me.  I took at stab at re-writing that so it is less confusing

```
# Intrinsic parameters, K

K = np.array([
    [1 / np.tan(hfov / 2.), 0., 0., 0.],
    [0., 1 / np.tan(hfov / 2.), 0., 0.],
    [0., 0.,  1, 0],
    [0., 0., 0, 1]])

# Now get an approximation for the true world coordinates -- see if they make sense
# [-1, 1] for x and [1, -1] for y as array indexing is y-down while world is y-up
xs, ys = np.meshgrid(np.linspace(-1,1,W), np.linspace(1,-1,W)) 
depth = depths[0].reshape(1,W,W) 
xs = xs.reshape(1,W,W)
ys = ys.reshape(1,W,W)

# Unproject
# negate depth as the camera looks along -Z
xys = np.vstack((xs * depth , ys * depth, -depth, np.ones(depth.shape)))
xys = xys.reshape(4, -1)
xy_c0 = np.matmul(np.linalg.inv(K), xys)

# Now load in the cameras, are in the format camera --> world
# Camera 1:
quaternion_0 = cameras[0].sensor_states['depth'].rotation
translation_0 = cameras[0].sensor_states['depth'].position
rotation_0 = quaternion.as_rotation_matrix(quaternion_0)
T_world_camera0 = np.eye(4)
T_world_camera0[0:3,0:3] = rotation_0
T_world_camera0[0:3,3] = translation_0

# Camera 2:
translation_1 = cameras[1].sensor_states['depth'].position
quaternion_1 = cameras[1].sensor_states['depth'].rotation
rotation_1 = quaternion.as_rotation_matrix(quaternion_1)
T_world_camera1 = np.eye(4)
T_world_camera1[0:3,0:3] =  rotation_1
T_world_camera1[0:3,3] = translation_1

# Invert to get world --> camera
T_camera1_world = np.linalg.inv(T_world_camera1)

# Transformation matrix between views
# Aka the position of camera0 in camera1's coordinate frame
T_camera1_camera0 = np.matmul(T_camera1_world, T_world_camera0)

# Finally transform actual points
xy_c1 = np.matmul(T_camera1_camera0, xy_c0)
xy_newimg = np.matmul(K, xy_c1)

# Normalize by negative depth
xys_newimg = xy_newimg[0:2,:] / -xy_newimg[2:3,:]
# Flip back to y-down to match array indexing
xys_newimg[1] *= -1
print('Transformed points')
```

This outputs the same thing and is hopefully more straight forward (there are a bunch of different ways to do this, so it may be that we are just familiar with different ways :) )

oawiles:
OK that sounds good. I'll make those changes and re-push.

oawiles:
So I tried to change, as you suggested above to:
`obs = env._sim.get_observations_at(position=translation, rotation=rotation)`.
However, I got an error: AttributeError: 'HabitatSim' object has no attribute 'get_observations_at'/

I made the other changes. I agree that the way you wrote it makes the change in coordinate system clearer.

erikwijmans:
> So I tried to change, as you suggested above to:
obs = env._sim.get_observations_at(position=translation, rotation=rotation).
However, I got an error: AttributeError: 'HabitatSim' object has no attribute 'get_observations_at'/

You probably need to install habitat-api again (that was recently landed).

oawiles:
OK I tried to change as you suggested:

`    obs = env._sim.get_observations_at(position=translation, rotation=rotation)
      depths += [obs["depth"][...,0]]
      rgbs += [obs["rgb"]`

However, this causes problems as the given translation and rotation are not those that are used for the camera (they are the agent's transformations I believe). However, when I use `set_agent_state` then this updates the cameras correctly. I'm not sure as to the relationship between these two sets of cameras.

erikwijmans:
Ah yes, we need to add `keep_agent_at_new_pose=True`

oawiles:
OK I've made the change and now it all seems to work.

mathfac:
@oawiles thank you for the iterations. Can you add similar notebook test like [this](https://github.com/facebookresearch/habitat-api/blob/master/test/test_demo_notebook.py) and make sure it's passing?

oawiles:
OK I'll make the suggestions you said. However, I'm not sure the relative path update is right. When I run the notebook it expects the paths to be there (and this is also how the [other](https://github.com/facebookresearch/habitat-api/blob/master/notebooks/habitat-api-demo.ipynb) notebook does it) so I think it is a property of jupyter.

mathfac:
@oawiles my bad, you’re right.

oawiles:
OK done.

mathfac:
@oawiles thank you for contributing such non-trivial and important example notebook. I am sure others will benefit from it and for all follow-ups to keep it runnable/testable and self-explanatory.

oawiles:
@mathfac Thank you!

erikwijmans:
Yes, thank you very much!

PR #117, opened 2019-06-07
Fixes and upgrades the collision logic

erikwijmans:
## Motivation and Context

The current collision logic is kinda buggy and won't scale to arbitrary actions.  This PR uses https://github.com/facebookresearch/habitat-sim/pull/70 to detect collisions, which will hopefully be less buggy and scale to arbitrary actions!

@abhiskk @mathfac -- The API for using this is kinda weird as whether a collision occurred or not is calculated when a step happens and cannot easily be recalculated :/

## How Has This Been Tested

Lots and lots and lots of invocations of `pytest`

## Types of changes

### DIFF
--- a/habitat/core/simulator.py
+++ b/habitat/core/simulator.py
@@ -366,3 +366,13 @@ class Simulator:
 
     def close(self) -> None:
         raise NotImplementedError
+
+    @property
+    def previous_step_collided(self):
+        r"""Whether or not the previous step resulted in a collision
+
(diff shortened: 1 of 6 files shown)


erikwijmans:
(This is expected to fail until facebookresearch/habitat-sim#70)

erikwijmans:
Landed on sim side

erikwijmans:
Added docstring

PR #119, opened 2019-06-11
Add helper function for mapping between action spaces

erikwijmans:
## Motivation and Context

We recently re-order our action number, which breaks checkpoints :(  This PR adds a helper function to help with mapping from the old action space to the new action space.  The help function maps on the level of probabilities or logits as this is more general and differentiable, making it also compatible with imitation learning.

## How Has This Been Tested

Via tests and applying this to our depth-PPO baselines.  On master, that baselines gets an SPL of 0, using this mapping, it gets 0.58 SPL (which is what it gets on the leaderboard).

## Types of changes

### DIFF
--- a/habitat/utils/action_space_mapping.py
+++ b/habitat/utils/action_space_mapping.py
@@ -1,0 +1,121 @@
+from typing import Dict
+
+import numpy as np
+
+
+def build_mapping_matrix(mapping: Dict[int, int]) -> np.ndarray:
+    r"""Creates a matrix to map probabilities from one action space to another
+
+    Params:
+        mapping (Dict[int, int]):  The mapping between action spaces.  Keys in the dictionary
+            are the index in the current action space and the corresponding value is the action
+            index in the new action space
+
+    Returns:
+        (np.ndarray): Matrix to map between action spaces
+
+
+    Usage::
+        
+        # Mapping between stop on 3 and stop on 0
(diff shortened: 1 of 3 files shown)


test/test_action_mapping.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
```
mathfac: Facebook header is missing.

habitat/utils/action_space_mapping.py
```diff
@@ -0,0 +1,121 @@
+from typing import Dict
```
mathfac: Facebook header is missing.

habitat/utils/action_space_mapping.py
```diff
@@ -0,0 +1,121 @@
...
+        mask = np.ones_like(logits) @ mapping_matrix
+        flmin = np.full(
+            (logits.shape[0], mask.shape[1]), np.finfo(np.float32).min
+        )
+
+    return logits @ mapping_matrix + (1.0 - mask) * flmin
+
+
+def map_action_distribution(
+    mapping: Dict[int, int], *, probs=None, logits=None, use_torch=False
```
mathfac: Can we add types for `probs , logits`?

mathfac:
Feel free to reopen the PR if it's relevant. Closing for now.

PR #121, opened 2019-06-12
Fix the citation

erikwijmans:
## Motivation and Context

Our bibtex is wrong.  `,` is used to denote `last, first` format while `and` is used to separate authors. I also added caps

## How Has This Been Tested

With my eyes

Current:
image

After:
image


mathfac:
I remember @abhiskk mentioned that stars mess sorting during referencing.

PR #127, opened 2019-06-14
Add tensorboard and video generation for ppo train and eval

JasonJiazhiZhang:
## Motivation and Context

  * Add checkpoint progress tracking for evalute_ppo. Now when specified with a checkpoint directory, 
evaluate_ppo will evaluate checkpoints in chronological order, and constantly check for new checkpoint.  
  * Add tensorboard visualization to both train_ppo and evaluate_ppo
  * Add video generation for navigation episode evaluation. Generated videos can be either saved locally or visualized through tensorboard.
  * Add shortest

 path to top-down map
  * Make api changes necessary for tensorboard and video generation support

## How Has This Been Tested
pytest passes. Tensorboard visualization works as expected for both training and evaluation.
Demo

## Types of changes

## Checklist


habitat/tasks/nav/nav_task.py
```diff
@@ -393,10 +393,12 @@ def reset_metric(self, episode):
...
         if self._metric is None:
-            self._metric = 0
-
+            self._metric = dict(count=0, is_collision=False)
         if self._sim.previous_step_collided:
-            self._metric += 1
+            self._metric["count"] += 1
+            self._metric["is_collision"] = True
+        else:
+            self._metric["is_collision"] = False
```
apsdehal: NIT. Can be written in this way:
```
         self._metric["is_collision"] = False
         if self._sim.previous_step_collided:
            self._metric["count"] += 1
            self._metric["is_collision"] = True
```

habitat/tasks/nav/nav_task.py
```diff
@@ -482,6 +485,22 @@ def reset_metric(self, episode):
...
+                p[0],
+                p[2],
+                self._coordinate_min,
+                self._coordinate_max,
+                self._map_resolution,
+            )
+            for p in self._shortest_path_points
+        ]
+        self._shortest_path_points = [
+            (y, x) for (x, y) in self._shortest_path_points
```
apsdehal: NIT:
```suggestion
            coord[::-1] for coord in self._shortest_path_points 
```

habitat/tasks/nav/nav_task.py
```diff
@@ -491,6 +510,29 @@ def reset_metric(self, episode):
...
+        self._draw_shortest_path()
+
+    def _draw_shortest_path(self):
+        for prev_pt, next_pt in zip(
+            self._shortest_path_points[:-1], self._shortest_path_points[1:]
+        ):
+            if (
+                self._top_down_map[prev_pt[0], prev_pt[1]]
+                != maps.MAP_SOURCE_POINT_INDICATOR
+            ):
```
apsdehal: make this more readable by moving `self._top_down_map[prev_pt[0], prev_pt[1]]` to a variable

habitat/tasks/nav/nav_task.py
```diff
@@ -491,6 +510,29 @@ def reset_metric(self, episode):
...
+                color = 7
+                thickness = int(
```
apsdehal: A comment will be helpful here.

habitat/tasks/nav/nav_task.py
```diff
@@ -491,6 +510,29 @@ def reset_metric(self, episode):
...
+
+    def _draw_shortest_path(self):
+        for prev_pt, next_pt in zip(
+            self._shortest_path_points[:-1], self._shortest_path_points[1:]
+        ):
+            if (
+                self._top_down_map[prev_pt[0], prev_pt[1]]
+                != maps.MAP_SOURCE_POINT_INDICATOR
+            ):
+                color = 7
```
apsdehal: a comment here too, what does 7 refer to? Maybe give a reference if someone wants to change color.

habitat/tasks/nav/nav_task.py
apsdehal: it seems like this constant is defined in other file. Maybe define here too or save all constants in a separate file.

habitat/tasks/nav/nav_task.py
```diff
@@ -491,6 +510,29 @@ def reset_metric(self, episode):
             self._map_resolution,
         )
         self._previous_xy_location = (a_y, a_x)
+        self._draw_shortest_path()
+
+    def _draw_shortest_path(self):
```
apsdehal: Ideally, this should be independent of the task and a separate utility function which can be used by any class.

habitat/tasks/nav/nav_task.py
apsdehal: This reversal can also be done in line 491.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+from habitat.utils.visualizations.utils import images_to_video
 from config.default import get_config as cfg_baseline
 from habitat.config.default import get_config
 from rl.ppo import PPO, Policy
 from rl.ppo.utils import batch_obs
 from train_ppo import make_env_fn
+from habitat.utils.visualizations import maps
+
+
+def poll_checkpoint_folder(checkpoint_folder, previous_ckpt_ind):
```
apsdehal: Move to utils.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    assert os.path.isdir(checkpoint_folder), "invalid checkpoint folder path"
```
apsdehal: I don't have background here, but when the checkpoint is not there in case of starting, should it not silently ignore and return?

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    models = os.listdir(checkpoint_folder)
+    models.sort(key=lambda x: int(x.strip().split(".")[1]))
+    ind = previous_ckpt_ind + 1
+    if ind < len(models):
+        return os.path.join(checkpoint_folder, models[ind])
+    return None
+
+
+def images_to_tb_video(video_name, step_idx, frames, writer, fps=10):
```
apsdehal: All these functions should go into a separate file for drawing and mapping utilities.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    frame_tensors = [
+        torch.from_numpy(np_arr).unsqueeze(0) for np_arr in frames
+    ]
+    video_tensor = torch.cat(tuple(frame_tensors))
```
apsdehal: Why are we converting to numpy and then converting back to torch tensors? torch tensors have an `unsqueeze` function. Also, cat can handle lists, any specific reason you want to convert this to tuple. Lastly, it is always a good practice to specify the dim over which you want to concatenate.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    video_tensor = video_tensor.permute(0, 3, 1, 2).unsqueeze(
```
apsdehal: Exact comment on (initial_dims) => (final_dims) would be helpful.

habitat_baselines/train_ppo.py
```diff
@@ -224,137 +225,184 @@ def run_training():
...
+                    step_observation = {
+                        k: v[step] for k, v in rollouts.observations.items()
+                    }
+
+                    (
+                        values,
+                        actions,
+                        actions_log_probs,
+                        recurrent_hidden_states,
+                    ) = actor_critic.act(
```
apsdehal: Let's fix readability here.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    writer.add_video(video_name, video_tensor, fps=fps, global_step=step_idx)
+
+
+def draw_collision(view):
+    size = view.shape[0]
+    strip_width = int(size / 20)
+    mask = np.ones((size, size))
+    mask[strip_width:-strip_width, strip_width:-strip_width] = 0
+    mask = mask == 1
+    alpha = 0.4
```
apsdehal: Constants.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    video_tensor = torch.cat(tuple(frame_tensors))
+    video_tensor = video_tensor.permute(0, 3, 1, 2).unsqueeze(
+        0
+    )  # Shape (N,T,C,H,W)
+    writer.add_video(video_name, video_tensor, fps=fps, global_step=step_idx)
+
+
+def draw_collision(view):
+    size = view.shape[0]
+    strip_width = int(size / 20)
```
apsdehal: Use `//`

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,59 @@
...
+    mask = np.ones((size, size))
+    mask[strip_width:-strip_width, strip_width:-strip_width] = 0
+    mask = mask == 1
+    alpha = 0.4
+    view[mask] = (alpha * np.array([255, 0, 0]) + (1.0 - alpha) * view)[mask]
```
apsdehal: You might want to look into `cv2.addWeighted` function which provides even more granular control.

habitat_baselines/evaluate_ppo.py
```diff
@@ -112,8 +188,17 @@ def main():
         args.num_processes, args.hidden_size, device=device
     )
     not_done_masks = torch.zeros(args.num_processes, 1, device=device)
+    stats_episodes = set()
+    if args.video == 1:
+        rgb_frames = [[]] * args.num_processes
+        if not os.path.exists(args.out_dir_video):
+            os.makedirs(args.out_dir_video)
+    else:
```
apsdehal: NIT: Set rgb_frames to None first and then jump into if clause, this will save one line.

habitat_baselines/evaluate_ppo.py
```diff
@@ -112,8 +188,17 @@ def main():
         args.num_processes, args.hidden_size, device=device
     )
     not_done_masks = torch.zeros(args.num_processes, 1, device=device)
+    stats_episodes = set()
+    if args.video == 1:
+        rgb_frames = [[]] * args.num_processes
+        if not os.path.exists(args.out_dir_video):
+            os.makedirs(args.out_dir_video)
```
apsdehal: Use `exist_ok=True` as argument to `os.makedirs`, then you don't need to check for the path.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +234,120 @@ def main():
...
 
+        # added for video support
+        next_episodes = envs.current_episodes()
+        envs_to_pause = []
+        n_envs = envs.num_envs
+        for i in range(n_envs):
+            if next_episodes[i].episode_id in stats_episodes:
+                envs_to_pause.append(i)
+
+            if not_done_masks[i].item() == 0:
```
apsdehal: Move all this to a separate function.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +234,120 @@ def main():
...
+                # new episode ended, record stats
+                stats_episodes.add(current_episodes[i].episode_id)
+                if args.video == 1 and len(rgb_frames[i]) > 0:
+
+                    video_name = "{}{}_{}{}_{}{:.2f}".format(
+                        "episode",
```
apsdehal: Move fixed string into format string.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +234,120 @@ def main():
...
+                        "episode{}".format(current_episodes[i].episode_id),
+                        checkpoint_idx,
+                        rgb_frames[i],
+                        writer,
+                        fps=10,
+                    )
+
+                    rgb_frames[i] = []
+
+            elif args.video == 1:
```
apsdehal: Separate function

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +234,120 @@ def main():
...
+                observation_size = observations[i]["rgb"].shape[0]
```
apsdehal: This all should again go into separate file, so it is easy to use in other training and evaluation scripts.

habitat_baselines/evaluate_ppo.py
JasonJiazhiZhang: In main() it constantly checks for new checkpoints. If not found it will sleep for 2 sec and check again.

habitat/tasks/nav/nav_task.py
```diff
@@ -491,6 +510,15 @@ def reset_metric(self, episode):
             self._map_resolution,
         )
         self._previous_xy_location = (a_y, a_x)
+        # draw optimal path
+        maps.draw_path(
```
erikwijmans: @mathfac @abhiskk Thoughts on draw order?  Atm the optimal path will be drawn over the agents path

habitat_baselines/rl/ppo/utils.py
```diff
@@ -422,4 +425,73 @@ def ppo_args():
...
+
+    top_down_map = info["top_down_map"]["map"]
+    top_down_map = maps.colorize_topdown_map(top_down_map)
+    map_agent_pos = info["top_down_map"]["agent_map_coord"]
+
+    top_down_map = maps.draw_agent(
+        image=top_down_map,
+        agent_center_coord=map_agent_pos,
+        agent_rotation=info["top_down_map"]["agent_angle"],
+        agent_radius_px=8,
```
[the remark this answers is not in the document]
JasonJiazhiZhang: Default map size for me is 135. Will scale based on 135 for now

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +212,71 @@ def main():
...
     episode_success_mean = (episode_success / episode_counts).mean().item()
 
-    print("Average episode reward: {:.6f}".format(episode_reward_mean))
-    print("Average episode success: {:.6f}".format(episode_success_mean))
-    print("Average episode spl: {:.6f}".format(episode_spl_mean))
+    logger.info("Average episode reward: {:.6f}".format(episode_reward_mean))
+    logger.info("Average episode success: {:.6f}".format(episode_success_mean))
+    logger.info("Average episode SPL: {:.6f}".format(episode_spl_mean))
+
+    writer.add_scalar("Average reward", episode_reward_mean, checkpoint_idx)
```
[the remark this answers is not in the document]
JasonJiazhiZhang: Currently train and eval process are storing metrics in different folders already and need to both be specified when launching tensorboard like this `tensorboard --logdir train:tbdir,eval:tb_eval`. This is mainly to accommodate video recording. However I think what you suggest is good practice and will make this change anyways.

habitat/tasks/nav/nav_task.py
JasonJiazhiZhang: Not sure what you mean by this. Draw_path will only happen in `reset_metric()`. Agents path will be drawn step by step in `update_metric()`-> `update_map()`, over the optimal path.

habitat_baselines/rl/ppo/utils.py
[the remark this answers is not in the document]
erikwijmans: I believe 135 is the size after crop. 1250 is the size before crop. IDK which one is correct to normalize against. Normalization against the size before the crop would match what’s going on now.

habitat/tasks/nav/nav_task.py
erikwijmans: Seems like I miss understood!

habitat_baselines/rl/ppo/utils.py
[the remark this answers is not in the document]
JasonJiazhiZhang: Edited my comments above. Will scale according to cropped map for now because that function does not have access to original resolution

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +212,71 @@ def main():
...
+                if args.video == 1 and len(rgb_frames[i]) > 0:
+                    video_name = "episode{}_ckpt{}_spl{:.2f}".format(
+                        current_episodes[i].episode_id,
+                        checkpoint_idx,
+                        infos[i]["spl"],
+                    )
+                    images_to_video(
+                        rgb_frames[i], args.out_dir_video, video_name
+                    )
+                    frames_to_tb_video(
```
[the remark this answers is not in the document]
JasonJiazhiZhang: Plan to make video option [0, 1, 2, 3] instead of [0, 1]. 0 for no video, 1 for saving to dir, 2 for tb, 3 for both, which avoids making a separate flag.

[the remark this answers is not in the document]
erikwijmans: Maybe named options make more sense? i.e. `["no_video", "video_dir", "video_tb", "video_both"]`?

habitat/core/vector_env.py
```diff
@@ -186,6 +187,9 @@ def _worker_env(
                     else:
                         result = getattr(env, function_name)(*function_args)
                     connection_write_fn(result)
+                elif command == EPISODE_COMMAND:
+                    current_episode = env.current_episode
```
abhiskk: This is not needed here, you can directly do `connection_write_fn(env.current_episode)`

habitat/tasks/nav/nav_task.py
```diff
@@ -393,10 +393,11 @@ def reset_metric(self, episode):
 
     def update_metric(self, episode, action):
         if self._metric is None:
-            self._metric = 0
-
+            self._metric = dict(count=0, is_collision=False)
```
abhiskk: Use `{"count": 0, "is_collision": False}` 

https://stackoverflow.com/a/6612024

habitat/tasks/nav/nav_task.py
```diff
@@ -458,7 +466,7 @@ def get_original_map(self, episode):
             point_padding = 2 * int(
                 np.ceil(self._map_resolution[0] / MAP_THICKNESS_SCALAR)
             )
-            top_down_map[
+            self._top_down_map[
                 s_x - point_padding : s_x + point_padding + 1,
```
abhiskk: Is a `+ 1` needed here?

habitat/tasks/nav/nav_task.py
```diff
@@ -515,8 +544,21 @@ def update_metric(self, episode, action):
                 map_agent_x - (self._ind_x_min - self._grid_delta),
                 map_agent_y - (self._ind_y_min - self._grid_delta),
             ),
+            "agent_angle": self.get_polar_angle(),
         }
 
+    def get_polar_angle(self):
+        agent_state = self._sim.get_agent_state()
+        # Quaternion is in x, y, z, w format
```
abhiskk: If possible start comments with a lower case.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,20 +5,38 @@
...
+from habitat.utils.visualizations.utils import images_to_video
 from rl.ppo import PPO, Policy
-from rl.ppo.utils import batch_obs
+from rl.ppo.utils import batch_obs, frames_to_tb_video, generate_frame
 from train_ppo import make_env_fn
 
 
+def poll_checkpoint_folder(checkpoint_folder, previous_ckpt_ind):
+    assert os.path.isdir(checkpoint_folder), "invalid checkpoint folder path"
+    models = os.listdir(checkpoint_folder)
```
abhiskk: I think `os.listdir` returns the filenames in a specific order, we might not needed to sort and check them. Also it would be useful to check if model has been successfully evaluated in a better manner, it could be the case that `ckpt23` will appear after `ckpt123` even though `ckpt123` is written after `ckpt23` and you would miss out on this detail. My suggestion is to order model files in the order of time they were created.

I see that below you are sorting them according to the index, it would be better to use timing I think instead of index, also can you add a comment to make this explicit.

habitat_baselines/evaluate_ppo.py
```diff
@@ -38,12 +56,53 @@ def main():
...
     args = parser.parse_args()
+    assert (args.model_path is not None) != (
+        args.tracking_model_dir is not None
+    ), "Must specify a single model or a directory of models, but not both"
 
-    device = torch.device("cuda:{}".format(args.pth_gpu_id))
+    writer_kwargs = dict(
+        log_dir=args.tensorboard_dir, purge_step=0, flush_secs=30
+    )
+    with (tensorboard.SummaryWriter(**writer_kwargs)) as writer:
```
abhiskk: Can you add an option to not specify a tensorboard writer also, if someone does not want to use that.

habitat_baselines/evaluate_ppo.py
```diff
@@ -38,12 +56,53 @@ def main():
...
+            prev_ckpt_ind = -1
+            while True:
+                current_ckpt = None
+                while current_ckpt is None:
+                    current_ckpt = poll_checkpoint_folder(
+                        args.tracking_model_dir, prev_ckpt_ind
+                    )
+                    time.sleep(2)  # sleep for 2 seconds before polling again
+
+                logger.warning(
```
abhiskk: use `logger.info` here instead of warning

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +219,82 @@ def main():
...
+        envs_to_pause = []
+        n_envs = envs.num_envs
+        for i in range(n_envs):
+            if next_episodes[i].episode_id in stats_episodes:
+                envs_to_pause.append(i)
+
+            if not_done_masks[i].item() == 0:
+                # episode ended, record and generate videos as specified
+                stats_episodes.add(current_episodes[i].episode_id)
+                if args.video_option != 0 and len(rgb_frames[i]) > 0:
```
abhiskk: Abstract out the video generation logic into a separate method, the main function is becoming really cumbersome here.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +219,82 @@ def main():
...
+        if len(envs_to_pause) > 0:
+            # stop tracking ended episodes
+
+            state_index = list(range(envs.num_envs))
+            for idx in reversed(envs_to_pause):
+                state_index.pop(idx)
+                envs.pause_at(idx)
+
+            # indexing along the batch dimensions
+            test_recurrent_hidden_states = test_recurrent_hidden_states[
```
abhiskk: Be sure to double check this properly in your local test, it is very easy to make a mistake here, the code looks correct.

habitat_baselines/rl/ppo/utils.py
```diff
@@ -422,4 +425,72 @@ def ppo_args():
         nargs=argparse.REMAINDER,
         help="Modify config options from command line",
     )
+    parser.add_argument(
```
abhiskk: Add an option to not specify the `--tensorboard-dir`, in case someone does not want to use tensorboard.

habitat_baselines/rl/ppo/utils.py
```diff
@@ -422,4 +425,72 @@ def ppo_args():
...
+        default="tb_train",
+        help="path to logging dir for Tensorboard",
+    )
     return parser
+
+
+def frames_to_tb_video(video_name, step_idx, frames, writer, fps=10):
+    """
+    write video into tensorboard from images frames
+    :param video_name: name of video string
```
abhiskk: Be consistent in writing docstring, I think this does not conform to the format used elsewhere.

habitat_baselines/train_ppo.py
abhiskk: I think the `train_ppo` code in general needs a revamp and better abstractions, I am at fault here as this was written in a hasty manner. We should discuss how to restructure this.

habitat/tasks/nav/nav_task.py
JasonJiazhiZhang: This is part of danielgordon10's original implementation that I just left intact. I do think, however, that I should make the if check outside of the function call to make the logic more clear.

habitat_baselines/rl/ppo/utils.py
JasonJiazhiZhang: Thanks for noticing that. I found that **Settings | Tools | Python Integrated Tools | Docstring format** to **Google** in PyCharm will output correct Docstring.

habitat_baselines/evaluate_ppo.py
JasonJiazhiZhang: Just changed the sorting mechanism to be based on last modification time. Now user's checkpoint naming convention won't affect sorting, however this does require user to ckpts in a new dir for each new run or delete ckpts of old run.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +200,153 @@ def main():
...
+                    rgb_frames[i],
+                    current_episodes[i].episode_id,
+                    checkpoint_idx,
+                    infos[i]["spl"],
+                    writer,
+                )
+                rgb_frames[i] = []
+
+            # episode continues
+            elif args.video_option != "no_video":
```
erikwijmans: This should just be an `if`.  ATM the video won't contain the first frame.

JasonJiazhiZhang: I can see how this can cause a problem. But with the current implementation, instead of missing the first frame, should it be missing the last frame? Making it an `if` block and move it before the `if` block for "episode ended" should fix the issue I think.

erikwijmans: When the stop action is called, the environment returns the observations for the start of the next episode (assuming auto reset is turned on). The last frame is the frame that lead to stop being taken, and that’s already being recorded.

JasonJiazhiZhang: @erikwijmans I see. The issue now is that when stop action is called, observation and info returned seem to be from different episodes (observation is from next episode, info from cur episode). This causes mis-match of topdown map and egocentric views. A video demo sent privately.

JasonJiazhiZhang: Reverted back to skipping first frame, as a tmp fix. Will need further discussion (and potential changes to how top-down-map is represented) for a full fix.

habitat/tasks/nav/nav_task.py
```diff
@@ -445,43 +450,42 @@ def get_original_map(self, episode):
...
-                self._map_resolution,
-            )
-            top_down_map[
-                t_x - point_padding : t_x + point_padding + 1,
-                t_y - point_padding : t_y + point_padding + 1,
-            ] = maps.MAP_TARGET_POINT_INDICATOR
-
         return top_down_map
 
+    def draw_source_and_target(self, episode):
```
mathfac: That would be great to simplify this code with creating TopDownMap class one day. Created separate issue: https://github.com/facebookresearch/habitat-api/issues/142.

habitat/tasks/nav/nav_task.py
```diff
@@ -515,8 +543,21 @@ def update_metric(self, episode, action):
                 map_agent_x - (self._ind_x_min - self._grid_delta),
                 map_agent_y - (self._ind_y_min - self._grid_delta),
             ),
+            "agent_angle": self.get_polar_angle(),
```
mathfac: Better be specific here:
```suggestion
            "agent_polar_angle": self.get_polar_angle(),
```

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +199,155 @@ def main():
...
+        default="configs/tasks/pointnav.yaml",
+        help="path to config yaml containing information about task",
+    )
+    parser.add_argument(
+        "--video-option",
+        type=str,
+        default="no_video",
+        choices=["no_video", "video_dir", "video_tb", "video_both"],
+        help="option for video generation. no_video (default): do not generate vidoe at all; "
+        "video_dir: generate videos and save in video-dir; "
```
mathfac: Confusing with other flag.
```suggestion
        "video_disk: generate videos and save in video-dir; "
```

habitat_baselines/train_ppo.py
```diff
@@ -224,137 +226,179 @@ def run_training():
...
-                    actions,
-                    actions_log_probs,
-                    recurrent_hidden_states,
-                ) = actor_critic.act(
-                    step_observation,
-                    rollouts.recurrent_hidden_states[step],
-                    rollouts.masks[step],
-                )
-            pth_time += time() - t_sample_action
+                outputs = envs.step([a[0].item() for a in actions])
```
mathfac: Hope we will simplify train_ppo in future PRs.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +199,155 @@ def main():
...
+                    current_episodes[i].episode_id,
+                    cur_ckpt_idx,
+                    infos[i]["spl"],
+                    writer,
+                )
+                rgb_frames[i] = []
+
+            # episode continues
+            elif args.video_option != "no_video":
+                frame = generate_frame(observations[i], infos[i])
```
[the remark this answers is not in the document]
JasonJiazhiZhang: Official fix for misaligned obs and info will be in a separate PR

habitat/tasks/nav/nav_task.py
erikwijmans: The method should be just `get_agent_angle()` as there isn't a difference between cartesian and polar for angles.  i.e. a polar_angle doesn't really make sense.  Maybe `agent_look_angle` makes the most sense?

habitat/tasks/nav/nav_task.py
```diff
@@ -515,8 +543,21 @@ def update_metric(self, episode, action):
...
+        agent_state = self._sim.get_agent_state()
+        # quaternion is in x, y, z, w format
+        ref_rotation = agent_state.rotation
+
+        heading_vector = quaternion_rotate_vector(
+            ref_rotation.inverse(), np.array([0, 0, -1])
+        )
+
+        phi = cartesian_to_polar(-heading_vector[2], heading_vector[0])[1]
+        return np.array(phi) - np.pi / 2
```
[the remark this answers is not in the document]
erikwijmans: This accounts for us flipping x and y in the topdown map

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +199,155 @@ def main():
...
+        "'DEPTH_SENSOR' are supported",
+    )
+    parser.add_argument(
+        "--task-config",
+        type=str,
+        default="configs/tasks/pointnav.yaml",
+        help="path to config yaml containing information about task",
+    )
+    parser.add_argument(
+        "--video-option",
```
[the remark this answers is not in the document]
mathfac: Discussed offline. tensorboard dir can be used for other metrics reporting. Then maybe please use `nargs='+'` option for parser and remove `video_both` option.

habitat/tasks/nav/nav_task.py
mathfac: @erikwijmans name should encode that it's top down single angle.

habitat/tasks/nav/nav_task.py
```diff
@@ -393,10 +393,11 @@ def reset_metric(self, episode):
 
     def update_metric(self, episode, action):
         if self._metric is None:
-            self._metric = 0
-
+            self._metric = {"count": 0, "is_collision": False}
```
abhiskk: Using strings directly for keys is not idea, can you store the keys as some kind of class attribute, something like `self._collision_key = "is_collision"`

habitat/tasks/nav/nav_task.py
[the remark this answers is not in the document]
mathfac: I see, then I would go with something like this:
```
flip_x_y_back = - np.pi / 2
return np.array(phi) + flip_x_y_back
```

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +199,155 @@ def main():
...
+    assert (args.model_path is not None) != (
+        args.tracking_model_dir is not None
+    ), "Must specify a single model or a directory of models, but not both"
+    if args.video_option in ["video_tb", "video_both"]:
+        assert (
+            args.tensorboard_dir is not None
+        ), "Must specify a tensorboard directory for video display"
+
+    with (
+        tensorboard.SummaryWriter(
```
abhiskk: We should wrap the summary writer creation in a function call.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +200,153 @@ def main():
...
+        "--task-config",
+        type=str,
+        default="configs/tasks/pointnav.yaml",
+        help="path to config yaml containing information about task",
+    )
+    parser.add_argument(
+        "--video-option",
+        type=str,
+        default="no_video",
+        choices=["no_video", "video_dir", "video_tb", "video_both"],
```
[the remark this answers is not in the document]
JasonJiazhiZhang: https://github.com/facebookresearch/habitat-api/pull/127#discussion_r297868162
Will do this instead as it simplifies things.

habitat_baselines/evaluate_ppo.py
JasonJiazhiZhang: Will make a Wrapper class in a separate utils file.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +199,155 @@ def main():
...
+        args.tracking_model_dir is not None
+    ), "Must specify a single model or a directory of models, but not both"
+    if args.video_option in ["video_tb", "video_both"]:
+        assert (
+            args.tensorboard_dir is not None
+        ), "Must specify a tensorboard directory for video display"
+
+    with (
+        tensorboard.SummaryWriter(
+            log_dir=args.tensorboard_dir, purge_step=0, flush_secs=30
```
mathfac: @JasonJiazhiZhang, can we use avoid `with` keyword here and use `close()` method like this:
```
import torch
import torchvision
from torch.utils.tensorboard import SummaryWriter
from torchvision import datasets, transforms

# Writer will output to ./runs/ directory by default
writer = SummaryWriter()

transform = transforms.Compose([transforms.ToTensor(), transforms.Normalize((0.5,), (0.5,))])
trainset = datasets.MNIST('mnist_train', train=True, download=True, transform=transform)
trainloader = torch.utils.data.DataLoader(trainset, batch_size=64, shuffle=True)
model = torchvision.models.resnet50(False)
# Have ResNet model take in grayscale rather than RGB
model.conv1 = torch.nn.Conv2d(1, 64, kernel_size=7, stride=2, padding=3, bias=False)
images, labels = next(iter(trainloader))

grid = torchvision.utils.make_grid(images)
writer.add_image('images', grid, 0)
writer.add_graph(model, images)
writer.close()
```

habitat/tasks/nav/nav_task.py
JasonJiazhiZhang: Discussed privately, will name `agent_angle` for now

habitat/core/vector_env.py
```diff
@@ -27,6 +27,7 @@
 OBSERVATION_SPACE_COMMAND = "observation_space"
 ACTION_SPACE_COMMAND = "action_space"
 CALL_COMMAND = "call"
+EPISODE_COMMAND = "current_episode"
```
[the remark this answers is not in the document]
mathfac: @JasonJiazhiZhang I though we agreed on using `CALL_COMMAND`.

habitat/utils/visualizations/maps.py
```diff
@@ -334,3 +335,20 @@ def colorize_topdown_map(top_down_map: np.ndarray) -> np.ndarray:
...
     """
     return TOP_DOWN_MAP_COLORS[top_down_map]
+
+
+def draw_path(
+    top_down_map: np.ndarray,
+    path_points: List[Tuple],
+    color: int,
+    thickness: int = 2,
+):
```
mathfac:
```suggestion
) -> None:
```

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,45 +5,84 @@
...
-        help="comma separated string containing different"
-        "sensors to use, currently 'RGB_SENSOR' and"
-        "'DEPTH_SENSOR' are supported",
-    )
-    parser.add_argument(
-        "--task-config",
-        type=str,
-        default="configs/tasks/pointnav.yaml",
-        help="path to config yaml containing information about task",
+def poll_checkpoint_folder(checkpoint_folder, previous_ckpt_ind):
```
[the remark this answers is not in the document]
mathfac: Looks like we decided to move it to utils file.

habitat_baselines/evaluate_ppo.py
```diff
@@ -5,45 +5,80 @@
...
+        if "tensorboard" in args.video_option:
+            tb_writer.add_video_from_np_images(
+                f"episode{episode_id}", checkpoint_idx, images, fps=fps
+            )
+
+
+def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
     env_configs = []
     baseline_configs = []
+    device = torch.device("cuda:{}".format(args.pth_gpu_id))
```
mathfac:
```suggestion
    device = torch.device(f"cuda:{args.pth_gpu_id}")
```

habitat_baselines/evaluate_ppo.py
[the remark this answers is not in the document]
JasonJiazhiZhang: > Looks like we decided to move it to utils file.
There isn't an existing utils file that is an obvious choice. But will have a proposal of refactoring these utilities coming soon.

habitat/core/vector_env.py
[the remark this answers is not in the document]
JasonJiazhiZhang: Missed this one, will fix.

habitat_baselines/evaluate_ppo.py
```diff
@@ -149,13 +195,151 @@ def main():
...
+                rgb_frames = [rgb_frames[i] for i in state_index]
+
     episode_reward_mean = (episode_rewards / episode_counts).mean().item()
     episode_spl_mean = (episode_spls / episode_counts).mean().item()
     episode_success_mean = (episode_success / episode_counts).mean().item()
 
-    print("Average episode reward: {:.6f}".format(episode_reward_mean))
-    print("Average episode success: {:.6f}".format(episode_success_mean))
-    print("Average episode spl: {:.6f}".format(episode_spl_mean))
+    logger.info("Average episode reward: {:.6f}".format(episode_reward_mean))
```
mathfac: To deduplicate logging we even can add option`log=True` to `writer.add_scalars` method and log scalars from TensorBoard Wrapper.

habitat_baselines/tensorboard_utils.py
```diff
@@ -0,0 +1,52 @@
+import torch
+from torch.utils.tensorboard import SummaryWriter
+
+
+class TensorboardWriter(SummaryWriter):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def add_video_from_np_images(self, video_name, step_idx, images, fps=10):
```
mathfac: Please add typing to the method signature and string doc for the class.

habitat_baselines/evaluate_ppo.py
erikwijmans: Even better, `device = torch.device("cuda", args.pth_gpu_id)` :)

habitat_baselines/train_ppo.py
```diff
@@ -224,137 +225,172 @@ def run_training():
...
-            pth_time += time() - t_update_stats
+                count_steps += envs.num_envs
+                pth_time += time() - t_update_stats
 
-        if len(window_episode_reward) == args.reward_window_size:
-            window_episode_reward.popleft()
-            window_episode_counts.popleft()
-        window_episode_reward.append(episode_rewards.clone())
-        window_episode_counts.append(episode_counts.clone())
+            if len(window_episode_reward) == args.reward_window_size:
```
erikwijmans: You can get rid of this by adding `maxlen=args.reward_window_size` to the constructor of the `deque`

habitat_baselines/tensorboard_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        pass
+
+    def close(self):
+        pass
+
+    def __getattr__(self, item):
+        return lambda *args, **kwargs: None
+
+
+def get_tensorboard_writer(log_dir, *args, **kwargs):
```
mathfac:
```suggestion
def get_tensorboard_writer(log_dir:str, *args, **kwargs) -> Union[DummyWriter, TensorboardWriter]:
<doc string>
```

PR #139, opened 2019-06-24
Update Habitat-API to allow for no rendering sensors

erikwijmans:
## Motivation and Context

https://github.com/facebookresearch/habitat-sim/pull/85 adds the ability to load-up habitat-sim with no sensors.  This PR removes the shim from habitat-api that was needed when we didn't allow for no sensors!

## How Has This Been Tested

The no-sensor test passes :)

(It will fail on the CI until the corresponding PR on the sim is landed)

## Types of changes

### DIFF
--- a/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -189,13 +189,6 @@ class HabitatSim(Simulator):
             sim_sensor_cfg.sensor_type = sensor.sim_sensor_type  # type: ignore
             sensor_specifications.append(sim_sensor_cfg)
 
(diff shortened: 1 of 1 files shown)


mathfac:
CI is paying off. We have an assert about non-empty sensors specification on habitat-sim side:
https://github.com/facebookresearch/habitat-sim/blob/d383c2011bf1baab2ce7b3cd40aea573ad2ddf71/habitat_sim/simulator.py#L86

erikwijmans:
> (It will fail on the CI until the corresponding PR on the sim is landed)

https://github.com/facebookresearch/habitat-sim/pull/85 needs to be landed for this to pass

erikwijmans:
@mathfac -- Landed on sim side

Issue #149, opened 2019-07-02
Multiprocess evaluate_ppo.py broken


JasonJiazhiZhang:
Multiple process evaluation for PPO was introduced in PR #127, but is currently broken due to buggy update of `test_recurrent_hidden_state`. Current code for calculating `episode_rewards` and `episode_counts` is also buggy for multiprocess. 
For now please use single-process eval for ppo (with flag `--num-processes 1`)

Working on a fix

chenwydj:
is this issue solved? when I use `--num-processes 2` the multiprocess still raise BrokenPipeError

erikwijmans:
This was a bug in logic and didn't cause `BrokenPipeError`.  Please open a new issue for this and provide the additional details it asks for.

chenwydj:
> This was a bug in logic and didn't cause `BrokenPipeError`. Please open a new issue for this and provide the additional details it asks for.

Did you mean this is a logic bug in my side? What did you mean by "didn't cause `BrokenPipeError`"?

Thank you!

erikwijmans:
May be a logic error on your side, may be a new/different bug.  The bug reported here didn't cause a BrokenPipeError so I doubt this has anything to do with that.

Issue #148, opened 2019-07-01
Evaluation and pre-trained models


knzo25:
I am currently using commit 43ab7241610804dc971344fa4b7d2b674b6f7045 which is currently the master branch, but I also tried previous commits.

I want to replicate the results from the paper, but running the pre-trained methods provide an average success of 0% with a equally null SPL.

Here is the executed command
```
python -u habitat_baselines/evaluate_ppo.py \
    --model-path "/path/to/checkpoint" \
    --sim-gpu-id 0 \
    --pth-gpu-id 0 \
    --num-processes 1 \
    --count-test-episodes 994 \
    --task-config "configs/tasks/pointnav.yaml"
    --sensors RGB_SENSOR 
```

And an example of the output:
```
2019-07-01 18:54:56,249 Average episode reward: -0.460795
2019-07-01 18:54:56,249 Average episode success: 0.000000
2019-07-01 18:54:56,250 Average episode SPL: 0.000000
```

Additionally, since pulling the master branch, with the evaluation scheme, it sometimes fails when it executes the policy, giving the cryptic:

```
cuDNN error: CUDNN_STATUS_BAD_PARAM
```

Which in this case I think it is because it tries to execute the policy on an empty batch (due to how it does not track ended episodes). To avoid this, for now I just keep tracking ended episodes. I don't think this one last error is to blame for the null performance of the pre-trained models since I had this error before this scheme was implemented.

It would seem to me that along the commits since training these models the policy must have changed a little.

Thank you for the great platform and I appreciate any help !

erikwijmans:
For the null performance, this is due to changing which number actions are on.  You can use this script to fix the model weights (we will update the ones probably after refactoring the baselines code):  https://gist.github.com/erikwijmans/398a89d0b793cdb878a903b83ceefad3

> Which in this case I think it is because it tries to execute the policy on an empty batch (due to how it does not track ended episodes).

Yeah, we need to add a `or envs.num_envs > 0` to the while loop to keep that from happening :)

JasonJiazhiZhang:
> cuDNN error: CUDNN_STATUS_BAD_PARAM

For this error, it happens because CNN module took in a batch of size 0. This is because evaluate_ppo.py checks if episodes are seen and will return batch of size 0 if there's no  unseen episodes (based on episode_id).

After dataset creation, can you do a `print([ep.episode_id for ep in dataset.episodes])` to see if there are 994 unique episode_ids?

[...]

PR #150, opened 2019-07-03
fix multiprocess eval

JasonJiazhiZhang:
## Motivation and Context

Fix issue #149, now evaluate_ppo will correctly run and output stats with more than one processes (given that there are scenes to split on).
Improved stats recording : stats are handled in a per-episode way and aggregated at the end of the eval run. 
It is still recommended to use single process for eval with test_scenes since its val split currently does not support split-by-scene and all work will be done in one of the N processes. 
Thank you @erikwijmans for raising this issue.

Edit:
Also fix `cuDNN error: CUDNN_STATUS_BAD_PARAM` issue in #148. Now eval will work as long as there are enough episodes with unique `(scene_id, episode_id)` combo (but not necessarily unique episode_id). 

## How Has This Been Tested

All tests pass and evaluate_ppo won't break with num_processes > 1. 
## Types of changes

## Checklist

### DIFF
--- a/habitat_baselines/evaluate_ppo.py
+++ b/habitat_baselines/evaluate_ppo.py
@@ -188,7 +188,10 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
         envs_to_pause = []
         n_envs = envs.num_envs
         for i in range(n_envs):
-            if next_episodes[i].episode_id in stats_episodes:
+            if (
+                next_episodes[i].scene_id,
+                next_episodes[i].episode_id,
+            ) in stats_episodes:
                 envs_to_pause.append(i)
 
             # episode ended
@@ -200,7 +203,7 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
                 current_episode_reward[i] = 0
                 # use scene_id + episode_id as unique id for storing stats
                 stats_episodes[
-                    "{}:{}".format(
+                    (
                         current_episodes[i].scene_id,
                         current_episodes[i].episode_id,
                     )
--- a/habitat_baselines/evaluate_ppo.py
+++ b/habitat_baselines/evaluate_ppo.py
@@ -143,21 +143,20 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
     for sensor in batch:
         batch[sensor] = batch[sensor].to(device)
 
-    episode_rewards, episode_success, episode_counts = 0, 0, 0
     current_episode_reward = torch.zeros(envs.num_envs, 1, device=device)
 
     test_recurrent_hidden_states = torch.zeros(
         args.num_processes, args.hidden_size, device=device
     )
     not_done_masks = torch.zeros(args.num_processes, 1, device=device)
-    stats_episodes = dict()
+    stats_episodes = dict()  # dict of dicts that stores stats per episode
 
     rgb_frames = None
     if args.video_option:
         rgb_frames = [[]] * args.num_processes
         os.makedirs(args.video_dir, exist_ok=True)
 
-    while episode_counts < args.count_test_episodes:
+    while len(stats_episodes) < args.count_test_episodes:
         current_episodes = envs.current_episodes()
 
         with torch.no_grad():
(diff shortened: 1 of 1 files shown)


habitat_baselines/evaluate_ppo.py
```diff
@@ -207,7 +194,15 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
 
             # episode ended
             if not_done_masks[i].item() == 0:
-                stats_episodes.add(current_episodes[i].episode_id)
+                stats_episodes[current_episodes[i].episode_id] = infos[i][
```
erikwijmans: Let's just make this a dictionary of dictionaries.  Then you can keep both SPL and reward in the dictionary and you can use the length of the dictionary to keep track of the number of finished episodes.

habitat_baselines/evaluate_ppo.py
```diff
@@ -198,8 +198,13 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
                 episode_stats["success"] = int(infos[i]["spl"] > 0)
                 episode_stats["reward"] = current_episode_reward[i].item()
                 current_episode_reward[i] = 0
-                stats_episodes[current_episodes[i].episode_id] = episode_stats
-
+                # use scene_id + episode_id as unique id for storing stats
+                stats_episodes[
+                    "{}:{}".format(
```
erikwijmans: Tuples can be keys to dictionaries, that is probably cleaner :)

erikwijmans: Also the logic of deciding if an environment should be paused or not needs to use this key.

JasonJiazhiZhang: > Also the logic of deciding if an environment should be paused or not needs to use this key.

Whoops. Should be fixed now :)

habitat_baselines/evaluate_ppo.py
```diff
@@ -207,7 +193,18 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
...
             if not_done_masks[i].item() == 0:
-                stats_episodes.add(current_episodes[i].episode_id)
+                episode_stats = dict()
+                episode_stats["spl"] = infos[i]["spl"]
+                episode_stats["success"] = int(infos[i]["spl"] > 0)
+                episode_stats["reward"] = current_episode_reward[i].item()
+                current_episode_reward[i] = 0
+                # use scene_id + episode_id as unique id for storing stats
+                stats_episodes[
+                    "{}:{}".format(
```
mathfac: use f string please.

habitat_baselines/evaluate_ppo.py
```diff
@@ -184,30 +180,34 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
...
-
         next_episodes = envs.current_episodes()
         envs_to_pause = []
         n_envs = envs.num_envs
         for i in range(n_envs):
-            if next_episodes[i].episode_id in stats_episodes:
+            if (
+                next_episodes[i].scene_id,
+                next_episodes[i].episode_id,
+            ) in stats_episodes:
```
mathfac: As I understand we will send for already paused environments `pause` command again. I think that's ok, but just double checking.

JasonJiazhiZhang: I don't think it will. It will only track environments that's not paused, i.e. `envs.num_envs` decreases after pause.

erikwijmans: Yep, pause will only be sent at most once to each environment

habitat_baselines/evaluate_ppo.py
```diff
@@ -143,24 +143,20 @@ def eval_checkpoint(checkpoint_path, args, writer, cur_ckpt_idx=0):
...
-    stats_episodes = set()
+    stats_episodes = dict()  # dict of dicts that stores stats per episode
 
     rgb_frames = None
     if args.video_option:
         rgb_frames = [[]] * args.num_processes
         os.makedirs(args.video_dir, exist_ok=True)
 
-    while episode_counts.sum() < args.count_test_episodes:
+    while len(stats_episodes) < args.count_test_episodes:
```
erikwijmans: Probably makes sense to add 
```suggestion
    while len(stats_episodes) < args.count_test_episodes and envs.num_envs > 0:
```
?

habitat_baselines/evaluate_ppo.py
mathfac: Got it, we are removing env from active envs once it's paused.

PR #144, opened 2019-06-27
Refactor baselines to allow for extension

erikwijmans:
## Motivation and Context

The current PPO baselines code isn't easily extendible to other tasks, architectures, or variants of on-policy RL.  This PR does some refactoring to allow other architectures and variants of on-policy RL to be added and helps slightly with other tasks.

This PR also introduces support for N-layer GRUs and support for N-layer LSTMs as the state-encoder whereas the current code is limited to a single layer GRU.

This PR also adds support for making the state-encoder auto-regressive (i.e. taking the previous action as input).

I also made this script to remap any current checkpoints so they can be loaded: https://gist.github.com/erikwijmans/94f33461982ac9e3cc354fb26e4b80e1

## How Has This Been Tested

Via the tests and running `train_ppo.py`, and `evaluate_ppo.py`.  I also

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to change)

PR #153, opened 2019-07-09
Refactor baselines: add generic trainer and common utils

JasonJiazhiZhang:
## Looking for Suggestions
Looking for a good name that competently describes [this config node](https://github.com/JasonJiazhiZhang/habitat-api/blob/0a7ff0280dbeb7e4a4f108c83f4f0e9b7808e446/habitat_baselines/config/default.py#L24) and all child nodes: used to be `BASELINE` and now renamed to `TRAINER`. Alternatives: `LEARNER`, `EXP(ERIMENT)`, `MODEL(S)`. 
Suggestions are greatly appreciated!
## Motivation and Context

This PR aims to bring modularity and improve reusability of habitat_baseline,  by introducing abstraction to the current experiment process and collecting common utilities. This is an early draft and all suggestions are welcomed :)
Since the amount of code changed is relatively large, here's a breakdown of what does this PR do, what directions am I particularly looking for suggestions/opinions, and what needs to be done before merging.

* **What does this PR do?**
  1. Provide a trainer template (`BaseTrainer`) and  create trainer registry.  I went through a couple of related repos ([Horizon](https://github.com/facebookresearch/Horizon), [stable_baselines](https://github.com/hill-a/stable-baselines), [Pythia](https://github.com/facebookresearch/pythia)) and chose a high-level structure similar to [this](https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/base_class.py) in stable_baselines (suggestions on implementation details are welcomed! ). This will  make it more straightforward to add new trainers. Note that this is complementory to #144 , as that one is more specific to ppo and aims to make it extensible to new tasks. 
  2. Collect trainer-independent utilities, such as env creation, batch transpose, and tensorboard logging, and move them to `habitat_baselines/common/` to give better reusability. 
  3. Merge logics in train_ppo and evaluate_ppo into one `ppo_trainer.py` file. Now launch commands for training/evaluation look like this:
  ```bash
  python  habitat_baselines/run.py --exp-config /path/to/config/ppo.yaml --run-type train
  ```
  4. Unify env_config, baseline_config, and previous command-line args into one config system so that it can be passed and stored in a single place.
  Here's a directory tree that might help visualize the structural changes:
```diff
habitat_baselines/
 ├── __init__.py
+├── common/                      	# directory that includes reusable functionalities
+│   ├── base_trainer.py           	# trainer template
+│   ├── env_utils.py
+│   ├── environments.py                # customized RL environments
+│   ├── tensorboard_utils.py
+│   ├── baseline_registry.py           # registry for envs and trainers
+│   └── utils.py
 ├── config/
+│   └── pointnav/
+│       └── ppo.yaml 
 ├── rl/
 │   ├── __init__.py
 │   ├── ppo/
 │   │   ├── __init__.py
 │   │   ├── policy.py
 │   │   ├── ppo.py
+│   │   ├── ppo_trainer.py  		# trainer class for ppo (train and eval)
-│   │   └── utils.py
 │   └── requirements.txt
+├── run.py                             # main entry point to all experiments
-├── evaluate_ppo.py
-├── train_ppo.py
 └── ...
```

* **Things need to be delivered before merging:**
  - [x] Polishing, docs, typing, header  
  - [x] Fix imports in habitat_baselines `__init__` 
  - [ ] change README
* **TODO (maybe in future PRs):**
     - Use saved configs in ppo_trainer eval(), and make compatible with old checkpoints.
     - Merge with #144.
     - Reorganize the config system :|

## How Has This Been Tested
Current tests pass and train/eval run properly from `run.py`

## Types of changes

## Checklist


JasonJiazhiZhang:
For now this PR is expected to fail CI test.
This command below in  circle-ci `run api tests` will have to be modified accordingly for it to pass. 
```
python -u habitat_baselines/train_ppo.py  ...
```
Maybe change to something like this when PR lands:
```
python -u habitat_baselines/run.py --exp-config habitat_baselines/config/pointnav/ppo.yaml --run-type train
```

habitat_baselines/common/base_trainer.py
```diff
@@ -0,0 +1,48 @@
+from abc import ABC, abstractmethod
+
+from habitat_baselines.common.trainer_registry import trainer_registry
+
+
+class BaseTrainer(ABC):
```
abhiskk: To stay uniform with other places in the codebase, let's avoid inheriting from `ABC` and instead using, `raise NotImplementedError`. 

Note: The ABC design is definitely valid but it is something which the libraries from our domain don't use (eg: pytorch), hence `ABC` was not included in the original design.

habitat_baselines/common/tensorboard_utils.py
```diff
@@ -1,8 +1,30 @@
-from typing import Optional, Union
+from typing import Union
 
 import numpy as np
 import torch
-from torch.utils.tensorboard import SummaryWriter
+
+
+class DummyWriter:
```
abhiskk: We can use `SummaryWriter = None` instead of using a `DummyWriter` here and handle the logic later.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
+import random
+
+import numpy as np
+
+import habitat
+from habitat import Config, SimulatorActions, make_dataset
+
+
+class NavRLEnv(habitat.RLEnv):
```
abhiskk: Place `NavRLEnv` inside a separate file under `common`.

habitat_baselines/common/base_trainer.py
abhiskk: We should think about more useful abstract methods to add to the `BaseTrainer`, @JasonJiazhiZhang if possible can you communicate on the PR what are the implementations you referred to when building out the base trainer and what you eventually ended up using.

habitat_baselines/common/tensorboard_utils.py
JasonJiazhiZhang: I think I'll create a separate PR for this. This PR did not change any of the tensorboard-specific logic and inherited directly from #127. I do agree that DummyWriter is a bit clunky and if we properly check `None` when appropriate we can get rid of the wrapper altogether.

habitat_baselines/common/base_trainer.py
JasonJiazhiZhang: Sure, this draft PR is precisely for discussions like this. This BaseTrainer structure is inspired mainly by the base_class.py in [stable-baselines](https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/base_class.py). I will put all the sources I looked in the description.

habitat_baselines/__init__.py
```diff
@@ -3,3 +3,10 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+
+# TODO include useful/reusable classes/functions
+
+# import for proper registry
+from habitat_baselines.rl.ppo.ppo_trainer import PPOTrainer, RolloutStorage
+
+__all__ = ["PPOTrainer", "RolloutStorage"]
```
mathfac: Include general Trainer as well to make baseline extension easier.

habitat_baselines/common/env_utils.py
mathfac: Looks like we will need Registry for RLEnvs and call this PointGoalRLEnv. Also add meaningful comment how RLEnvs should be used.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        done = False
+        if self._env.episode_over or self._episode_success():
+            done = True
+        return done
+
+    def get_info(self, observations):
+        return self.habitat_env.get_metrics()
+
+
+def make_env_fn(config, env_class, rank):
```
mathfac: Duplication of code from utils.py

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+from habitat import Config, get_config, make_dataset
+from habitat_baselines.config.default import get_config as baseline_cfg
+
+
+class Flatten(nn.Module):
```
mathfac: All this classes need docs, when it should be used.

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+        "opts",
+        default=None,
+        nargs=argparse.REMAINDER,
+        help="Modify config options from command line",
+    )
+    return parser
+
+
+def get_exp_config(cfg_path, opts=None):
+    config = Config(new_allowed=True)
```
mathfac: How `new_allowed`  works?

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    config.SIMULATOR.SCENE = dataset.episodes[0].scene_id
+    config.freeze()
+    env = env_class(
+        config_env=config, config_baseline=config.BASELINE, dataset=dataset
+    )
+    env.seed(rank)
+    return env
+
+
+def batch_obs(observations):
```
mathfac: Maybe outside of this PR, this can be refactored to be more general function. Maybe we can re-use t[he structured array functionality from Numpy ](https://jakevdp.github.io/PythonDataScienceHandbook/02.09-structured-data-numpy.html#RecordArrays:-Structured-Arrays-with-a-Twist)or find analogy in Pytorch. Conversion of array of structures to structure of arrays is common pattern in Data science.

habitat_baselines/config/default.py
```diff
@@ -22,13 +22,51 @@
 # BASELINE
 # -----------------------------------------------------------------------------
 _C.BASELINE = CN()
+_C.BASELINE.TRAINER_NAME = "ppo"
```
mathfac:
```suggestion
_C.TRAINER.TYPE = "ppo"
```
Let's be consistent with habitat-api naming: https://github.com/facebookresearch/habitat-api/blob/master/habitat/config/default.py#L61. Maybe, even `BASELINE` -> `TRAINER`.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,738 @@
...
+            "eval_SPL", {"average SPL": episode_spl_mean}, cur_ckpt_idx
+        )
+        writer.add_scalars(
+            "eval_success",
+            {"average success": episode_success_mean},
+            cur_ckpt_idx,
+        )
+
+
+def generate_video(
```
mathfac: Maybe, move to ppo_uitils or etc all other classes/functions except PPOTrainer.  PPOTrainer already large enough.

habitat_baselines/run.py
```diff
@@ -0,0 +1,18 @@
+from habitat_baselines.common.base_trainer import get_trainer
+from habitat_baselines.common.utils import experiment_args, get_exp_config
+
+
+def main():
```
mathfac: I really like simplicity of the entry point and how easy we can contribute more baselines without messing other ones.

mathfac:
For CI, can you rebase on latest habitat-api changes? Thanks

habitat_baselines/common/utils.py
JasonJiazhiZhang: `new_allowed=True` allows merging configs even if they have non-overlapping config nodes. For example, merging baseline_config into env_config with `new_allowed=False` won't work since env_config by default doesn't have the `BASELINE` node.

habitat_baselines/common/utils.py
JasonJiazhiZhang: Already on TODO list :)

habitat_baselines/config/default.py
JasonJiazhiZhang: > Maybe, even BASELINE -> TRAINER

This is definitely something I would like to discuss. I'm thinking something like `BASELINE` -> `EXPERIMENT`.

mathfac:
Master branch CI is passing: https://circleci.com/gh/facebookresearch/habitat-api.

habitat_baselines/common/utils.py
mathfac: @JasonJiazhiZhang is that new yacs feature? It wasn't there before.

JasonJiazhiZhang: @mathfac I don't think it's new. See [here](https://github.com/rbgirshick/yacs/blob/aabe5bd54d193996ceedcff650cff10c38b1f90b/yacs/config.py#L74).

habitat_baselines/common/base_trainer.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+
+def get_trainer(trainer_name, trainer_cfg):
+    trainer = trainer_registry.get_trainer(trainer_name)
+    assert trainer is not None, f"{trainer_name} is not supported"
+    return trainer(trainer_cfg)
+
+
+class BaseRLTrainer(BaseTrainer):
```
apsdehal: Is there a point in having this when it is not adding any new functionality on top of `BaseTrainer`?

habitat_baselines/common/env_utils.py
apsdehal: Ideally the registry should support nesting, something like `registry.register_env('rl.point_nav', EnvClassHere)`.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
+import random
+
+import numpy as np
+
+import habitat
+from habitat import Config, SimulatorActions, make_dataset
+
+
+class NavRLEnv(habitat.RLEnv):
+    def __init__(self, config_env, config_baseline, dataset):
```
apsdehal: Comments here would be helpful to explain more about the parameters.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        self._config_env = config_env.TASK
+        self._config_baseline = config_baseline
+        self._previous_target_distance = None
+        self._previous_action = None
+        self._episode_distance_covered = None
+        super().__init__(config_env, dataset)
+
+    def reset(self):
+        self._previous_action = None
```
apsdehal: Does `self._episode_distance_covered` also need to be set to `None`. It probably won't affect anything but just in case.

habitat_baselines/common/env_utils.py
apsdehal: Also, it is unnatural to write it as `config_env` and `config_baseline` instead of `env_config` and `baseline_config`.

apsdehal: And instead of `baseline_config`, it should be `model_config` now.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        ]
+        return observations
+
+    def step(self, action):
+        self._previous_action = action
+        return super().step(action)
+
+    def get_reward_range(self):
+        return (
+            self._config_baseline.BASELINE.RL.SLACK_REWARD - 1.0,
```
apsdehal: Same here. These should be `MODEL` instead of `BASELINE`. I might be missing some context though.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+            self._config_baseline.BASELINE.RL.SUCCESS_REWARD + 1.0,
+        )
+
+    def get_reward(self, observations):
+        reward = self._config_baseline.BASELINE.RL.SLACK_REWARD
```
apsdehal: Comments.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    config.SIMULATOR.SCENE = dataset.episodes[0].scene_id
+    config.freeze()
+    env = env_class(config_env=config, config_baseline=config, dataset=dataset)
+    env.seed(rank)
+    return env
+
+
+def construct_envs(config: Config, env_class):
+
+    random.seed(config.SEED)
```
apsdehal: Seed only in one single place. This can create mess later.

habitat_baselines/common/env_utils.py
apsdehal: So, if I have to use my own model, I will have to create a config with `BASELINE` attribute in it?

habitat_baselines/common/env_utils.py
apsdehal: Some general code philosophy here: If you are using function such as `make_*` be consistent with that. In trainer above, you added `get_trainer`, either change it to `make_trainer` or change everything else to `get_*`

habitat_baselines/common/base_trainer.py
```diff
@@ -0,0 +1,48 @@
...
+    @abstractmethod
+    def save_checkpoint(self, file_name):
+        pass
+
+    @abstractmethod
+    def load_checkpoint(self):
+        pass
+
+
+def get_trainer(trainer_name, trainer_cfg):
```
apsdehal: At this point, i will suggest to move all of the `make_*` and `get_*` functions to a separate file in utils. They are all over the place now.

habitat_baselines/common/env_utils.py
apsdehal: Also, why this isn't single config instead of being `env_config` separately and `baseline_config` separately.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        assert len(scenes) >= baseline_cfg.num_processes, (
+            "reduce the number of processes as there "
+            "aren't enough number of scenes"
+        )
+        scene_split_size = int(
+            np.floor(len(scenes) / baseline_cfg.num_processes)
+        )
+
+    scene_splits = [[] for _ in range(baseline_cfg.num_processes)]
+    for j, s in enumerate(scenes):
```
apsdehal:
```suggestion
    for idx, scene in enumerate(scenes):
```

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        if len(scenes) > 0:
+            config_env.DATASET.CONTENT_SCENES = scene_splits[i]
+
+        config_env.SIMULATOR.HABITAT_SIM_V0.GPU_DEVICE_ID = (
+            baseline_cfg.sim_gpu_id
+        )
+
+        agent_sensors = baseline_cfg.sensors.strip().split(",")
+        for sensor in agent_sensors:
+            assert sensor in ["RGB_SENSOR", "DEPTH_SENSOR"]
```
apsdehal: Will you keep updating this list in future if something changes. This ideally should be somewhere in the default config.

habitat_baselines/common/trainer_registry.py
```diff
@@ -0,0 +1,27 @@
+from typing import Optional
+
+from habitat.core.registry import Registry
+
+
+class TrainRegistry(Registry):
```
apsdehal: Why create a new registry class? Whole point of Registry pattern is to create a singleton which is easy to maintain instead of multiple sources of global truth. I would suggest using better functionalities in Registry itself to encapsulate everything there. For reference, check Pythia's registry again.

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+        nn.init.orthogonal_(self.linear.weight, gain=0.01)
+        nn.init.constant_(self.linear.bias, 0)
+
+    def forward(self, x):
+        x = self.linear(x)
+        return CustomFixedCategorical(logits=x)
+
+
+def _flatten_helper(t, n, tensor):
```
apsdehal: Wasn't this define above also? And why do we have `Flatten` as well as `_flatten_helper`. A single class should be enough to handle both of the usages.

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    return tensor.view(t * n, *tensor.size()[2:])
+
+
+def update_linear_schedule(optimizer, epoch, total_num_epochs, initial_lr):
```
apsdehal: Change this to a class `LRScheduler.

habitat_baselines/common/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    print(config)
+    config.merge_from_other_cfg(get_config(config.BASELINE.RL.PPO.task_config))
+    if opts is not None:
+        config.merge_from_list(opts)
+    return config
+
+
+def make_env_fn(config, env_class, rank):
+    dataset = make_dataset(config.DATASET.TYPE, config=config.DATASET)
+    config.defrost()
```
apsdehal: This is a comment on older habitat design. Seeing defrost and freeze being repeatedly and exhaustively everywhere has defeated the whole purpose of `freeze`. It is time to reconsider that.

habitat_baselines/common/base_trainer.py
JasonJiazhiZhang: @apsdehal The rationale is that `common/base_trainer.py` should in the future support imitation learning and SLAM based agents and `BaseTrainer` would be a template for those. I do think we need more discussion on what functionalities to be included inside `BaseTrainer` vs. `BaseRLTrainer`

habitat_baselines/common/env_utils.py
JasonJiazhiZhang: @apsdehal This naming choices are largely inherited from the original [`train_ppo.py`](https://github.com/facebookresearch/habitat-api/blob/782f1f9aec09895dc27347927f01998ba383b8b4/habitat_baselines/train_ppo.py#L25). I do have some alternatives in mind (see [comments here](https://github.com/facebookresearch/habitat-api/pull/153#discussion_r302721173)). Also, as to why it's not a single config yet: I definitely intend to merge two configs (see [here](https://github.com/JasonJiazhiZhang/habitat-api/blob/f6478167b3e46b897e4b7b1158b265b63a3b6c31/habitat_baselines/common/env_utils.py#L76) and [here](https://github.com/JasonJiazhiZhang/habitat-api/blob/f6478167b3e46b897e4b7b1158b265b63a3b6c31/habitat_baselines/common/utils.py#L90-L97), comparing to [this](https://github.com/JasonJiazhiZhang/habitat-api/blob/782f1f9aec09895dc27347927f01998ba383b8b4/habitat_baselines/train_ppo.py#L92)), and should modify NavRLEnv for use one config for sure. Thank you for catching this 👍 .  I'll make those changes to NavRLEnv to  take one unified config, if people agree with this change.

habitat_baselines/common/env_utils.py
JasonJiazhiZhang: I think setting to None makes sense. @abhiskk @mathfac did you guys choose to not reset it on purpose?

habitat_baselines/common/env_utils.py
JasonJiazhiZhang: If someone implements another algorithm, say A2C, they will create algorithm-specific config under `BASELINE.RL.A2C`. In this particular case tho, `SLACK_REWARD` is not algorithm-specific and is more related to the task and environment, so it lives under `BASELINE.RL`. Then again, it's probably a good idea to have a more informative/representative name than **BASELINE**

habitat_baselines/common/utils.py
JasonJiazhiZhang: Looks like another design choice inherited from old code that I did not pay much attention when refactoring. @abhiskk and @mathfac maybe this is something I can clean up?

habitat_baselines/common/env_utils.py
JasonJiazhiZhang: `make_env_fn` will be removed in a later PR as datasets will support scene split. Will stick to `get_*`

habitat_baselines/common/utils.py
erikwijmans: The `_flatten_helper` is a helper function for creating the batches to learn from the rollout.  It's very specific to that class and probably belongs as a static method on the class.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,713 @@
...
+            "eval_SPL", {"average SPL": episode_spl_mean}, cur_ckpt_idx
+        )
+        writer.add_scalars(
+            "eval_success",
+            {"average success": episode_success_mean},
+            cur_ckpt_idx,
+        )
+
+
+class RolloutStorage:
```
erikwijmans: The RolloutStorage isn't specific to PPO and should be in a general utils.

JasonJiazhiZhang: @erikwijmans I'll put it under `habitat_baselines/common/utils.py` for now and make `_flatten_helper` a static class method. Thanks!

mathfac:
CI test `python -u habitat_baselines/run.py --exp-config habitat_baselines/config/pointnav/ppo_train_test.yaml --run-type train` is trying to load checkpoint:
`FileNotFoundError: [Errno 2] No such file or directory: 'data/checkpoints/ckpt.0.pth'`

JasonJiazhiZhang:
> `FileNotFoundError: [Errno 2] No such file or directory: 'data/checkpoints/ckpt.0.pth'`
Looks like I didn't check for dir existence, thanks for spotting that.

habitat_baselines/config/pointnav/ppo.yaml
```diff
@@ -0,0 +1,40 @@
...
+      max_grad_norm: 0.5
+      num_steps: 128
+      hidden_size: 512
+      use_gae: True
+      gamma: 0.99
+      tau: 0.95
+      use_linear_clip_decay: True
+      use_linear_lr_decay: True
+      reward_window_size: 50
+      sensors: "RGB_SENSOR,DEPTH_SENSOR"
```
erikwijmans: This should be a list

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+            for update in range(ppo_cfg.num_updates):
+                if ppo_cfg.use_linear_lr_decay:
+                    update_linear_schedule(
+                        self.agent.optimizer,
+                        update,
+                        ppo_cfg.num_updates,
+                        ppo_cfg.lr,
+                    )
+
+                self.agent.clip_param = ppo_cfg.clip_param * (
```
erikwijmans: This needs an `if ppo_cfg.use_linear_clip_decay` around it.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+            config.TASK.MEASUREMENTS.append("TOP_DOWN_MAP")
+            config.TASK.MEASUREMENTS.append("COLLISIONS")
+        config.freeze()
+
+        logger.info(f"env config: {config}")
+        self.envs = construct_envs(config, NavRLEnv)
+
+        ckpt = torch.load(checkpoint_path, map_location=device)
+
+        actor_critic = Policy(
```
erikwijmans: Use `self._setup` here?

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+    batch_obs,
+    generate_video,
+    poll_checkpoint_folder,
+    update_linear_schedule,
+)
+from habitat_baselines.rl.ppo import PPO, Policy
+
+
+@baseline_registry.register_trainer(name="ppo")
+class PPOTrainer(BaseRLTrainer):
```
erikwijmans: Since the actually update rule (in this case PPO) is in a separate class, the vast majority of this trainer is really as `ActorCriticTrainer` and more general than just PPO.

habitat_baselines/config/pointnav/ppo.yaml
JasonJiazhiZhang: I think this not being a list originally is to support the command line option augmentation [feature](https://github.com/facebookresearch/habitat-api/blob/732377878c8c4b2503358e184bac4b17f2a8937e/habitat/config/default.py#L188). I kept the feature to add config option via command line so maybe a string is better suited here.

habitat_baselines/rl/ppo/ppo_trainer.py
JasonJiazhiZhang: Thanks for spotting that. I'll actually load saved config and merge them with config specified on evaluation.

habitat_baselines/rl/ppo/ppo_trainer.py
JasonJiazhiZhang: This definitely makes sense. I can make a separate PR for this structural change.

habitat_baselines/config/pointnav/ppo.yaml
erikwijmans: The `video_option` field is a list tho?  Is that not able to be changed and should be a comma separated string?

JasonJiazhiZhang: I think that should also be a string instead for consistency. Will fix :)

habitat_baselines/common/base_trainer.py
```diff
@@ -0,0 +1,55 @@
...
+class BaseTrainer:
+    """
+    Most generic trainer class that serves as a base template for more
+    specific trainer classes like RL trainer, SLAM or imitation learner.
+    Includes only the most basic functionality.
+    """
+
+    supported_tasks: ClassVar[List[str]]
+
+    def __init__(self):
```
abhiskk: You don't need to write out the constructor explicitly.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    Create VectorEnv object with specified config and env class type.
+    To allow better performance, dataset are split into small ones for
+    each individual env, grouped by scenes.
+    Args:
+        config: configs that contain num_processes as well as information
+        necessary to create individual environments.
+        env_class: class type of the envs to be created.
+
+    Returns:
+        vectorEnv object created according to specification.
```
abhiskk: nit: V should be capital

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    scenes = dataset.get_scenes_to_load(config.DATASET)
+
+    if len(scenes) > 0:
+        random.shuffle(scenes)
+
+        assert len(scenes) >= baseline_cfg.num_processes, (
+            "reduce the number of processes as there "
+            "aren't enough number of scenes"
+        )
+        scene_split_size = int(
```
abhiskk: This variable is no longer needed

habitat_baselines/common/tensorboard_utils.py
abhiskk: Can you create a TODO for above.

habitat_baselines/config/pointnav/ppo.yaml
```diff
@@ -0,0 +1,40 @@
...
+      checkpoint_folder: "data/new_checkpoints"
+      task_config: "configs/tasks/pointnav.yaml"
+      num_updates: 10000
+      # eval specific:
+      count_test_episodes: 2
+      video_option: ["disk", "tensorboard"]
+      video_dir: "video_dir"
+      model_path: "data/checkpoints/ckpt.123.pth"
+      tracking_model_dir: ""
+      # 1 of the above 2 options needs to be empty
```
abhiskk: Is there an explicit check for this somewhere.

habitat_baselines/common/utils.py
```diff
@@ -5,12 +5,60 @@
 # LICENSE file in the root directory of this source tree.
```
abhiskk: This file has become overloaded with a significant amount of code, distributing it even in a simple way across files will help in keeping the modularity. This can be done in this PR or can be added as a TODO for a later PR, @JasonJiazhiZhang  you can take a call on this.

habitat_baselines/common/utils.py
abhiskk: @JasonJiazhiZhang  please add a TODO around this, it is a good suggestion.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
+import os
```
abhiskk: This is also another file which is really heavy in terms of code size, I have made suggestions below to abstract out the logics to other modules.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+            aggregated_stats[stat_key] = sum(
+                [v[stat_key] for v in stats_episodes.values()]
+            )
+        num_episodes = len(stats_episodes)
+
+        episode_reward_mean = aggregated_stats["reward"] / num_episodes
+        episode_spl_mean = aggregated_stats["spl"] / num_episodes
+        episode_success_mean = aggregated_stats["success"] / num_episodes
+
+        logger.info(
```
abhiskk: The logging and writer accesses below should be abstracted out to a separate util method.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+                    prev_ckpt_ind += 1
+                    self._eval_checkpoint(
+                        current_ckpt,
+                        ppo_cfg,
+                        writer,
+                        cur_ckpt_idx=prev_ckpt_ind,
+                    )
+
+    def _eval_checkpoint(self, checkpoint_path, args, writer, cur_ckpt_idx=0):
+        device = torch.device("cuda", args.pth_gpu_id)
```
abhiskk: This function needs to be broken down into separate modular methods.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+            os.path.join(
+                self.config.TRAINER.RL.PPO.checkpoint_folder, file_name
+            ),
+        )
+
+    def load_checkpoint(self, checkpoint_path, *args, **kwargs) -> Dict:
+        return torch.load(checkpoint_path, map_location=self.device)
+
+    def train(self):
+        assert (
```
abhiskk: The train method for PPO is really heavy, we should abstract out logic present inside this class to separate methods.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+            self.config is not None
+        ), "trainer is not properly initialized, need to specify config file"
+
+        ppo_cfg = self.config.TRAINER.RL.PPO
+        logger.info(
```
abhiskk: We should abstract out the logging logic to a single function call, having it strewn across the train method makes things messy.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,519 @@
...
+                        update,
+                        ppo_cfg.num_updates,
+                        ppo_cfg.lr,
+                    )
+
+                self.agent.clip_param = ppo_cfg.clip_param * (
+                    1 - update / ppo_cfg.num_updates
+                )
+
+                for step in range(ppo_cfg.num_steps):
```
abhiskk: The stepping logic below should be abstracted out across function calls.

habitat_baselines/config/pointnav/ppo.yaml
JasonJiazhiZhang: I did a check [here](https://github.com/JasonJiazhiZhang/habitat-api/blob/c672710abfa3c58cce086ff623afb92596f7f47f/habitat_baselines/rl/ppo/ppo_trainer.py#L291-L293).

habitat_baselines/common/utils.py
JasonJiazhiZhang: Let's do this in a later PR!

habitat_baselines/rl/ppo/ppo_trainer.py
JasonJiazhiZhang: Discussed offline. Will make more abstraction to `ppo_trainer.py`in a separate PR, stuff like moving ppo update and rollout stage out of `train()`

habitat_baselines/config/pointnav/ppo.yaml
mathfac: Agree with Erik it should be list, but is in scope of next RL refactoring PR.

habitat_baselines/rl/ppo/ppo_trainer.py
mathfac: Agree with @abhiskk and @JasonJiazhiZhang need refactoring in next PR.

JasonJiazhiZhang:
Updated PR: fixing merge conflict and modified README.
Changes from @abhiskk 's suggestion:
  * RolloutStorage now is in it's own file

Changes from @mathfac 's comments:
  * Now trainer config won't be passed into envs as a child config node.
  * Saved checkpoint config will be loaded to provide more info at evaluation time. 
  * Merging of saved config and eval config now has proper order: `eval_opts > saved_opts > eval_base_config > saved_base_config`.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,104 @@
...
+        config: configs that contain num_processes as well as information
+        necessary to create individual environments.
+        env_class: class type of the envs to be created.
+
+    Returns:
+        VectorEnv object created according to specification.
+    """
+    trainer_config = config.TRAINER.RL.PPO
+    rl_env_config = config.TRAINER.RL
+    config = config.TASK_CONFIG  # exclude trainer-specific configs
```
mathfac: Usually one variable is used for one meaning, re-assign config isn't good idea, as it can have totally different meaning (experiment/baseline config and task config.) depends on code line.

habitat_baselines/common/env_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    dataset = make_dataset(config.DATASET.TYPE, config=config.DATASET)
+    config.defrost()
+    config.SIMULATOR.SCENE = dataset.episodes[0].scene_id
+    config.freeze()
+    env = env_class(config_env=config, config_baseline=config, dataset=dataset)
+    env.seed(rank)
+    return env
+
+
+def construct_envs(config: Config, env_class: Type) -> VectorEnv:
```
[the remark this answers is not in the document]
mathfac: Above comment is still relevant.

habitat_baselines/config/default.py
```diff
@@ -17,48 +17,87 @@
 # Config definition
 # -----------------------------------------------------------------------------
 _C = CN()
-_C.SEED = 100
+_C.BASE_TASK_CONFIG_PATH = "configs/tasks/pointnav.yaml"
+_C.TASK_CONFIG = get_config()  # task_config will be stored as a config node
```
mathfac: Task config will be overwritten with default config otherwise:
```suggestion
_C.TASK_CONFIG = CN()  # task_config will be stored as a config node
```

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -0,0 +1,568 @@
+import os
```
mathfac: Add Facebook headers.

habitat_baselines/run.py
```diff
@@ -0,0 +1,67 @@
...
+        opts: list additional options or options to be overwritten.
+
+    Returns:
+        config object created.
+    """
+
+    config = Config(new_allowed=True)
+    config.merge_from_other_cfg(baseline_cfg(cfg_path))
+    task_config = get_config(config.BASE_TASK_CONFIG_PATH)
+    config.TASK_CONFIG = task_config
```
mathfac: You kill `config.TASK_CONFIG` node that was defined before. I think we discussed opposite:
`config.TASK_CONFIG.merge(task_config)`.

habitat_baselines/run.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import List
+
+import numpy as np
+
+from habitat import Config, get_config
+from habitat_baselines.common.utils import get_trainer
+from habitat_baselines.config.default import get_config as baseline_cfg
+
+
+def get_exp_config(cfg_path: str, opts: List[str] = None) -> Config:
```
mathfac: Why not to replace `get_config` for baselines with this function?

habitat_baselines/run.py
JasonJiazhiZhang: Maybe `config.TASK_CONFIG` should take more priority in merging than base `task config`? Like:
```
config.TASK_CONFIG = task_config.merge(config.TASK_CONFIG)
```

habitat_baselines/config/default.py
JasonJiazhiZhang: I agree, this makes the behavior more obvious for user. But this does prohibit user from overwrite task_config in `ppo.yaml`

habitat_baselines/config/pointnav/ppo.yaml
apsdehal: Command line arguments from yacs support list since it does literal eval on what you pass. So this can be converted to an array.

JasonJiazhiZhang: @apsdehal Thanks for pointing out! I didn't test `option '["a","b"]'` and turns out that works for literal eval. Updated [here](https://github.com/facebookresearch/habitat-api/pull/172/commits/bba7ac81591d447eb31be2623e6c69b9fcec6840) in #172.

PR #156, opened 2019-07-11
Dataset change: envs access episodes via iterator

JasonJiazhiZhang:
## Motivation and Context
This PR makes environments access episodes via an iterator instead of keeping a list of episodes, and is the first a series of Dataset improvements. The main motivation is that we want to customize how episodes are iterated by environment, and this should be handled by a iterator of Dataset class instead of the Env class. 
Now we can modify the iterator logic to change how envs iterate through episodes. It's also a little bit less costly for environments to keep one iterator plus one episode than a list of episodes.  

Potential use cases of such iterator (that will be supported later): 1.  shuffle episodes when all episodes have been trained once. 2. switch to episodes of different scenes when an environment is trained on episodes from the same scene for too long to prevent overfitting.

Roadmap of Dataset update:
1. Make envs take iterator. (this PR)
2. Add options for creating iterator (scene switching, shuffling, seeding).
3. (Remove environments' access to episode list, left intact for now for backward compatibility).
4. Make PointNav dataset detect scenes with `\scene-datasets` (current must have `content_scene_folder`).
4. Make dataset object splittable by scene, and replace [this logic](https://github.com/facebookresearch/habitat-api/blob/782f1f9aec09895dc27347927f01998ba383b8b4/habitat_baselines/train_ppo.py#L110-L134).

## How Has This Been Tested
All tests pass/skip. Some tests are modified to use iterator instead of list of episodes. One new test of iterator is added. 

## Types of changes

## Checklist

### DIFF
--- a/habitat/core/dataset.py
+++ b/habitat/core/dataset.py
@@ -10,6 +10,7 @@ of a ``habitat.Agent`` inside ``habitat.Env``.
 """
 import copy
 import json
+from itertools import cycle
 from typing import Callable, Dict, Generic, List, Optional, Type, TypeVar
 
 import attr
@@ -88,6 +89,16 @@ class Dataset(Generic[T]):
             list of episodes corresponding to indexes.
         """
         return [self.episodes[episode_id] for episode_id in indexes]
+
+    def get_episode_iterator(self):
+        r"""
+        Creates and returns an iterator that iterates through self.episodes
+        in the desirable way specified.
+        Returns:
+            iterator for episodes
+        """
+        # TODO: support shuffling between epoch and  scene switching
+        return cycle(self.episodes)
 
     def to_json(self) -> str:
         class DatasetJSONEncoder(json.JSONEncoder):
--- a/habitat/core/env.py
+++ b/habitat/core/env.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import time
-from typing import Any, Dict, List, Optional, Tuple, Type
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Type
 
 import gym
 import numpy as np
@@ -46,6 +46,8 @@ class Env:
     _dataset: Optional[Dataset]
     _episodes: List[Type[Episode]]
     _current_episode_index: Optional[int]
+    _current_episode: Optional[Type[Episode]]
+    _episode_iterator: Optional[Iterator]
     _sim: Simulator
     _task: EmbodiedTask
     _max_episode_seconds: int
@@ -69,6 +71,8 @@ class Env:
                 id_dataset=config.DATASET.TYPE, config=config.DATASET
             )
         self._episodes = self._dataset.episodes if self._dataset else []
+        self._current_episode = None
+        self._episode_iterator = self._dataset.get_episode_iterator()
 
         # load the first scene if dataset is present
         if self._dataset:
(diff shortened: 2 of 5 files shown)


habitat/core/dataset.py
```diff
@@ -89,6 +90,16 @@ def get_episodes(self, indexes: List[int]) -> List[T]:
         """
         return [self.episodes[episode_id] for episode_id in indexes]
 
+    def get_episode_iterator(self):
+        r"""
+        Creates and returns an iterator that iterates through self.episodes
+        in the desirable way specified.
```
apsdehal: What are the ways?

JasonJiazhiZhang: Basically there are two demanded features: one is shuffling before looping back after all episodes have been exhausted. The other one is switching scenes when episodes from the same scene is loaded too many times on a simulator, to prevent overfitting to that scene. Those features will be supported very soon (but maybe on another PR after this high-level change has been approved). For now the iterator is a simple infinite-loop over episodes.

JasonJiazhiZhang:
What future iterator features might look like: [link](https://gist.github.com/JasonJiazhiZhang/b3b6e6d0041475f49f046aa73d2955a0)

test/test_sensors.py
```diff
@@ -25,15 +25,17 @@ def _random_episode(env, config):
...
-        NavigationEpisode(
-            episode_id="0",
-            scene_id=config.SIMULATOR.SCENE,
-            start_position=random_location,
-            start_rotation=random_rotation,
-            goals=[],
-        )
-    ]
+    env.episode_iterator = iter(
+        [
```
abhiskk: If possible, you can skip creating the list here and just use a basic generator.

JasonJiazhiZhang: @abhiskk  for now I think `iter([Episode])` is straightforward enough and requires minimal changes to test scripts. An alternative I can think of is maybe `(Episode for _ in range(1))`, which isn't much better IMO. I'll leave this unchanged if you agree.

I think for longer list of episodes, it's definitely worth it to create generator instead. And I can add a helper function to create such generator when I add tests to new features in next PR.

habitat/core/dataset.py
```diff
@@ -89,6 +90,16 @@ def get_episodes(self, indexes: List[int]) -> List[T]:
         """
         return [self.episodes[episode_id] for episode_id in indexes]
 
+    def get_episode_iterator(self):
```
mathfac: Maybe name it `get_cycle_episode_iterator` and add more method later with more sophisticated functionality.

habitat/core/env.py
```diff
@@ -46,6 +46,8 @@ class Env:
     _dataset: Optional[Dataset]
     _episodes: List[Type[Episode]]
     _current_episode_index: Optional[int]
```
mathfac: We should start adding DEPRECATED comment to fields that will be deprecated.

habitat/core/dataset.py
JasonJiazhiZhang: Actually, I plan to change `episode_iterator` a class property, and read functionality options from config file (added configs for this). Since a new PR will happen very soon, I think I'll leave this as is for now if you agree @mathfac .

test/test_sensors.py
abhiskk: this is fine, my mistake.

PR #160, opened 2019-07-17
Fog of war

erikwijmans:
## Motivation and Context

Adds fog-of-war revealing to the top-down map!

## How Has This Been Tested

With the shortest path follower example.

trajectory

## Types of changes


habitat/tasks/nav/nav_task.py
```diff
@@ -584,6 +611,18 @@ def update_map(self, agent_position):
                 thickness=thickness,
             )
 
+        if self._config.FOG_OF_WAR.DRAW:
```
mathfac: All this can be moved to method `update_fog_of_war(self)`

habitat/utils/visualizations/fog_of_war.py
```diff
@@ -0,0 +1,162 @@
...
+            * np.array(
+                [np.cos(current_angle + angle), np.sin(current_angle + angle)]
+            ),
+        )
+
+
+def reveal_fog_of_war(
+    top_down_map: np.ndarray,
+    current_fog_of_war_mask: np.ndarray,
+    current_pt: np.ndarray,
```
mathfac:
```suggestion
    current_position: np.ndarray,
```

examples/shortest_path_follower_example.py
```diff
@@ -35,7 +35,9 @@ def get_info(self, observations):
 
 
 def draw_top_down_map(info, heading, output_size):
-    top_down_map = maps.colorize_topdown_map(info["top_down_map"]["map"])
+    top_down_map = maps.colorize_topdown_map(
+        info["top_down_map"]["map"], info["top_down_map"]["fog_of_war_mask"]
```
mathfac: As I understand functionality be tested as fog became enabled by default. Curious if it doesn't fail when `TASK.TOP_DOWN_MAP.FOG_OF_WAR.DRAW = FALSE`.

erikwijmans: `info["top_down_map"]["fog_of_war_mask"]` is `None` when `TASK.TOP_DOWN_MAP.FOG_OF_WAR.DRAW = FALSE` and the logic in `maps.colorize_topdown_map` defaults to the old behavior when the fog-of-war mask is None, so it works :) (I have also tested this).

dhruvbatra:
This is awesome @erikwijmans. 

Sorry for giving feedback on a merged commit (we move so fast!), but can we make the unexplored areas a bit darker (e.g. by changing the alpha-blending weights), so the explored vs unexplored contrast is higher?

JasonJiazhiZhang:
> can we make the unexplored areas a bit darker (e.g. by changing the alpha-blending weights), so the explored vs unexplored contrast is higher?

Just to add to that: if we are using dark grey as unexplored region and alpha-blending white hue, maybe we should make unexplorable/unreachable region black (or some other dark color) instead of white. This way it's easy to make explored area even whiter (higher contrast with dark grey unexplored area) without causing confusion. Plus this is what FOW in games does. This will change the color coding greatly though, so just an idea 😃 .

erikwijmans:
Made this PR to make things darker: https://github.com/facebookresearch/habitat-api/pull/166

PR #161, opened 2019-07-17
Added ccache to CI and fixed PyTorch caching

mathfac:
- fixed PyTorch caching
- added ccache to CI
- broke habitat api step on installation and testing to save extra 5 min on saving cache in parallel
The testing time drops from first run to second run from **17.5 minutes to 9 minutes (2x times)** for full testing.

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ jobs:
                 export PATH=$HOME/miniconda/bin:$PATH
                 conda create -y -n habitat python=3.6
                 . activate habitat
-                conda install -q -y -c conda-forge ninja numpy pytest
+                conda install -q -y -c conda-forge ninja ccache numpy pytest
               fi
       - run:
           name: Install pytorch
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -127,11 +140,13 @@ jobs:
                   python setup.py install --headless
                 fi
               fi
-      - save_cache:
-          key: conda-{{ checksum "habitat-api/.circleci/config.yml" }}
-          background: true
-          paths:
-            - ~/miniconda
+      - run:
+          name: ccache stats
```
apsdehal:
```suggestion
          name: CCache Stats
```

.circleci/config.yml
```diff
@@ -107,6 +107,19 @@ jobs:
       - restore_cache:
           keys:
             - habitat-sim-{{ checksum "./hsim_sha" }}
+      - restore_cache:
+          keys:
+            - ccache-{{ arch }}-{{ .Branch }}-{{ epoch }}
```
erikwijmans: Won't this not work as `{{ epoch }}` is the current time and the cache-key will forever be different?  Since this is the API, probably `ccache-{{ arch }}-{{ checksum "./hsim_sha" }}` makes sense as the key.

.circleci/config.yml
```diff
@@ -152,8 +167,14 @@ jobs:
               python examples/example.py --scene data/scene_datasets/habitat-test-scenes/van-gogh-room.glb --silent --test_fps_regression $FPS_THRESHOLD
       - save_cache:
           key: habitat-sim-{{ checksum "./hsim_sha" }}
+          background: true
           paths:
             - ./habitat-sim
+      - save_cache:
+          key: ccache-{{ arch }}-{{ .Branch }}-{{ epoch }}
```
erikwijmans: Same comment about the key.

.circleci/config.yml
erikwijmans: Actually, it should just be `ccache-{{ arch }}` as the install of the simulator is already cached.

PR #163, opened 2019-07-17
Episode-iterator feature: scene switching and shuffling

JasonJiazhiZhang:
## Motivation and Context
This PR is a follow up to #156 that implements the proposed features to episode iterator. Features include:
1. shuffle episodes when all episodes have been trained once. 
2. switch to episodes of different scenes when an environment is trained on episodes from the same scene for too long to prevent overfitting.
3. option to turn off episode cycling when all episodes are iterated once
4. sample episodes (refactored from `Dataset.__init__`)

Impact on performance (both in terms of preventing overfitting and reducing  simulator loading overhead) is __NOT__ tested. However it's expected to be strictly positive.

## How Has This Been Tested
New test added to test these features. All tests pass. 

## Types of changes

## Checklist


habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+        self.episodes = episodes
+        self.cycle = cycle
+        self.group_by_scene = group_by_scene
+        if group_by_scene:
+            num_scene_groups = len(
+                list(groupby(episodes, key=lambda x: x.scene_id))
+            )
+            num_unique_scenes = len(set([e.scene_id for e in episodes]))
+            if num_scene_groups >= num_unique_scenes:
+                self.episodes = sorted(self.episodes, key=lambda x: x.scene_id)
```
erikwijmans: Probably makes sense to always sort.  The only times when you'd save any time is when `len(self.episodes)` is short enough that sorting doesn't take meaningful time.  In the general case where you need to sort, the checks to see if you need to sort will take non-trivial time.

Also, the sorting algo python uses is `O(n)` when the list is already sorted, so there isn't even any savings there!

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+        Sample episodes: sample the specified number of episodes.
+    """
+
+    def __init__(
+        self,
+        episodes: List[T],
+        cycle: bool = True,
+        shuffle_scene_when_cycle: bool = False,
+        group_by_scene: bool = True,
+        max_scene_repeat: int = -1,
```
erikwijmans: IMO it makes more sense to re-shuffle the list of episodes by scene once every `k` actions taken.  This is logic that would have to live in the `Env` class tho.

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+        Returns:
+            next episode.
+        """
+        try:
+            next_episode = next(self._iterator)
+        except StopIteration:
+            if not self.cycle:
+                raise StopIteration
+            self._iterator = iter(self.episodes)
+            if self.shuffle_scene_when_cycle:
```
erikwijmans: Should this go before `self._iterator = iter(self.episodes)`?

JasonJiazhiZhang: `self._shuffle_iterator_scene()` ideals with iterator directly. The main reason behind that is that I can then use `groupby` function in `itertools`, and I can use this `self._shuffle_iterator_scene()` method for scene switching as well.

habitat/core/dataset.py
JasonJiazhiZhang: Maybe `_shuffle_iterator_scene()` should not be an internal helper method then. Per-k-actions shuffling option is not hard to do, though I need to feature out a good way to encode these options and make them configurable.

habitat/core/dataset.py
JasonJiazhiZhang: @erikwijmans one scenario that might be worth considering: episodes are grouped by scene already but the scenes themselves are not in order (i.e. episodes shuffled by scene): in this case sorting is unnecessary and will take `O(nlgn)`.

habitat/core/dataset.py
erikwijmans: Maybe make it an `if` and `else`?  Then it'll be clear that `self._iterator = iter(self.episodes)` and `self._shuffle_iterator_scene()` are mutually exclusive

habitat/core/dataset.py
erikwijmans: Better plan.  Use `pydash.group_by` as that doesn't have the limitation that things need to be sorted! https://pydash.readthedocs.io/en/latest/api.html#pydash.collections.group_by

Almost certainly some dependency already relies on pydash

erikwijmans: pydash also provides some cool syntactic sugar, so `num_scene_groups` would be:
```
num_scene_groups = len(pydash.group_by(episodes, 'scene_id'))
```

habitat/core/dataset.py
JasonJiazhiZhang: hmm.. `self._iterator = iter(self.episodes)` and `self._shuffle_iterator_scene()` aren't actually mutually exclusive atm tho. (It can be if I use `pydash.groupby` that you suggested)

habitat/core/dataset.py
JasonJiazhiZhang: I checked with `pipdeptree`, `pydash` is not currently replied on. I'll be conservative and not using `pydash` for now, we can definitely discuss this more though.

habitat/core/dataset.py
erikwijmans: Or use `groupby(self.episodes, key=lambda x: x.scene_id)` in `self._shuffle_iterator_scene()`?

JasonJiazhiZhang: I also called `self._shuffle_iterator_scene()` [here](https://github.com/JasonJiazhiZhang/habitat-api/blob/9a56168de80418cdd4af4fdf322b2b77ef3983e1/habitat/core/dataset.py#L344) for scene switching when the episode-count for same-scene episode is reached. In this case it's in the middle of an iteration and `self._shuffle_iterator_scene()` shuffles the whatever are still on the iterator, which would be part of `episodes`, possibly not in the original order. I made `self._shuffle_iterator_scene()` dealing with iterator instead of list mainly for that case 😃

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+            num_episode_sample: number of episodes to be sampled.
+                -1 for no sampling.
         """
-        if num_episodes == -1:
-            return
-        if num_episodes < -1:
-            raise ValueError(
-                f"Invalid number for episodes to sample: {num_episodes}"
+        # sample episodes
+        if num_episode_sample != -1:
```
abhiskk: Can we use `None` for not sampling?

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+
+    def __next__(self):
+        r"""
+        The main logic for handling how episodes will be iterated.
+        Returns:
+            next episode.
+        """
+        try:
+            next_episode = next(self._iterator)
+        except StopIteration:
```
abhiskk: you can use something like `except StopIteration as e` and then use `raise e` later.

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+            and self.rep_count >= self.max_scene_repetition - 1
+        ):
+            self._shuffle_iterator_scene()
+            self.rep_count = 0
+
+        self.prev_scene_id = next_episode.scene_id
+        return next_episode
+
+    def _shuffle_iterator_scene(self) -> None:
+        r"""
```
abhiskk: Can you follow the same convention for docstrings as elsewhere, I think the comment starts from the first line.

habitat/core/dataset.py
```diff
@@ -90,15 +96,21 @@ def get_episodes(self, indexes: List[int]) -> List[T]:
         """
         return [self.episodes[episode_id] for episode_id in indexes]
 
-    def get_episode_iterator(self):
+    def get_episode_iterator(self, *args: Any, **kwargs: Any) -> Iterator:
         r"""
-        Creates and returns an iterator that iterates through self.episodes
-        in the desirable way specified.
+        Gets episode iterator with options. Options are specified in
```
abhiskk: Update the formatting of docstring.

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
             self.episodes = new_episodes
         return new_datasets
 
-    def sample_episodes(self, num_episodes: int) -> None:
-        """
-        Sample from existing episodes a list of episodes of size num_episodes,
-        and replace self.episodes with the list of sampled episodes.
+
+class EpisodeIterator(Iterator):
+    r"""
```
abhiskk: Formatting of docstring.

habitat/core/dataset.py
```diff
@@ -236,20 +246,116 @@ def get_splits(
...
+            next_episode = next(self._iterator)
+        except StopIteration:
+            if not self.cycle:
+                raise StopIteration
+            self._iterator = iter(self.episodes)
+            if self.shuffle_scene_when_cycle:
+                self._shuffle_iterator_scene()
+            next_episode = next(self._iterator)
+
+        if self.prev_scene_id == next_episode.scene_id:
```
abhiskk: We can make this variable private.

habitat/core/dataset.py
JasonJiazhiZhang: @abhiskk  The params are read off of yaml config, which doesn't like `None` mixed with `int`s

habitat/core/dataset.py
```diff
@@ -236,20 +243,115 @@ def get_splits(
...
+        num_episode_sample: int = -1,
+    ):
+        r"""
         Args:
-            num_episodes: number of episodes to sample, input -1 to use
-            whole episodes
+            episodes: list of episodes.
+            cycle: if true, cycle back to first episodes when StopIteration.
+            shuffle_scene_when_cycle: if true, shuffle scene groups when cycle.
+            group_by_scene: if true, group episodes from same scene.
```
apsdehal: Discussed offline with Jiazhi about `group_by_scene` attribute and that it is not being properly used at this point. In current state, group_by_scene is kind of redundant as when cycle is true, episodes are always grouped by default.

habitat/core/dataset.py
```diff
@@ -236,20 +243,115 @@ def get_splits(
...
+        if group_by_scene:
+            num_scene_groups = len(
+                list(groupby(episodes, key=lambda x: x.scene_id))
+            )
+            num_unique_scenes = len(set([e.scene_id for e in episodes]))
+            if num_scene_groups >= num_unique_scenes:
+                self.episodes = sorted(self.episodes, key=lambda x: x.scene_id)
+        self.max_scene_repetition = max_scene_repeat
+        self.shuffle_scene_when_cycle = shuffle_scene_when_cycle
+        self.rep_count = 0
```
apsdehal: All 3 can probably be private variables.

habitat/core/dataset.py
JasonJiazhiZhang: @apsdehal added a check in `_shuffle_iterator_scene` to make this option useful.

habitat/core/dataset.py
```diff
@@ -236,20 +243,116 @@ def get_splits(
...
+        self.group_by_scene = group_by_scene
+        if group_by_scene:
+            num_scene_groups = len(
+                list(groupby(episodes, key=lambda x: x.scene_id))
+            )
+            num_unique_scenes = len(set([e.scene_id for e in episodes]))
+            if num_scene_groups >= num_unique_scenes:
+                self.episodes = sorted(self.episodes, key=lambda x: x.scene_id)
+        self.max_scene_repetition = max_scene_repeat
+        self.shuffle_scene_when_cycle = shuffle_scene_when_cycle
```
abhiskk: condense the name of this argument.

habitat/core/dataset.py
```diff
@@ -236,20 +243,116 @@ def get_splits(
...
+        return self
+
+    def __next__(self):
+        r"""The main logic for handling how episodes will be iterated.
+
+        Returns:
+            next episode.
+        """
+        try:
+            next_episode = next(self._iterator)
```
abhiskk: Just double checking, is there any other better way to handle this besides catching an exception, if not this seems good.

habitat/core/dataset.py
```diff
@@ -236,20 +243,116 @@ def get_splits(
...
+        except StopIteration as e:
+            if not self.cycle:
+                raise e
+            self._iterator = iter(self.episodes)
+            if self.shuffle_scene_when_cycle:
```
abhiskk: condense the variable name, 4 words in variable look ugly :/

habitat/core/dataset.py
JasonJiazhiZhang: As far as I know, there's no simple way to check if the internal iterator has exhausted all episodes.

JasonJiazhiZhang:
>a question regarding backwards compatibility, in the default setting is the behavior of new data iterator same as the previous data iterator?

@abhiskk, yes the default iterator behavior is the same as before.

habitat/core/dataset.py
erikwijmans: You can do
```python
next_episode = next(self._iterator, None)
if next_episode is None:
    ....
```

PR #166, opened 2019-07-18
Make fog-of-war darker and make that value changeable

erikwijmans:
## Motivation and Context

The fog-of-war should be darker by default and the level of darkness should be configurable

## How Has This Been Tested

Via a less fancy gif:
trajectory

## Types of changes

### DIFF
--- a/habitat/utils/visualizations/maps.py
+++ b/habitat/utils/visualizations/maps.py
@@ -327,11 +327,10 @@ def get_topdown_map(
     return top_down_map
 
 
-FOG_OF_WAR_COLOR_DESAT = np.array([[0.7], [1.0]])
-
-
 def colorize_topdown_map(
-    top_down_map: np.ndarray, fog_of_war_mask: Optional[np.ndarray] = None
+    top_down_map: np.ndarray,
+    fog_of_war_mask: Optional[np.ndarray] = None,
+    fog_of_war_desat_amount: float = 0.5,
 ) -> np.ndarray:
     r"""Convert the top down map to RGB based on the indicator values.
         Args:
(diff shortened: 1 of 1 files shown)


mathfac:
Will leave for @dhruvbatra to review. One thing I noticed, that it is logical to draw fog before drawing shortest path and start and end points.

dhruvbatra:
Contrast looks good. But shouldn't the darkness affect the non-navigable points as well (they seem to be just as bright/white as before)?  

More generally, I agree with @JasonJiazhiZhang's comment from #160 -- we are inconsistent in the semantics of brightness/whiteness. Currently: 
-- white = not navigable (inside or outside fog)
-- dark gray = navigable + fog/unexplored
-- light gray = navigable + no-fog/explored

If we want fog to be dark and no-fog to be light (which is sensible and default in games), be consistent that more brightness = more certainty about navigability, then it should be
-- black = not navigable (inside or outside fog)
-- dark gray = navigable + fog/unexplored
-- light gray/white = navigable + no-fog/explored.

JasonJiazhiZhang:
@dhruvbatra I can see this scheme becoming a bit problematic for people that don't care much of the Fog of War and only mostly care about what's navigable. **black <-> dark gray** may be not enough contrast in that situation. Ideally we want all three terrain situations to be highly distinguishable from each other.

JasonJiazhiZhang:
One other comment (that comes to mind becomes of the large room example in this PR): Maybe we should make the circle radius configurable? In the example, the agent can see more (e.g. left/right wall) than suggested by the FoW.

dhruvbatra:
> @dhruvbatra  I can see this scheme becoming a bit problematic for people that don't  care much of the Fog of War and only mostly care about what's navigable.  black <-> dark gray may be not enough contrast  in that situation. Ideally we want all three terrain situations to be  highly distinguishable from each other.

I agree. So we can leave this as is.

erikwijmans:
> One other comment (that comes to mind becomes of the large room example in this PR): Maybe we should make the circle radius configurable? In the example, the agent can see more (e.g. left/right wall) than suggested by the FoW.

The distance the agent can see and the FOV the agent's view cone are configurable: https://github.com/facebookresearch/habitat-api/blob/master/habitat/config/default.py#L77

JasonJiazhiZhang:
> The distance the agent can see and the FOV the agent's view cone are configurable: https://github.com/facebookresearch/habitat-api/blob/master/habitat/config/default.py#L77

Sorry I missed that 😅 . Was looking for a value close to `100` in the config.

erikwijmans:
Yeah, the value in the config is in meters, but the draw functions take things thing pixels :)

dhruvbatra:
LGTM.

PR #172, opened 2019-07-24
Update config system, resolve TODOs from refactoring

JasonJiazhiZhang:
## Motivation and Context
This PR is a follow-up on #153, including several refactoring efforts that did not make into that PR, which includes:
  * Clean up the config system. `TRAINER` layer is deleted, all its children are moved one layer up in the config tree. Now general experiment-related configs are moved to the outmost config node. PPO related config will be under `RL.PPO`. This is the most significant change of this PR, and should improve structural clarity of config system. 
  * Config merging logic now checks to ensure old checkpoints won't break on evaluation. 
  * Fix Typing for environment,  mainly for `habitat/core/vector_env.py` and `habitat_baselines/common/env_utils.py`. (note: `RLEnv` is NOT a subclass of `Env`, using `Union` in some cases is required, e.g. in VectorEnv `__init__`)
  * Properly use  baseline environment registry for vector env construction  (environment registry was added in #153 but was not used in `env_utils.py`).
  * Refactor `eval()` in `PPOTrainer` to "BaseTrainer" since it's general enough.
  * Fix signture inconsistency of `__init__` between `NavRLEnv` and `RLEnv`. Now `NavRLEnv.__init__` will take in the same args as base class.  This consistency is necessary for the new environment registration system. 
  * Remove DummyWriter in tensorboard_utils.py (deliver this [TODO](https://github.com/facebookresearch/habitat-api/blob/31318f81db05100099cfd308438d5930c3fb6cd2/habitat_baselines/common/tensorboard_utils.py#L13))
  * Make use `torch.LambdaLR` as LR scheduler (related to this [TODO](https://github.com/facebookresearch/habitat-api/blob/31318f81db05100099cfd308438d5930c3fb6cd2/habitat_baselines/common/utils.py#L80))
  * Fix a bug mentioned [here](https://github.com/facebookresearch/habitat-api/issues/171#issuecomment-514410294).

## Types of changes

## Checklist


habitat_baselines/config/default.py
```diff
@@ -14,92 +14,83 @@
 DEFAULT_CONFIG_DIR = "configs/"
 CONFIG_FILE_SEPARATOR = ","
 # -----------------------------------------------------------------------------
-# Config definition
+# EXPERIMENT CONFIG
 # -----------------------------------------------------------------------------
 _C = CN()
 _C.BASE_TASK_CONFIG_PATH = "configs/tasks/pointnav.yaml"
 _C.TASK_CONFIG = CN()  # task_config will be stored as a config node
 _C.CMD_TRAILING_OPTS = ""  # store command line options"
```
mathfac: Why do we store command lines in config, not just merge them in config when passed? That can create confusion.

habitat_baselines/config/default.py
```diff
@@ -14,92 +14,83 @@
...
+_C.TRAINER_NAME = "ppo"
+_C.ENV_NAME = "NavRLEnv"
+_C.SIMULATOR_GPU_ID = 0
```
mathfac: We have setups when we use multiple  GPU ids for simulators. Maybe we want to be more conservative what to put in root node. `_C.ENV_NAME` is user for RL trainers mostly.

habitat/core/env.py
```diff
@@ -366,7 +366,7 @@ def step(self, action: int) -> Tuple[Observations, Any, bool, dict]:
 
         return observations, reward, done, info
 
-    def seed(self, seed: int) -> None:
+    def seed(self, seed: Optional[int] = None) -> None:
```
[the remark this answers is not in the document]
JasonJiazhiZhang: I did this to keep signature consistency with [gym.Env](https://github.com/openai/gym/blob/9dea81b48a2e1d8f7e7a81211c0f09f627ee61a9/gym/core.py#L115), since RLEnv is a subclass of gym.Env.

habitat_baselines/common/env_utils.py
```diff
@@ -5,42 +5,42 @@
...
         env object created according to specification.
     """
     dataset = make_dataset(
-        task_config.DATASET.TYPE, config=task_config.DATASET
-    )
-    env = env_class(
-        config_env=task_config, config_baseline=rl_env_config, dataset=dataset
+        config.TASK_CONFIG.DATASET.TYPE, config=config.TASK_CONFIG.DATASET
     )
+    env = env_class(config=config, dataset=dataset)
```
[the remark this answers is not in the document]
JasonJiazhiZhang: This is a design decision I made. The reasoning behind it is to ensure successful init from registry.  To do so, we want all the env_class (e.g. `NavRLEnv`) to have the same `__init__` signature as base class `RLEnv`. However, each sub env class will inevitably have its own additional parameter, somewhere under root config that we don't know the exact location when calling `env_class()`. The other thing is that we want the core config node to be intact, and free of extra params, for interval env (`self._env`) construction. I think here it makes sense to pass in a root_config, which contains a clean task_config at `config.TASK_CONFIG`,  and let each env subclasss retrieve its specific config node.

habitat_baselines/config/default.py
JasonJiazhiZhang: This is to support the ideal overwrite behavior when loading saved config upon evaluation run. See [here](https://github.com/JasonJiazhiZhang/habitat-api/blob/743e82e764e2a8b29dee1c93ef54815b7d192b03/habitat_baselines/rl/ppo/ppo_trainer.py#L344-L354).

habitat_baselines/common/base_trainer.py
```diff
@@ -4,10 +4,19 @@
...
 
+import torch
+
+from habitat import Config, logger
+from habitat_baselines.common.tensorboard_utils import TensorboardWriter
+from habitat_baselines.common.utils import poll_checkpoint_folder
+
 
 class BaseTrainer:
+
```
abhiskk: nit: this new line is not needed.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -252,25 +244,21 @@ def train(self) -> None:
         count_steps = 0
         count_checkpoints = 0
 
-        with (
-            get_tensorboard_writer(
-                log_dir=ppo_cfg.tensorboard_dir,
-                purge_step=count_steps,
-                flush_secs=30,
-            )
+        lr_scheduler = LambdaLR(
```
abhiskk: This looks great 👍

habitat_baselines/common/base_trainer.py
```diff
@@ -32,15 +41,77 @@ class BaseRLTrainer(BaseTrainer):
...
+            assert (
+                len(self.config.TENSORBOARD_DIR) > 0
+            ), "Must specify a tensorboard directory for video display"
+        if "disk" in self.config.VIDEO_OPTION:
+            assert (
+                len(self.config.VIDEO_DIR) > 0
+            ), "Must specify a directory for storing videos on disk"
+
+        with TensorboardWriter(
+            self.config.TENSORBOARD_DIR, purge_step=0, flush_secs=30
```
abhiskk: we should make the `purge_step` and `flush_secs` as class attributes instead of constants.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -350,69 +338,22 @@ def train(self) -> None:
...
-                        )
-                    )
-                    prev_ckpt_ind += 1
-                    self._eval_checkpoint(
-                        checkpoint_path=current_ckpt,
-                        writer=writer,
-                        cur_ckpt_idx=prev_ckpt_ind,
-                    )
-
     def _eval_checkpoint(
```
abhiskk: The modularity for eval is also looking better, one suggestion can we split the `_eval_checkpoint` into further modular steps, like in 1) construction of env, 2) evaluation inside envs.

PR #170, opened 2019-07-22
Update baseline test import and fix docstring

JasonJiazhiZhang:
## Motivation and Context
Currently we run test scripts on `habitat_baselines` modules [if `torch` is installed](https://github.com/facebookresearch/habitat-api/blob/7eb3da128c5cdd1f8a2fe0281f503009c6d5402d/test/test_baseline_agents.py#L14-L19). This is becoming problematic since more dependencies are coming to habitat_baselines (e.g. tensorboard). This PR fixes that and allows for future expansion of test scripts on `habtiat_baselines`. This PR also fixes some docstring inconsistency introduced by #153. 

## How Has This Been Tested

Tests skip properly if not all dependencies are installed for habitat_baselines. 
Tests pass if habitat_baselines is properly installed.  
## Types of changes

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #173, opened 2019-07-25
Disable tensorboard to avoid VectorEnv error on closing

JasonJiazhiZhang:
## Motivation and Context
Currently enabling tensorboard will cause an issue upon `VectorEnv` deletion. When training/eval is completed and `VectorEnv`'s `__del__` function gets called, the connection between worker and master is already lost before the close command gets send into working env. This seems to be rooted from tensorboard's underlying async writer using and closing BytesIO. See the `Run api tests` section of [circle-ci test result](https://circleci.com/gh/facebookresearch/habitat-api/576?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for this stealthy error. 

Working on a proper fix, but for now this will patch the CI test. 

## How Has This Been Tested
The [tiny run script](https://github.com/facebookresearch/habitat-api/blob/d3d6a5eb2fa8d1a0878e647e4f965834898af6cc/.circleci/config.yml#L199) in CI will finish properly. 

## Types of changes

## Checklist

### DIFF
--- a/habitat_baselines/config/pointnav/ppo_train_test.yaml
+++ b/habitat_baselines/config/pointnav/ppo_train_test.yaml
@@ -10,3 +10,4 @@ TRAINER:
       task_config: "configs/tasks/pointnav.yaml"
       num_updates: 10
       num_mini_batch: 1
+      tensorboard_dir: ""


mathfac:
@JasonJiazhiZhang, as we don't have dependencies on command line arguments we can move train testing from CI command line to our test suite and avoid changes to circleci configs.

JasonJiazhiZhang:
@mathfac totally agree. #170 was aiming to enable that. As habitat_baselines gets more modular we should definitely introduce more tests. Though this PR doesn't really change the script, just the config yaml file.

Just to loop you in on the bug this PR addresses: it seems that upon program completion/ garbage collection, internal IO writer for tensorboard will interfere with connection writing/reading of VectorEnv.

mathfac:
@JasonJiazhiZhang, yes that suggestion is out of scope of this PR. 
For example, with redesign of baselines and moving whole configuration to config files that became easy to create parameterized test that loop through registered trainers and run short training on test configs. That will guarantee that new baselines should have testing by default.

PR #178, opened 2019-08-02
Make ppo_trainer eval use generalized metric

JasonJiazhiZhang:
## Motivation and Context
Making `eval()` in ppo_trainer retrieve the name (`sensor_uuid`) of performance metric used instead of always using `spl`. This change is necessary for ppo_trainer to support new tasks that might have performance metric other than `spl`. 

## How Has This Been Tested
Tests pass.

## Types of changes

## Checklist

### DIFF
--- a/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -379,6 +379,15 @@ class PPOTrainer(BaseRLTrainer):
         self.agent.load_state_dict(ckpt_dict["state_dict"])
         self.actor_critic = self.agent.actor_critic
 
+        # get name of performance metric, e.g. "spl"
+        metric_name = self.config.TASK_CONFIG.TASK.MEASUREMENTS[0]
+        metric_cfg = getattr(self.config.TASK_CONFIG.TASK, metric_name)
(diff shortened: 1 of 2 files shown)


habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -379,6 +379,15 @@ def _eval_checkpoint(
...
         self.actor_critic = self.agent.actor_critic
 
+        # get name of performance metric, e.g. "spl"
+        metric_name = self.config.TASK_CONFIG.TASK.MEASUREMENTS[0]
+        metric_cfg = getattr(self.config.TASK_CONFIG.TASK, metric_name)
+        measure_type = baseline_registry.get_measure(metric_cfg.TYPE)
+        assert measure_type is not None, "invalid measurement type {}".format(
+            metric_cfg.TYPE
+        )
+        self.metric_uuid = measure_type(None, None)._get_uuid()
```
mathfac: @JasonJiazhiZhang what `(None, None)` stands for here?

JasonJiazhiZhang: Since we just want the uuid, I created a dummy measure object with `(None, None)` and called its `_get_uuid()`.  Although this isn't the most elegant code, I discussed with abhishek and we decided to do it this way :)

PR #187, opened 2019-08-14
Manual control python demo

angelxuanchang:
## Motivation and Context
This PR introduces a demo that allows a user to run through a PointGoal navigation episode in the same way as trained agents, using keyboard controls.  The episode is then compared against oracle shortest path and a trained blind RL agent.

## How Has This Been Tested

Install Habitat-API as usual. Run `python examples/demo.py --task-config configs/tasks/pointnav.yaml --overlay` to get the following interactive view:
image
Use the indicated keys to play and the `Q` key to exit.

Follow the instructions at the top of the `demo.py` file to run more advanced examples (e.g. `python examples/demo.py --task-config configs/tasks/pointnav.yaml,data/replica_demo/replica_test.yaml --agent blind --overlay --scenes-dir .` which will run an episode in a Replica scene)
The following interactive view will be opened
image

## Types of changes

The only change in existing Habitat-API code was to allow the `Env.reset()` function to be called but keep the same episode instead of iterating forward (through a new `keep_current_episode` argument which defaults to false).

## Checklist

### DIFF
--- a/examples/demo.py
+++ b/examples/demo.py
@@ -13,6 +13,7 @@ import numpy as np
 from habitat.tasks.nav.nav_task import NavigationEpisode, NavigationGoal
 from habitat.tasks.nav.shortest_path_follower import ShortestPathFollower
 from habitat.utils.visualizations import maps
+from examples.agent_demo.demo_blind_agent import DemoBlindAgent
 
 from time import sleep
 from typing import List, NamedTuple, Tuple
@@ -135,7 +136,7 @@ def draw_top_down_map(info, heading, output_size):
 
 class Viewer:
     def __init__(self, initial_observations, 
-            overlay_goal_radar=None,            
+            overlay_goal_radar=None,
             goal_display_size=128,
             show_map=False):
         self.overlay_goal_radar = overlay_goal_radar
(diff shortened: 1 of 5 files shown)


examples/demo.py
```diff
@@ -0,0 +1,572 @@
...
+from typing import List, NamedTuple, Tuple
+
+class ActionKeyMapping(NamedTuple):
+    name: str
+    key: int
+    action_id: int
+    is_quit: bool = False
+
+AGENT_ACTION_KEYS = [
+    ActionKeyMapping("FORWARD", ord("w"), 0),
```
mathfac: Let's use `habitat.SimulatorActions` instead of `0, 1, 2`. That is proper way of mapping actions.

habitat/core/env.py
```diff
@@ -165,7 +165,7 @@ def _reset_stats(self) -> None:
         self._elapsed_steps = 0
         self._episode_over = False
 
-    def reset(self) -> Observations:
+    def reset(self, keep_current_episode=False) -> Observations:
```
mathfac: I would advice not to add `keep_current_episode` parameter to `Env.reset` as we don't have enough use-cases and there is another way to get around here using [iterator options in config](https://github.com/facebookresearch/habitat-api/blob/master/habitat/config/default.py#L25):
```
ENVIRONMENT.ITERATOR_OPTIONS.CYCLE = True
ENVIRONMENT.ITERATOR_OPTIONS.NUM_EPISODE_SAMPLE = 1
```.
That will repeat same episode over and over again.

File: examples/demo.py [hunk 3]
```diff
@@ -0,0 +1,572 @@
...
+
+    @property
+    def center(self):
+        return (self.left + int(self.width/2), self.top + int(self.height/2))    
+
+def transform_rgb_bgr(image):
+    return image[:, :, [2, 1, 0]]
+
+
+def write_textlines(output, textlines, size=1, offset=(0,0), fontcolor=(255,255,255)):
```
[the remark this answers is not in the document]
angelxuanchang: Good suggestions.  Will leave to another refactor pass to determine what is generally useful in other settings.

PR #196, opened 2019-08-30
Introduce task actions to support continuous actions and not pure navigation tasks.

mathfac:
## Motivation and Context
Redesign Habitat-API to support continues actions and not pure navigation tasks.
- **New class for Task Actions**:
  - Added registry for actions classes. We use a class not a method for action definition to provide relevant action space that uses Task/Dataset instance's information like number of words for answer vocabulary, resolution of frame etc.
  - Change `step` function interface to support action parameters 
  - Added step function to the task with actions defined as task's methods
  - Provide action space for each type of action
  - Add reset function to Actions 
  - Added teleport action as example of parameterized action.
  - Provide functionality to register several actions of same type using config only
  - Create ActionSpace GYM space to describe action space in elegant way for multiple actions. Propagate action space through env.

- **Moved `Simulator` calls to `EmbodiedTask`**:
  - Move `is_episode_over` functionality to Task class from Simulator, handle capturing finished conditions for different tasks differently
  - Remove simulator `STOP` action
  - Make Task methods reset and step provide observations that include sim observations
  - Add possible actions to config and task
  - Move from SimulatorAction list to defined Tasks in actions, handle non-stop actions
  - Check usage of `*args, **kwargs` for all sensors, action, measures.
  - Use dict as return type for `act` methods.

- **Made `VecEnv` independent from `Env` step signature.**

## How Has This Been Tested
Added additional test for habitat Tasks and fixed old ones.

## Types of changes

## Checklist


habitat/core/env.py
```diff
@@ -215,12 +215,19 @@ def _update_step_stats(self) -> None:
...
+    ) -> Observations:
         r"""Perform an action in the environment and return observations.
 
         Args:
             action: action (belonging to ``action_space``) to be performed 
-                inside the environment.
+                inside the environment. Action is a name or index of allowed
+                task's action.
+            action_args: action arguments (belonging to action's
+            ``action_space``) to support parametrized and continues actions.
```
msavva: typo: `continues` -> `continuous`

habitat/core/registry.py
```diff
@@ -147,6 +148,20 @@ def register_measure(cls, to_register=None, *, name: Optional[str] = None):
             "measure", to_register, name, assert_type=Measure
         )
 
+    @classmethod
+    def register_task_action(
+        cls, to_register=None, *, name: Optional[str] = None
+    ):
+        r"""Register a measure to registry with key 'name'
```
msavva: Perhaps a clearer to say `Add a measure in this registry under key 'name'`

habitat/core/registry.py
```diff
@@ -147,6 +148,20 @@ def register_measure(cls, to_register=None, *, name: Optional[str] = None):
...
+
+        Args:
+            name: Key with which the measure will be registered.
+                If None will use the name of the class
```
msavva: clarify by saying `will use the name of the measure's class`

habitat/core/embodied_task.py
```diff
@@ -131,6 +135,43 @@ def __init__(
         self._sim = sim
         self._dataset = dataset
 
+        self._possible_actions = OrderedDict()
+        for action in config.POSSIBLE_ACTIONS:
+            assert action in registry.mapping["task_action"]
```
erikwijmans: use `registry. get_task_action` then assert that the result is not `None`

habitat/core/embodied_task.py
```diff
@@ -131,6 +135,43 @@ def __init__(
...
+            self._possible_actions[action] = registry.mapping["task_action"][
+                action
+            ]
```
erikwijmans: Maybe worth adding something like
```python
for i, v in enumerate(self._possible_actions.value()):
    self._possible_actions[i] = v
```

for cleaner lookup by index later?

habitat/core/embodied_task.py
```diff
@@ -131,6 +135,43 @@ def __init__(
...
+        action_args: Dict[str, Any] = None,
+    ):
+        if action_args is None:
+            action_args = {}
+        if isinstance(action, (int, np.integer)):
+            if action >= len(self._possible_actions):
+                raise ValueError(f"Action index '{action}' is out of range.")
+            action = list(self._possible_actions.keys())[action]
+        assert (
+            action in self._possible_actions.keys()
```
erikwijmans:
```suggestion
            action in self._possible_actions
```

habitat/tasks/nav/nav_task.py
```diff
@@ -719,6 +720,30 @@ def __init__(
...
+    def look_up(self):
+        return self._sim.step(SimulatorActions.LOOK_UP)
+
+    @registry.register_task_action(name="look_down")
+    def look_down(self):
+        return self._sim.step(SimulatorActions.LOOK_DOWN)
+
+    @registry.register_task_action(name="stop")
+    def stop(self):
+        return self._sim.step(SimulatorActions.STOP)
```
erikwijmans: We should probably rename `SimulatorActions` to `Actions` or `BaseActions` as `STOP` isn't a simulator action anymore :)

mathfac: @erikwijmans, but `SimulatorActions`is used only as bridge between Simulator Actions and Task Actions. There will be no task action enum anymore. Registry will be used for that. I can remove `STOP` action for HSIM here.

mathfac:
> The `habitat.SimulatorActions` can't be numeric anymore. The number an action is on is entirely dependent on how many actions the task loads.

`habitat.SimulatorActions` is separate notion and is used only in Task action method and will depend on ActionConfiguration activated. My personal point of view that HabitatSimulator actions can be just method on HabitatSim class will be wrapped with correspondent TaskActions.

mathfac:
@msavva and @erikwijmans thank you for the review, addressed the comments. `SimulatorActions` left unchanged for now. Added teleport action as example of parameterized action and added action spec registration for the task action registry.

habitat/core/embodied_task.py
mathfac: @erikwijmans, make sense, but then it tricky to check number of possible actions. Maybe worth to have separate dict with indexes.

examples/benchmark.py
```diff
@@ -16,7 +16,7 @@ def reset(self):
 
     def act(self, observations):
         action = SimulatorActions.MOVE_FORWARD
-        return action
+        return action, None
```
apsdehal: Suggestion: Can we move to dict based input and output system now? Adding more arguments and return items to the signatures is cumbersome and requires doing a lot of changes to make sure there is no regression. Moving to dict based output and input system will remove these concerns. I foresee in future that if we need to add more arguments and return items, we will have to do this again. 

So, I suggest moving to dict based output, input structure. I moved to it in Pythia and life has been a breeze since then.

habitat/core/embodied_task.py
```diff
@@ -131,6 +143,50 @@ def __init__(
...
+        self._possible_actions = OrderedDict()
+        for action in config.POSSIBLE_ACTIONS:
+            action_method = registry.get_task_action(action)
+            self._possible_actions[action] = action_method
+
+    def step(
+        self,
+        action: Union[str, int],
+        episode: Type[Episode],
+        action_args: Dict[str, Any] = None,
```
apsdehal: Can we add `*args, **kwargs` here and everywhere, to easily add new arguments later in future and easy backward compatibility?

mathfac:
Working on this change I notice that the scope of changes increases as long as other parts of API change. That would be great to break changes in smaller parts, but they are pretty coupled. I updated list of changes required. Please, take a look if it looks reasonable. Most of them are implemented and as soon as tests will be passing I will update the PR.

mathfac:
CI tests are passing, PR is ready for the review @abhiskk, @erikwijmans, @apsdehal.

configs/test/habitat_mp3d_eqa_test.yaml
```diff
@@ -1,5 +1,8 @@
 TASK:
   TYPE: EQA-v0
+  SENSORS: ['QUESTION_SENSOR']
+  POSSIBLE_ACTIONS: ['STOP', 'MOVE_FORWARD', 'TURN_LEFT', 'TURN_RIGHT', 'ANSWER']
```
erikwijmans: Shouldn't `ANSWER` and `STOP` be the same action? You only get the try to answer the question once.

examples/new_actions.py
```diff
@@ -126,19 +127,46 @@ def get(self):
...
+        return self._sim.step(habitat.SimulatorActions.STRAFE_RIGHT)
+
+
 def main():
     habitat.SimulatorActions.extend_action_space("STRAFE_LEFT")
     habitat.SimulatorActions.extend_action_space("STRAFE_RIGHT")
 
     config = habitat.get_config(config_paths="configs/tasks/pointnav.yaml")
     config.defrost()
+    # config.TASK.TYPE = "NavTaskWithStrafe-v0"
```
erikwijmans:
```suggestion
```

habitat/core/embodied_task.py
```diff
@@ -140,6 +230,117 @@ def __init__(
...
+                sensor_cfg.TYPE
+            )
+            print("sensor type {}".format(sensor_cfg.TYPE))
+            task_sensors.append(
+                sensor_type(sim=sim, config=sensor_cfg, dataset=dataset)
+            )
+
+        self.sensor_suite = SensorSuite(task_sensors)
+
+        # TODO: move sensors and measures to `_init_entities`
```
erikwijmans: Does this TODO still need to happen?

habitat/core/embodied_task.py
```diff
@@ -140,6 +230,117 @@ def __init__(
...
+    def step(self, action: Union[int, Dict[str, Any]], episode: Type[Episode]):
+        if "action_args" not in action or action["action_args"] is None:
+            action["action_args"] = {}
+        action_name = action["action"]
+        if isinstance(action_name, (int, np.integer)):
+            if action_name >= len(self.actions):
+                raise ValueError(
+                    f"Action index '{action_name}' is out of range."
+                )
+            action_name = list(self.actions.keys())[action_name]
```
erikwijmans: Probably doesn't matter, but can we have a `self._action_keys = list(self.actions.keys())` in the `__init__` method and use that here?  Making this on every step seems excessive.

habitat/core/embodied_task.py
```diff
@@ -140,6 +230,117 @@ def __init__(
...
+        self._is_episode_active = self._check_episode_is_active(
+            observations=observations, action=action, episode=episode
+        )
+
+        return observations
+
+    def get_action_name(self, action_index: int):
+        if action_index >= len(self.actions):
+            raise ValueError(f"Action index '{action}' is out of range.")
+        return list(self.actions.keys())[action_index]
```
erikwijmans: And here `self._action_keys`!

habitat/core/embodied_task.py
erikwijmans: Actually, shouldn't this be `action_name = self.get_action_name(action_name)`

habitat/core/env.py
```diff
@@ -107,7 +107,7 @@ def __init__(
                 **self._task.sensor_suite.observation_spaces.spaces,
             }
         )
-        self.action_space = self._sim.action_space
+        self.action_space = self._task.action_space()
```
erikwijmans: Why the change to a method instead of a property?

habitat/datasets/eqa/mp3d_eqa_dataset.py
```diff
@@ -52,6 +58,19 @@ def __init__(self, config: Config = None) -> None:
         with gzip.open(config.DATA_PATH.format(split=config.SPLIT), "rt") as f:
             self.from_json(f.read())
 
+        questions_vocabulary = {
```
erikwijmans: This PR is big enough, but the question stuff here for EQA needs to be fixed.  The questions are a list of tokens, not an individual token themself.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -239,23 +234,16 @@ def reset(self):
...
-        self._is_episode_active = True
         return self._sensor_suite.get_observations(sim_obs)
 
     def step(self, action):
-        assert self._is_episode_active, (
-            "episode is not active, environment not RESET or "
-            "STOP action called previously"
-        )
 
         if action == self.index_stop_action:
```
erikwijmans: Do we need references to stop?

habitat/tasks/eqa/eqa_task.py
```diff
@@ -48,77 +45,249 @@ class EQAEpisode(NavigationEpisode):
...
-    def _get_observation(
+    def get_observation(
         self,
         observations: Dict[str, Observations],
         episode: EQAEpisode,
-        **kwargs
+        **kwargs,
     ):
-        return episode.question.question_text
+        return self._dataset.get_questions_vocabulary()[
```
erikwijmans: Same comment about fixing the question stuff

habitat/tasks/nav/nav_task.py
```diff
@@ -691,35 +720,166 @@ def update_fog_of_war_mask(self, agent_position):
...
+
+    def _get_uuid(self, *args: Any, **kwargs: Any) -> str:
+        return self.name
+
+    def step(self, *args: Any, **kwargs: Any):
+        r"""Update ``_metric``, this method is called from ``Env`` on each
+        ``step``.
+        """
+        self._task.is_stop_called = True
+        return self._sim.get_observations_at()
```
erikwijmans: Yep, looks like the stuff about the stop action can be removed from `simulator.step`

habitat/tasks/nav/nav_task.py
```diff
@@ -691,35 +720,166 @@ def update_fog_of_war_mask(self, agent_position):
...
+        rotation: List[float],
+        **kwargs: Any
+    ):
+        r"""Update ``_metric``, this method is called from ``Env`` on each
+        ``step``.
+        """
+
+        if not isinstance(rotation, list):
+            rotation = list(rotation)
+        # if not isinstance(position, list):
```
erikwijmans: Delete or uncomment

habitat/core/vector_env.py
```diff
@@ -291,14 +302,15 @@ def step_at(self, index_env: int, action: int):
         self._is_waiting = False
         return results
 
-    def async_step(self, actions: List[int]) -> None:
+    def async_step(self, data: List[Dict[str, Any]]) -> None:
```
erikwijmans: Let's push the backwards compatibility stuff into here and make `async_step` and `step` have the same signature

test/test_mp3d_eqa.py
```diff
@@ -15,9 +15,13 @@
 from habitat.core.embodied_task import Episode
 from habitat.core.logging import logger
 from habitat.datasets import make_dataset
+from habitat.tasks.eqa.eqa_task import AnswerAction
+from habitat.utils.test_utils import sample_non_stop_action
 
 CFG_TEST = "configs/test/habitat_mp3d_eqa_test.yaml"
-CLOSE_STEP_THRESHOLD = 0.028
+CLOSE_STEP_THRESHOLD = 0.03
+# CLOSE_STEP_THRESHOLD = 0.028
```
erikwijmans:
```suggestion
```

configs/test/habitat_mp3d_eqa_test.yaml
mathfac: I was following setup we had for EQA paper, but agree STOP is redundant.

habitat/core/embodied_task.py
mathfac: Already finished in next update.

habitat/datasets/eqa/mp3d_eqa_dataset.py
mathfac: Yes, EQA is moving to separate PR. It was here to test the change end-to-end.

configs/test/habitat_mp3d_eqa_test.yaml
erikwijmans: In the setup for the paper, we had STOP == ANSWER.  The agent only every tried to answer the question when stop was called and there were no additional sensor inputs provided after STOP was called (i.e. the last frame wasn't duplicated)

habitat/core/embodied_task.py
erikwijmans: I don't see this?  Or has that update not been pushed yet?

habitat/tasks/nav/nav_task.py
```diff
@@ -455,7 +461,8 @@ def update_metric(self, *args: Any, episode, action, **kwargs: Any):
         )
 
         if (
-            action == self._sim.index_stop_action
+            hasattr(self._task, "is_stop_called")
```
erikwijmans: EmbodiedTask should always have an `is_stop_called` field that gets set to `false` on reset and init.  Avoids this clunkiness.

habitat/core/embodied_task.py
```diff
@@ -309,7 +296,7 @@ def step(self, action: Union[int, Dict[str, Any]], episode: Type[Episode]):
                 raise ValueError(
```
erikwijmans: This check isn't needed, `get_action_name` does it.

habitat/tasks/nav/nav_task.py
mathfac: @erikwijmans what if your task doesn't have notion of stopping. When success or finished episode is just reach a target to some proximity.

mathfac:
- Removed EQA related changes. 
- Removed `STOP` action for EQA

habitat/tasks/nav/nav_task.py
erikwijmans: That's fine.  There is nothing that says it ever has to be set true, but the `hasattr(self._task, "is_stop_called")` check before everything is clunky.  One other suggestion:

Add
```
@property
def is_stop_called(self):
    return getattr(self, "_is_stop_called", False)
```

to the EmbodiedTask class.  The actions can then set `self._is_stop_called = True`.

erikwijmans: Btw, with the current method, isn't `_task .is_stop_called`, if it exists, always true?  I don't see where it gets set to false or deleted.

habitat/tasks/nav/nav_task.py
```diff
@@ -396,19 +421,27 @@ class SPL(Measure):
     https://arxiv.org/pdf/1807.06757.pdf
     """
 
-    def __init__(self, sim: Simulator, config: Config):
+    def __init__(
+        self,
+        *args: Any,
+        sim: Simulator,
+        task: EmbodiedTask,
```
erikwijmans: I think this will end up casing a circular reference i.e. the task holds a reference to the measure and the measure holds a reference to the task, and thus you can't every free the memory used by the task.  (I believe this is also a problem for the actions and the sensors)

There are a couple ways to deal with this:
1. The environment is the owner of actions/sensors/measures -- no circular references here.  Kinda a clunky design tho.
1. Use `weakref.proxy` when giving the measure -- gets rid of the circular here by giving the X a non-owning reference to the task.  This is kinda a weird solution tho.
1. Different API where there is a `task_state` that gets manipulate/populated by actions, then sensors, then measures.  It is passed in as an argument to their step/get/update functions.

The second one feels the most clean.  1 and 2 are both band-aids where the first makes the internals less clean and the second uses a kinda weird python API.

mathfac: That's valuable point. I expect generational garbage collector should take care of this as soon as there is no outside reference to the task. I will pass task in `reset` and `step` method as param to avoid this by design.

habitat/tasks/nav/nav_task.py
mathfac: The Environment class has step count limit after which it will stop the episode. `is_episode_active` is going to be redefined once it will be inheritance.

habitat/tasks/nav/nav_task.py
erikwijmans: Yeah, looking at this again, this does look like one of cases where the garbage collector can deal with this.

PR #198, opened 2019-09-06
Add py.typed file to indicate PEP 561 compatibility

Danmou:
See https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 graft habitat/utils/visualizations/assets
 include habitat/py.typed
+include habitat_baselines/py.typed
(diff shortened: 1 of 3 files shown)


Danmou:
> As I understand `setup.py` shouldn't be changed as we have packages defined as `packages=setuptools.find_packages(exclude=DEFAULT_EXCLUSION)`?

Actually it's because we have `include_package_data=True`, which makes setuptools include the files specified in `MANIFEST.in`.

>     1. Let's add comment `# Marker file for PEP 561.  The mypy package uses inline types.` to the `py.typed` file.
> 
>     2. Let's create same file for `habitat_baselines` package.

Done!

mathfac:
Thank you @Danmou, merged!

PR #199, opened 2019-09-08
PyRobot integration

abhiskk:
## Motivation and Context

Provides support for [PyRobot](https://github.com/facebookresearch/pyrobot) in Habitat. This is a step in the direction of trainining and testing models across simulation and reality. Along with `pyrobot` integration this PR also involves:

- Renaming of `SimulatorActions` to `HabitatSimActions`.
- Moving `HabitatSimActions` (previously known as `SimulatorActions`) away from `core` to `habitat_simulator` module.
- The above changes have been put in place to pave the way for implementing a `TaskActions`
 type API for `SimulatorAction`.
## How Has This Been Tested

1. Install python3 version of PyRobot ([branch](https://github.com/facebookresearch/pyrobot/tree/python3))([pyrobot.org](https://pyrobot.org)).
2. Example usage:
```python
import habitat
from habitat.sims import make_sim
import numpy as np

config = habitat.get_config()
config.freeze()

reality = make_sim(id_sim="PyRobot-v0", config=config.PYROBOT)

observations = reality.reset()
observations = reality.step(
    "go_to_relative",
    {
          "xyt_position": [0, 0, (10 / 180) * np.pi],
          "use_map": False,
          "close_loop": True,
          "smooth": False,            
     }
)

print("rgb:", observations["rgb"].shape)
print("depth:", observations["depth"].shape)
print("bump:", observations["bump"].shape)
```

## Pending
- [x] Action space usage, building on top of #196 
- [x] Fix `cv2` import.
- [x] Collision sensor.

speedmaster


habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,178 @@
...
+        if obs.shape != self.observation_space.shape:
+            obs = cv2.resize(
+                obs,
+                (
+                    self.observation_space.shape[1],
+                    self.observation_space.shape[0],
+                ),
+            )
+
+        obs = np.clip(obs, self.config.MIN_DEPTH, self.config.MAX_DEPTH)
```
mathfac: Looks like it should be in `else` statement for `if self.config.MAX_DEPTH:`, otherwise normalization doesn't make sense as `self.config.MAX_DEPTH = 1` if `self.config.MAX_DEPTH == True`.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,178 @@
...
+        return self._robot.base
+
+    @property
+    def camera(self):
+        return self._robot.camera
+
+    # TODO(akadian): add action space support.
+
+    def reset(self):
+        self._robot.camera.reset()
```
mathfac: What does `robot.camera.reset()` command do?

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,178 @@
...
+
+        observations = self._sensor_suite.get_observations(self._robot_obs())
+        return observations
+
+    def step(self, action, action_params):
```
mathfac: This method differs from `def step(self, action: int) -> Observations` form `Simulator` interface.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,178 @@
...
+            getattr(self._robot.camera, action)(**action_params)
+        else:
+            raise ValueError("Invalid action {}".format(action))
+
+        observations = self._sensor_suite.get_observations(self._robot_obs())
+
+        return observations
+
+    def seed(self, seed: int) -> None:
+        raise NotImplementedError("No support for seeding in reality")
```
mathfac: Maybe there is an opportunity to implement those methods as well form Simulator API 
  -  `def render(self, mode: str = "rgb") -> Any:` -> return observations from sensors
  - `def close(self) -> None`: -> clean up or just empty
  -  `def get_agent_state(self, agent_id: int = 0) -> any description of position/state that Pyrobot can provide 
  -  `def reconfigure(self, config: Config) -> None:` -> PyRobot reconfiguration for next run.

habitat/config/default.py
```diff
@@ -163,6 +163,40 @@
 _C.SIMULATOR.HABITAT_SIM_V0 = CN()
 _C.SIMULATOR.HABITAT_SIM_V0.GPU_DEVICE_ID = 0
 # -----------------------------------------------------------------------------
+# PYROBOT
+# -----------------------------------------------------------------------------
+_C.PYROBOT = CN()
+_C.PYROBOT.ROBOTS = ["locobot"]  # types of robots supported
+_C.PYROBOT.ROBOT = "locobot"
```
apsdehal: Maybe 
```suggestion
_C.PYROBOT.SELECTED_ROBOT = "locobot"
```

habitat/sims/pyrobot/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+        pyrobot_import_error = e
+
+    if has_pyrobot:
+        from habitat.sims.pyrobot.pyrobot import PyRobot
+    else:
+
+        @registry.register_simulator(name="PyRobot-v0")
+        class PyRobotImportError(Simulator):
+            def __init__(self, *args, **kwargs):
+                raise pyrobot_import_error
```
apsdehal: Shouldn't this happen inside the PyRobot class itself? I don't believe we need a separate check.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,178 @@
...
+    # TODO(akadian): add action space support.
+
+    def reset(self):
+        self._robot.camera.reset()
+
+        observations = self._sensor_suite.get_observations(self._robot_obs())
+        return observations
+
+    def step(self, action, action_params):
+        if action in self._robot_config.BASE_ACTIONS:
```
apsdehal: This technique of checking actions doesn't seem extensible. What if a new type of action comes in tomorrow? Will you add if check if you add one more action type?

habitat/utils/visualizations/maps.py
```diff
@@ -7,7 +7,14 @@
 import os
 from typing import List, Optional, Tuple
 
-import cv2
+# TODO(akadian): remove the below pyrobot hack
```
apsdehal: Seeing this at multiple places. Maybe move this to a function?

habitat/sims/pyrobot/__init__.py
erikwijmans: This patten lets you use `pyrobot` types in the type annotations in the definition of the class and in other related things.

apsdehal: If pyrobot is present in your PYTHONPATH, type annotator will automatically pick it up. Type annotations can be an issue for the custom created classes like `PyRobot` here. The import for `PyRobot` class should still happen in `__init__.py` but point of failure for initialization without PyRobot should be in init. Also, to use this class with type annotation you would still need to export it with `__all__` property.

erikwijmans: Yeah.  The point of failure become when you try to initialize the simulator.  The reason to have a class dedicated to the failure is that we want habitat to work correctly even if you don't have pyrobot installed.  We can't use the one we instantiate normally as if that has any pyrobot types on annotations, those will fail!

habitat/tasks/nav/nav_task.py
```diff
@@ -7,7 +7,15 @@
 from typing import Any, List, Optional, Type
 
 import attr
-import cv2
+
+# TODO(akadian): remove the below pyrobot hack
```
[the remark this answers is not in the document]
abhiskk: This is an artifact of PyRobot python3 which uses a backend running ROS in python2.7. As that path is present in the `sys.path` there  would be a conflict when trying to import cv2. I have localized this into a utils method with assert and error messages to make the import cleaner.

habitat/sims/pyrobot/pyrobot.py
abhiskk: I think the current implementation is correct, I will try to explain the usage below, please let me know if you think the interpretation is awry.

`MAX_DEPTH` and `NORMALIZE_DEPTH` are orthogonal. `MAX_DEPTH` specifies at what max depth rating should the depth sensor's observation be clipped. `NORMALIZE_DEPTH` flag controls whether the observations coming from the depth sensor should be in `[0, 1]` or `[MIN_DEPTH, MAX_DEPTH]`.

habitat/sims/pyrobot/pyrobot.py
abhiskk: It reset the physical camera to a 0 pitch and 0 yaw.

habitat/sims/pyrobot/pyrobot.py
abhiskk: Thanks, except `close` and `reconfigure` I have implemented the other methods as they don't have a good analogous on the pyrobot side.

abhiskk:
> Looks elegant! Left some comments. What are plans/possibilities to have testing for this integration?

As of now there is no way to do an automated testing of this. For now I think we can defer to the tests in PyRobot (https://github.com/facebookresearch/pyrobot/tree/master/tests). Also this integration right now uses a preliminary python3 branch from PyRobot which hasn't been merged into master. The PyRobot tests also require a physical Locobot for running.

The other option is we can write a mock PyRobot class and emulate the functionality in order for us to have CI on our side. If it sounds right I would like to push this as a future TODO instead of as a part of this PR.

kalyanvasudev:
> > Looks elegant! Left some comments. What are plans/possibilities to have testing for this integration?
> 
> As of now there is no way to do an automated testing of this. For now I think we can defer to the tests in PyRobot (https://github.com/facebookresearch/pyrobot/tree/master/tests). Also this integration right now uses a preliminary python3 branch from PyRobot which hasn't been merged into master. The PyRobot tests also require a physical Locobot for running.
> 
> The other option is we can write a mock PyRobot class and emulate the functionality in order for us to have CI on our side. If it sounds right I would like to push this as a future TODO instead of as a part of this PR.

You can run the PyRobot tests on the Gazebo simulator too. You dont need a real robot.

dhruvbatra:
> You can run the PyRobot tests on the Gazebo simulator too. You dont need a real robot.

Interesting! Are there any plan to integrate other simulators for such testing?

habitat/config/default.py
```diff
@@ -200,6 +200,47 @@
...
+# -----------------------------------------------------------------------------
+# BUMP SENSOR
+# -----------------------------------------------------------------------------
+_C.PYROBOT.BUMP_SENSOR = CN()
+_C.PYROBOT.BUMP_SENSOR.TYPE = "PyRobotBumpSensor"
+# -----------------------------------------------------------------------------
+# ACTIONS LOCOBOT
+# -----------------------------------------------------------------------------
+_C.PYROBOT.LOCOBOT = CN()
+_C.PYROBOT.LOCOBOT.ACTIONS = ["BASE_ACTIONS", "CAMERA_ACTIONS"]
```
abhiskk: This way of defining actions does not work well with how we have defined `NAVIGATION_ACTIONS` above (line 53). Until now as we have been closely tied to `habitat-sim` most of our API around the simulator has become tightly wound to it. I think being tightly wound on the environment level makes sense here but we should keep things open on the simulator level.

In my integration of PyRobot I have exposed the native actions available on the Robot so that the user has the flexibility to use these as it suits him/her.

habitat/core/simulator.py
```diff
@@ -307,7 +324,7 @@ def reset(self) -> Observations:
         """
         raise NotImplementedError
 
-    def step(self, action: int) -> Observations:
+    def step(self, action, *args, **kwargs) -> Observations:
```
abhiskk: This signature change is needed as we need a generic API here to support stepping from different simulators.

habitat/sims/pyrobot/pyrobot.py
abhiskk: Yes, this is needed. I have modified the type hinting in the abstract `Simulator` class to make it general. The modification of type signature in the abstract `Simulator` class will allow us to support differently signatured steps of Simulators.

habitat/sims/pyrobot/pyrobot.py
abhiskk: That's correct, @kalyanvasudev is there a way I can get a list of supported actions in PyRobot sot that we can avoid duplication of action definition on the habitat side, I tried having a look around but did not find anything.

kalyanvasudev:
> > You can run the PyRobot tests on the Gazebo simulator too. You dont need a real robot.
> 
> Interesting! Are there any plan to integrate other simulators for such testing?

We have V-Rep and Mujoco tight integrations for LoCoBot on the way.

habitat/sims/pyrobot/pyrobot.py
kalyanvasudev: No, there is no method that would return the available list of actions. I think what you have at the moment is fine or you could just change it to ```try: and except:``` routine.

mathfac:
Try to rebase on master to make `black` and `isort` happy.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def get_observation(self, robot_obs):
+        obs = robot_obs.get(self.uuid, None)
+
+        assert obs is not None, "Invalid observation for {} sensor".format(
+            self.uuid
+        )
+
+        if obs.shape != self.observation_space.shape:
+            obs = cv2.resize(
```
erikwijmans: Perhaps include the center-crop thing if the aspect ratios are different?

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+            "go_to_relative": spaces.Box(low=-np.inf, high=np.inf, shape=(3,)),
+            "go_to_absolute": spaces.Box(low=-np.inf, high=np.inf, shape=(3,)),
+        }
+    )
+
+
+def _locobot_camera_action_space():
+    return spaces.Dict(
+        {
+            "set_pan": spaces.Box(low=-np.inf, high=np.inf, shape=(1,)),
```
mathfac: What is `pan/tilt` unit of measurements? If that's degrees then we can specify proper limits.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def action_space(self) -> Space:
+        return self._action_space
+
+    def reset(self):
+        self._robot.camera.reset()
+
+        robot_obs = self.get_robot_observations()
+        observations = self._sensor_suite.get_observations(robot_obs)
```
mathfac:
```suggestion
        observations = self._sensor_suite.get_observations(robot_obs=robot_obs)
```

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+        observations = self._sensor_suite.get_observations(
+            self.get_robot_observations()
+        )
+
+        return observations
+
+    def render(self, mode: str = "rgb") -> Any:
+        robot_obs = self.get_robot_observations()
+        observations = self._sensor_suite.get_observations(robot_obs)
```
mathfac: If we add one more PyRobot sensor type that doesn't need `robot_obs` argument it will start failing. That's important to use keyword arguments using Suites:  
```suggestion
        observations = self._sensor_suite.get_observations(robot_obs=robot_obs)
```

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        if action in self._robot_config.BASE_ACTIONS:
+            getattr(self._robot.base, action)(**action_params)
+        elif action in self._robot_config.CAMERA_ACTIONS:
+            getattr(self._robot.camera, action)(**action_params)
+        else:
+            raise ValueError("Invalid action {}".format(action))
+
+        observations = self._sensor_suite.get_observations(
+            self.get_robot_observations()
```
mathfac:
```suggestion
            robot_obs=self.get_robot_observations()
```

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _get_observation_space(self, *args: Any, **kwargs: Any):
+        return spaces.Box(
+            low=self.min_depth_value,
+            high=self.max_depth_value,
+            shape=(self.config.HEIGHT, self.config.WIDTH, 1),
+            dtype=np.float32,
+        )
+
+    def get_observation(self, robot_obs):
```
mathfac:
```suggestion
    def get_observation(self, robot_obs, *args: Any, **kwargs: Any):
```

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+        return obs
+
+
+@registry.register_sensor
+class PyRobotBumpSensor(BumpSensor):
+    def _get_observation_space(self, *args: Any, **kwargs: Any):
+        return spaces.Box(low=False, high=True, shape=(1,), dtype=np.bool)
+
+    def get_observation(self, robot_obs):
```
mathfac:
```suggestion
    def get_observation(self, robot_obs, *args: Any, **kwargs: Any):
```

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _get_observation_space(self, *args: Any, **kwargs: Any):
+        return spaces.Box(
+            low=0,
+            high=255,
+            shape=(self.config.HEIGHT, self.config.WIDTH, 3),
+            dtype=np.uint8,
+        )
+
+    def get_observation(self, robot_obs):
```
mathfac: We need to keep sensor's method tolerant to new arguments that can be required by new added sensors: 
```suggestion
    def get_observation(self, robot_obs, *args: Any, **kwargs: Any):
```
Otherwise it will start throwing errors.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,281 @@
...
+        return self._action_space
+
+    def reset(self):
+        self._robot.camera.reset()
+
+        robot_obs = self.get_robot_observations()
+        observations = self._sensor_suite.get_observations(robot_obs)
+        return observations
+
+    def step(self, action, action_params):
```
mathfac: Maybe still have default for action without params.
```suggestion
    def step(self, action, action_params=None):
```
Overall, it looks expected `Simulator.step` follow `Task.step` interfaces with `step(action: Dict,...)`.

test/test_pyrobot.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+
+import numpy as np
+
+import mock
+from habitat.config.default import get_config
+from habitat.sims import make_sim
+
+
+class CameraMock:
```
mathfac: Really like Mock extendibility to test PyRobot integration without installing PyRobot.

test/test_pyrobot.py
```diff
@@ -0,0 +1,67 @@
...
+
+    config = get_config()
+    reality = make_sim("PyRobot-v0", config=config.PYROBOT)
+
+    observations = reality.reset()
+    observations = reality.step(
+        "go_to_relative",
+        {
+            "xyt_position": [0, 0, (10 / 180) * np.pi],
+            "use_map": False,
```
mathfac: `use_map`, `close_loop`, `smooth` parameters are not part of action space, I think it is expected.

habitat/core/simulator.py
```diff
@@ -229,6 +154,23 @@ def get_observation(self, *args: Any, **kwargs: Any):
         raise NotImplementedError
 
 
+class BumpSensor(Sensor):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
mathfac: Yes, we need to make sure that all `Sensor`'s children methods have `*args: Any, **kwargs: Any` in the end of argument lists.

habitat/sims/pyrobot/pyrobot.py
abhiskk: The PAN and TILT from pyrobot aren't restricted to an angle range like `(-pi, pi)`.

test/test_pyrobot.py
abhiskk: `use_map`, `close_loop`, `smooth` are extra parameters for the action, hence they haven't been included in the action space.

habitat/datasets/pointnav/__init__.py
```diff
@@ -3,3 +3,24 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+
+from habitat.core.registry import registry
+from habitat.core.dataset import Dataset
+
+
+def _try_register_pointnavdatasetv1():
```
apsdehal: Can we make it more generic? If end user wants to add their own datasets, this is unnecessary cumbersome that needs to be followed now. A wrapper for registration in this form and documentation update for adding datasets should suffice.

habitat/sims/pyrobot/pyrobot.py
```diff
@@ -0,0 +1,324 @@
...
+    RGBSensor,
+    SensorSuite,
+    Simulator,
+)
+from habitat.core.utils import try_cv2_import
+
+cv2 = try_cv2_import()
+
+
+def _center_crop(obs, observation_space):
```
apsdehal: Move to utils? Someone else might need this too.

test/test_pyrobot.py
apsdehal: As discussed offline, it will be better to mock the builder function itself to return mock of the class you are testing. This would ensure that you never have to mock anything on the PyRobot side.

test/test_pyrobot.py
```diff
@@ -0,0 +1,67 @@
...
+        mock_pyrobot.Robot = RobotMock
+        sys.modules["pyrobot"] = mock_pyrobot
+
+        # Re-register pyrobot with mock
+        from habitat.sims.registration import _try_register_pyrobot
+
+        _try_register_pyrobot()
+
+    config = get_config()
+    reality = make_sim("PyRobot-v0", config=config.PYROBOT)
```
apsdehal: This will be mocked here.

habitat/datasets/pointnav/__init__.py
mathfac: I can't imagine the situation when registration of any dataset class will fail. It will fail during instance initialization if required scenes or episode files are missing. I tend to just include `from habitat.datasets.pointnav.pointnav_dataset import PointNavDatasetV1` in `__init__.py` for registration.

mathfac:
Some tests like `test/test_sensors.py` aren't happy with task files renaming.

habitat/datasets/pointnav/__init__.py
abhiskk: Before this try catch things were failing if a user did not have `habitat-sim` installed and only had `pyrobot` installed. This is because in our package tree we were importing actions from `habitat-sim` which was a result of having an action space in our `core` module which used `habitat-sim`. In this PR I moved that `action_space` away from `core` but we still had dependencies in our other modules like task and dataset which were still calling `habitat-sim` actions. 

This is stopgap measure, ideally I should have implemented an action space inside our `core` and made `pyrobot` and `habitat-sim` actions derive from it. Will do that in another PR and these try-imports will go away then.

PR #200, opened 2019-09-12
Fix pausing logic

erikwijmans:
## Motivation and Context

Fixing a bug.  Also added back CPU support so I could do this locally.

## How Has This Been Tested

Locally and with the unit test since this is the second time I have fixed the pausing logic :)

## Types of changes

### DIFF
--- a/habitat_baselines/common/base_trainer.py
+++ b/habitat_baselines/common/base_trainer.py
@@ -69,7 +69,11 @@ class BaseRLTrainer(BaseTrainer):
         Returns:
             None
         """
-        self.device = torch.device("cuda", self.config.TORCH_GPU_ID)
+        self.device = (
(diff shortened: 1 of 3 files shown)


mathfac:
@erikwijmans what is bug about, any related issue?

erikwijmans:
It's about the lines here:
https://github.com/facebookresearch/habitat-api/blob/be177ba1ae1319b011f548210da7fcc218ef897e/habitat_baselines/rl/ppo/ppo_trainer.py#L519-L521

It should be 
```
test_recurrent_hidden_states = test_recurrent_hidden_states[ 
     :, state_index 
 ] 
```

I have fixed this before (I think as part of a PR), but its kinda difficult logic, so its easy to mess it up in a merge conflict or something, thus unit test.

PR #209, opened 2019-09-26
Fix bug with RGB frames

erikwijmans:
## Motivation and Context

Fix another bug like in #208.  Lists are also objects and `[[]] * 5` just creates a list of 5 pointers to the same thing.

## How Has This Been Tested

Locally

### DIFF
--- a/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -424,8 +424,8 @@ class PPOTrainer(BaseRLTrainer):
         stats_episodes = dict()  # dict of dicts that stores stats per episode
 
         rgb_frames = [
-            []
-        ] * self.config.NUM_PROCESSES  # type: List[List[np.ndarray]]
+            [] for _ in range(self.config.NUM_PROCESSES)
+        ]  # type: List[List[np.ndarray]]
         if len(self.config.VIDEO_OPTION) > 0:
             os.makedirs(self.config.VIDEO_DIR, exist_ok=True)
 


habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -424,8 +424,8 @@ def _eval_checkpoint(
         stats_episodes = dict()  # dict of dicts that stores stats per episode
 
         rgb_frames = [
-            []
-        ] * self.config.NUM_PROCESSES  # type: List[List[np.ndarray]]
+            [] for _ in range(self.config.NUM_PROCESSES)
+        ]  # type: List[List[np.ndarray]]
```
mathfac: why not:
```
        rgb_frames = [
            [] * self.config.NUM_PROCESSES
        ]   # type: List[List[np.ndarray]]
```

erikwijmans: `[] * N == []` .  `[] * N` just creates an empty list.  `lst * N` just does `flatten([lst for _ in range(N)])`

mathfac: That's surprise, as well as thing above:
```
>>> k =[[]] * 10
>>> k[0].append(1)
>>> k
[[1], [1], [1], [1], [1], [1], [1], [1], [1], [1]]
```

erikwijmans: The `k =[[]] * 10` thing makes sense.  That is basically doing
```python
k = []
k_ele = []
for _ in range(10):
    k.append(k_ele)
```
which looks very wrong.

PR #213, opened 2019-09-27
Eval on val and add an option to not take the checkpoints config

erikwijmans:
## Motivation and Context

The default should be to eval on val.  This PR makes it so that by default, `--run-type eval` will load-up the val split if there is one.

It also adds an option to not use the config from the checkpoint.  

It is currently actually impossible without modifying code to use habitat_baselines to run on the val set...... which seems wrong.....

## How Has This Been Tested

Locally

## Types of changes


habitat_baselines/config/default.py
```diff
@@ -37,6 +37,12 @@
 _C.LOG_FILE = "train.log"
 _C.CHECKPOINT_INTERVAL = 50
 # -----------------------------------------------------------------------------
+# EVAL CONFIG
+# -----------------------------------------------------------------------------
+_C.EVAL = CN()
+_C.EVAL.USE_VAL = True
```
mathfac: Maybe better make it configurable, then we can use different splits for evaluation, but have "val" as default:
```suggestion
_C.EVAL.SPLIT = "val"
```

erikwijmans: That seems like a better way to do it!

PR #216, opened 2019-09-28
Episode iterator upgrades

erikwijmans:
## Motivation and Context

Add some upgrades to the episode iterator functionality.  The primary thing this adds is the ability to shuffle scenes after some number of _steps_ are taken.  If you are shuffling after some number of episodes have ended, you end up shuffling too often at the start of training and not enough towards the end.  This value is set to 10k by default.  I consider this to be a "high but sane" default value.  Not so low that it will dramatically slow-down training, but not so high that scenes will never be swapped.

Also randomizes the shuffle switch values to be within `[0.8, 1.2]` of the configured value.  This helps not have all workers swap the same point for the shuffle by steps mode.  

Adds an initial shuffle to the list of episodes as otherwise the same scene will always be initially loaded (which is bad).

Sets the max steps per scene to a high but sane default value.  We generated a huge number of episodes per scene for training (50k IIRC) and you basically never switch scenes toward the end of training with that.

Also makes the scene order get shuffled by default as this makes sense for the default (people rarely change the defaults, so they should be set reasonably).

## How Has This Been Tested

Via the tests

## Types of changes


mathfac:
@JasonJiazhiZhang feel free to comment on the PR.

habitat/core/dataset.py
```diff
@@ -334,14 +349,11 @@ def __next__(self):
...
-        if (
-            self.max_scene_repetition > 0
-            and self._rep_count >= self.max_scene_repetition - 1
-        ):
-            self._shuffle_iterator()
+        if self._prev_scene_id != next_episode.scene_id:
             self._rep_count = 0
+            self._step_count = 0
+
+        self._switch_scene_if()
```
mathfac: `next_episode` is not updated based on results of `self._switch_scene_if()`.

habitat/core/dataset.py
```diff
@@ -361,3 +373,47 @@ def _shuffle_iterator(self) -> None:
...
             random.shuffle(episodes)
             self._iterator = iter(episodes)
+
+    def step_taken(self):
+        self._step_count += 1
+
+    def _set_shuffle_intervals(self):
+        if self.max_scene_repetition_episodes > 0:
+            self._max_rep_episode = random.randint(
+                int(0.8 * self.max_scene_repetition_episodes),
```
mathfac: Let's create `repetition_rand_interval = 0.2` param for `EpisodeIterator` and do:
```
            self._max_rep_step = random.randint(
                int((1 - self.repetition_rand_interval) * self.max_scene_repetition_steps),
                int((1 + self.repetition_rand_interval) * self.max_scene_repetition_steps),
            )
```
Or even create util method:
```
@static
def _randomize_value(value, interval):
    return random.randint(
                int((1 - interval) * value),
                int((1 + interval) * value),
            )
```

habitat/core/env.py
```diff
@@ -213,6 +213,11 @@ def _update_step_stats(self) -> None:
         if self._past_limit():
             self._episode_over = True
 
+        if self.episode_iterator is not None and isinstance(
```
mathfac: While `self.episode_iterator` mentioned as Optional I don't see Env functioning without it here. Maybe check for subclass of `EpisodeIterator`.

habitat/core/dataset.py
```diff
@@ -361,3 +373,47 @@ def _shuffle_iterator(self) -> None:
...
+
+        # Shuffle if a scene has been used for more than _max_rep_step steps in a row
+        if (
+            self._max_rep_step is not None
+            and self._step_count >= self._max_rep_step
+        ):
+            do_switch = True
+
+        if do_switch:
+            self._shuffle_iterator()
```
mathfac: Should we check for `self.shuffle == True`, here when we shuffling.

erikwijmans: No.  The `_shuffle_iterator` is used both to initiate a scene switch and to just the shuffle the episodes.  `self.shuffle` decides whether or not to shuffle the episode order on cycle or on load.

habitat/core/env.py
erikwijmans: A separate PR probably makes sense for changing that functionality (dataset is also marked as optional).

habitat/core/dataset.py
mathfac: That's strange as there was no option to enable `cycling` without episode `shuffling`.

erikwijmans: The cycle logic doesn't really on the shuffle

habitat/core/dataset.py
```diff
@@ -290,32 +291,43 @@ def __init__(
...
         :param group_by_scene: if :py:`True`, group episodes from same scene.
-        :param max_scene_repeat: threshold of how many episodes from the same
+        :param max_scene_repeat_episodes: threshold of how many episodes from the same
             scene can be loaded consecutively. :py:`-1` for no limit
+        :param max_scene_repeat_steps: threshold of how many steps from the same
+            scene can be taken consecutively. :py:`-1` for no limit
         :param num_episode_sample: number of episodes to be sampled. :py:`-1`
             for no sampling.
         """
+        self._repetition_rand_interval = 0.2
```
mathfac: Let's move `repetition_rand_interval` to `init` argument with default value, otherwise no option to turn it off.

habitat/core/dataset.py
```diff
@@ -278,8 +278,10 @@ def __init__(
         cycle: bool = True,
         shuffle: bool = False,
         group_by_scene: bool = True,
-        max_scene_repeat: int = -1,
+        max_scene_repeat_episodes: int = -1,
+        max_scene_repeat_steps: int = -1,
         num_episode_sample: int = -1,
+        repetition_rand_interval: float = 0.2,
```
JasonJiazhiZhang: Maybe add some explanation to `repetition_rand_interval ` in the docstring since one might not be able to tell what it does just by the variable name.  Also, is there a better name for it, since it's a bit weird to name a `float` type "interval".

habitat/core/dataset.py
```diff
@@ -355,9 +367,65 @@ def _shuffle_iterator(self) -> None:
...
+
+    @staticmethod
+    def _randomize_value(value, interval):
+        return random.randint(
+            int(value * (1 - interval)), int(value * (1 + interval))
+        )
+
+    def _set_shuffle_intervals(self):
+        if self.max_scene_repetition_episodes > 0:
+            self._max_rep_episode = self._randomize_value(
```
JasonJiazhiZhang: I'm not sure if we need to randomize the episode rep count. I understand that it is helpful for step rep count but it's kind of counter-intuitive for episode rep count, with a default value of `0.2`.

habitat/core/dataset.py
```diff
@@ -355,9 +367,65 @@ def _shuffle_iterator(self) -> None:
...
+            self._max_rep_episode = None
+
+        if self.max_scene_repetition_steps > 0:
+            self._max_rep_step = self._randomize_value(
+                self.max_scene_repetition_steps, self.repetition_rand_interval
+            )
+        else:
+            self._max_rep_step = None
+
+    def _switch_scene_if(self):
```
JasonJiazhiZhang: Question here: would it be switching to frequently at the later stage of training? with potentially less than 100 episode per scene switch? Will it help to incorporate both count schemes, like a logical AND?

erikwijmans: Shuffling on steps is consistent from an optimization standpoint, you swap scenes every N parameters updates, which is why I like it :)

I don't think you can shuffle scenes "too often".  Ideally, we just randomly sample a new episode irrespective of the scene it is in, but this incurs the non-trivial cost of swapping the scene way too often.

habitat/core/dataset.py
erikwijmans: Makes sense to rename it and not do it the episode count method.

habitat/core/dataset.py
```diff
@@ -355,9 +392,57 @@ def _shuffle_iterator(self) -> None:
                 list(g)
                 for k, g in groupby(self._iterator, key=lambda x: x.scene_id)
             ]
+
             random.shuffle(grouped_episodes)
+            for i in range(len(grouped_episodes)):
```
mathfac: Maybe use same approach to shuffling as in init method:
```
        random.shuffle(self.episodes)
        if group_by_scene:
            self.episodes = sorted(self.episodes, key=lambda x: x.scene_id)
```

habitat/core/dataset.py
```diff
@@ -324,29 +342,48 @@ def __next__(self):
...
+            self._step_count = 0
 
         self._prev_scene_id = next_episode.scene_id
         return next_episode
 
-    def _shuffle_iterator(self) -> None:
+    def _switch_scene(self) -> None:
+        r"""Internal method to switch the scene
+        """
+        grouped_episodes = [
```
mathfac: It looks like this part of code doesn't respect `group_by_scene` as well as `shuffle` option.

erikwijmans: `groupby` creates a new group everytime the key changes and groups only get re-ordered if `shuffle` is true.

habitat/core/dataset.py
erikwijmans: Yeah, makes sense to do that here instead.  The more complicated shuffle logic to be `O(n)` instead of `O(n log(n))` was worth it when it was being called on every swap, but this now gets called very rarely.

habitat/core/dataset.py
erikwijmans: So it'll respect both things :)

mathfac: I meant situation when `group_by_scene = False` is not considered.

habitat/core/dataset.py
```diff
@@ -324,40 +342,100 @@ def __next__(self):
...
+        ]
+
+        if len(grouped_episodes) > 1:
+            # Ensure we swap by moving the current group to the end
+            if self.shuffle:
+                random.shuffle(grouped_episodes[1:])
+
+            grouped_episodes = grouped_episodes[1:] + grouped_episodes[0:1]
+
+        self._iterator = iter(sum(grouped_episodes, []))
```
mathfac: Will the iterator reach the end anytime? Looks like it will be reset each time.

erikwijmans: `grouped_episodes` is computed by running out `self._iterator`, thus the length of `self._iterator` before and after this function will be exactly the same and it will eventually hit the end.

mathfac:
Pulled PR to understand the logic of changes. As result added some tests and small fix.

habitat/core/dataset.py
```diff
@@ -376,9 +377,6 @@ def _switch_scene(self) -> None:
 
         if len(grouped_episodes) > 1:
             # Ensure we swap by moving the current group to the end
-            if self.shuffle:
```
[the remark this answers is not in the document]
mathfac: Because, shuffles are already happening at the beginning and on each cycle. That's why this shuffle is redundant and can be omitted with no test failing.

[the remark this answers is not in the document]
erikwijmans: Yeah, I guess both logics are valid.  Either you cycle through all scenes in a predetermined but random order, or you randomly draw a new scene from the set of other scenes.  One doesn't seem inherently better than the other.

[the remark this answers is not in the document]
mathfac: My point is that it's already in `predetermined but random order` because of `self.shuffle` in other places. So there is no need to shuffle again.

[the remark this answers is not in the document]
erikwijmans: Yep yep, it's just different ideas of how scenes should be selected on a swap.  I am fine with this way if it makes more sense to you, both make sense to me.

habitat/core/dataset.py
erikwijmans: Yep, it'll also be fine in that situation. `groupby` doesn't do exactly what you would think, i.e.

```python
>> lst = ["a", "a", "b", "a"]
>> grouped = [list(g) for k, g in groupby(lst, lambda x: x)]
>> grouped
[['a', 'a'], ['b'], ['a']]
```

PR #219, opened 2019-09-30
docs: separate public config with GA and improved search

mosra:
## Motivation and Context

As requested by @abhiskk. Adding GA tracking to the docs, but to avoid privacy issues in a local build, there's now a separate public config. Additionally, for the public version the search downloads a binary data (which is smaller, but won't work locally), plus also provides a helpful link to Google search if nothing is found.

## How Has This Been Tested

I pressed some keys and it made pixels shine in a correct way.

### DIFF
--- a/docs/build-public.sh
+++ b/docs/build-public.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Propagate failures properly
+set -e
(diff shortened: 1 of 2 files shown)


mathfac:
Thank you, merging!

PR #256, opened 2019-11-22
Fix the doc build

erikwijmans:
## Motivation and Context

Fix the doc build

## How Has This Been Tested

Build


erikwijmans:
in `conf-public.py`, the linking to habitat-sim's docs needs to be done with `../../habitat-sim/build/docs-public/habitat-sim/objects.inv`, not `../../habitat-sim/build/docs/habitat-sim/objects.inv`

PR #272, opened 2020-01-08
Fix contains method in space classes

### LINKED ISSUE #271: Fix contains method in habitat space classes
## 🐛 Bug

The `contains` method for the `EmptySpace`, `ActionSpace`, and `ListSpace` classes in `habitat.core.spaces` all contain errors. Specifically:
1. `EmptySpace.contains(x)` should return `True` if `x is None`.
2. `ActionSpace.contains(x)` improperly handles invalid keys and invalid `x["action"]` values.
3. `ListSpace.contains(x)` improperly checks the length of `x`.

## Command

`EmptySpace.contains()`
`ActionSpace.contains()`
`ListSpace.contains()`

## To Reproduce

Code to reproduce the behavior:

```python
>>> import gym
>>> from habitat.core.spaces import EmptySpace, ActionSpace, ListSpace
>>>
>>> space = EmptySpace()
>>> print(space.contains(None))  # should be True
False
>>>
>>> space = ActionSpace({
...     "move": gym.spaces.Dict({
...         "position": gym.spaces.Discrete(2),
...         "velocity": gym.spaces.Discrete(3)
...     }),
...     "move_forward": EmptySpace(),
... })
>>> space.contains({'action': 'move'}) # should be False
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/habitat/habitat-api/habitat/core/spaces.py", line 64, in contains
    if not self.spaces[x["action"]].contains(x["action_args"]):
KeyError: 'action_args'
>>>
>>> space.contains({'action': None})  # should be false
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/habitat/habitat-api/habitat/core/spaces.py", line 64, in contains
    if not self.spaces[x["action"]].contains(x["action_args"]):
KeyError: None
>>>
>>> space = ListSpace(gym.spaces.Discrete(2), 5, 10)
>>> print(space.contains([0, 1, 0, 1]))  # should be True
False
```

## Expected behavior

The `contains` methods should properly return `True`, `False`, `False`, and `True` in the four cases above, respectively.

## Additional context

There is also an error in example code in the docs for the `ActionSpace` class.

arjunmajum:
## Motivation and Context

The `contains` methods in the three `habitat.core.spaces` classes (`EmptySpace`, `ActionSpace`, and `ListSpace`) have errors that are described in #271. This pull request fixes those errors, corrects the documentation for the `ActionSpace` class, and adds a `__repr__` to the `EmptySpace` class.

closes #271 

## How Has This Been Tested

This code was test by passing both valid and invalid samples to the `contains` method of instantiations of each of the three classes mentioned above.

## Types of changes

## Checklist

### DIFF
--- a/habitat/core/spaces.py
+++ b/habitat/core/spaces.py
@@ -63,16 +63,12 @@ class ActionSpace(gym.spaces.Dict):
         }
 
     def contains(self, x):
-        if not isinstance(x, dict):
+        if not isinstance(x, dict) or "action" not in x:
             return False
-        if not all(key in x for key in {"action", "action_args"}):
-            return False
-
         if x["action"] not in self.spaces:
             return False
-        if not self.spaces[x["action"]].contains(x["action_args"]):
+        if not self.spaces[x["action"]].contains(x.get("action_args", None)):
             return False
-
         return True
 
     def __repr__(self):
--- a/habitat/core/spaces.py
+++ b/habitat/core/spaces.py
@@ -21,7 +21,12 @@ class EmptySpace(Space):
         return None
 
     def contains(self, x):
+        if x is None:
+            return True
         return False
+
+    def __repr__(self):
+        return "EmptySpace()"
 
 
 class ActionSpace(gym.spaces.Dict):
@@ -30,14 +35,13 @@ class ActionSpace(gym.spaces.Dict):
 
     .. code:: py
 
-        self.observation_space = spaces.ActionSpace(
+        self.observation_space = spaces.ActionSpace({
             "move": spaces.Dict({
                 "position": spaces.Discrete(2),
                 "velocity": spaces.Discrete(3)
-                },
-            "move_forward": EmptySpace,
-            )
-        )
+            }),
+            "move_forward": EmptySpace(),
+        })
     """
 
     def __init__(self, spaces):
@@ -59,10 +63,16 @@ class ActionSpace(gym.spaces.Dict):
         }
 
     def contains(self, x):
-        if not isinstance(x, dict) and {"action", "action_args"} not in x:
+        if not isinstance(x, dict):
+            return False
+        if not all(key in x for key in {"action", "action_args"}):
+            return False
+
+        if x["action"] not in self.spaces:
             return False
         if not self.spaces[x["action"]].contains(x["action_args"]):
             return False
+
         return True
 
     def __repr__(self):
@@ -100,7 +110,7 @@ class ListSpace(Space):
         if not isinstance(x, Sized):
             return False
 
-        if self.min_seq_length <= len(x) <= self.max_seq_length:
+        if len(x) < self.min_seq_length or len(x) > self.max_seq_length:
             return False
 
         return all([self.space.contains(el) for el in x])
--- a/habitat/core/spaces.py
+++ b/habitat/core/spaces.py
@@ -106,7 +106,7 @@ class ListSpace(Space):
         if not isinstance(x, Sized):
             return False
 
-        if len(x) < self.min_seq_length or len(x) > self.max_seq_length:
+        if not (self.min_seq_length <= len(x) <= self.max_seq_length):
             return False
 
         return all([self.space.contains(el) for el in x])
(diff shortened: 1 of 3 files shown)


habitat/core/spaces.py
```diff
@@ -59,10 +63,16 @@ def sample(self):
         }
 
     def contains(self, x):
-        if not isinstance(x, dict) and {"action", "action_args"} not in x:
+        if not isinstance(x, dict):
+            return False
+        if not all(key in x for key in {"action", "action_args"}):
```
erikwijmans:
```suggestion
        if not set(x.keys()).issubset({"action", "action_args"}):
```

erikwijmans: Actually, looking at the code bellow, it seems to assume that `x` has both "action" and "action_args" keys, so maybe this should just be `x.keys() == {"action", "action_args"}`?

CC @mathfac

arjunmajum: I agree, I think it makes sense to restrict it to the two keys.

erikwijmans: To nicely support empty space, it seems like `x["action_args"]` -> `x.get("action_args", None)` would be better.

arjunmajum: Good point, that would simplify empty space actions from `x = {"action": "move_forward", "action_args": None}` to `x = {"action": "move_forward"}`.

arjunmajum: I made this change.

arjunmajum:
I added some basic tests for these classes. Let me know if there is anything obvious missing.

habitat/core/spaces.py
```diff
@@ -100,7 +106,7 @@ def contains(self, x):
         if not isinstance(x, Sized):
             return False
 
-        if self.min_seq_length <= len(x) <= self.max_seq_length:
+        if len(x) < self.min_seq_length or len(x) > self.max_seq_length:
```
mathfac: Maybe `if not( self.min_seq_length <= len(x) <= self.max_seq_length):` for shorter form.

test/test_habitat_task.py
```diff
@@ -53,6 +54,7 @@ def test_task_actions_sampling_for_teleport():
     config.freeze()
 
     env = habitat.Env(config=config)
+    assert env.action_space.contains(env.action_space.sample())
     env.reset()
     while not env.episode_over:
         action = sample_non_stop_action(env.action_space)
```
mathfac: That maybe even better to move line 57 here:
 `assert env.action_space.contains(action)`

test/test_habitat_task.py
```diff
@@ -86,6 +88,7 @@ def test_task_actions_sampling(config_file):
         )
 
     env = habitat.Env(config=config)
+    assert env.action_space.contains(env.action_space.sample())
     env.reset()
     while not env.episode_over:
         action = sample_non_stop_action(env.action_space)
```
mathfac: Same here:
```suggestion
         action = sample_non_stop_action(env.action_space)
         assert env.action_space.contains(action)
```

test/test_habitat_task.py
```diff
@@ -24,6 +24,7 @@ def test_task_actions():
     config.freeze()
 
     env = habitat.Env(config=config)
+    assert env.action_space.contains(env.action_space.sample())
     env.reset()
     env.step(
         action={
```
mathfac: That maybe even better to move line 27 here like:
`env.action_space.contains(action)`

arjunmajum:
@mathfac I made the changes that you suggested above and it revealed a minor issue...

I had to change the `TELEPORT_POSITION` and `TELEPORT_ROTATION` to `np.arrays` to pass the `Space.contains()` check. However, it looks like in most of the library these types of objects are usually lists.

I will take a look tomorrow at why the `ActionSpace` expects a `np.array`, and let you know what I find.

arjunmajum:
It turns out that `TELEPORT_POSITION` and `TELEPORT_ROTATION` are `gym.spaces.Box` spaces so they should be `np.arrays`.

The action spaces is defined here: https://github.com/facebookresearch/habitat-api/blob/master/habitat/tasks/nav/nav.py#L873

mathfac:
Yes, looks like `Box` space definition is reason for that. Thank you for fixing last bits. Merging.

PR #275, opened 2020-01-14
Allow free form yaml configs: enable allow_new flag to default config point

mathfac:
## Motivation and Context
There was multiple request, that our config has pretty rigid structure: no new fields allowed unless it specialized in default config or created by code. This PR allow free form yaml configs to be loaded:
- Added allow_new flag to default config point
- Added test to cover the change.

## Types of changes

## Checklist


habitat/config/default.py
```diff
@@ -6,7 +6,15 @@
 
 from typing import List, Optional, Union
 
-from habitat.config import Config as CN  # type: ignore
+import yacs.config
+
+# from habitat.config import Config as CN # type: ignore
+
+# Default config node
+Config = lambda *args, **kwargs: yacs.config.CfgNode(
```
erikwijmans: We should do changing the default value of `new_allowed` via a subclass.  That way static analysis, code completion, and mypy will still work correctly.

mathfac:
Addressed the comments.

Issue #281, opened 2020-01-23
Configure RGB Input


saketd403:
## ❓ Questions and Help
Hi,
    I am trying to replicate the results from Habitat paper for various modes of inputs such as RGBD, RGB, Blind. The agent is initially configured for RGBD input. It works well for RGBD input. In order to change the input to RGB, I changed the ppo_pointnav.yaml config file attribute as SENSORS:["RGB_SENSOR"], hence removing the depth sensor. But this causes problem in training the ppo agent as the reward is stuck at -0.009766 and is not improving. Does anything else needs to be changed to configure RGB input?  Also, please let me know the hyperparameters for obtaining the results in Habitat paper for RGB input.

Thanks,
Saurabh

saketd403:
Hi,
When I run this with new config file, the code still does not perform well for RGB input. Can you tell me is there something that I am doing wrong while trying to configure the code for getting RGB input? I have configured the code for getting depth as the only input and it worked well.

Thanks,
Saurabh

erikwijmans:
It looks like the modified for 1 GPU hyperparameters don't work for RGB unfortunately.  Changing the number of processes to 8 and the number of mini batches to 4 (as in the paper) makes things work for me.

saketd403:
Okay. I will try that. By the way in order to get RGB input you just need to change "SENSORS" field in config right? That's how I am doing it.
Thanks,
Saurabh

erikwijmans:
Yes, that’s correct.

saketd403:
Thank you.

PR #282, opened 2020-01-23
Make the defaults in the config better

erikwijmans:
## Motivation and Context

The default values in `ppo_pointnav.yaml` are nice as for a "just see if it runs" experiment, but are not well chosen for training something meaningful.  This PR changes the default values in `ppo_pointnav.yaml` to be closer to what was used in the paper (modified to work for 1 GPU).  The old `ppo_pointnav.yaml` is now `ppo_pointnav_example.yaml`

This is likely the issue in #281 .

## How Has This Been Tested

Locally.

## Types of changes

### DIFF
--- a/habitat_baselines/config/pointnav/ppo_pointnav.yaml
+++ b/habitat_baselines/config/pointnav/ppo_pointnav.yaml
@@ -14,7 +14,10 @@ TEST_EPISODE_COUNT: 994
 EVAL_CKPT_PATH_DIR: "data/new_checkpoints"
 # This was 6 for mp3d and 8 for gibson in the paper
 NUM_PROCESSES: 4
-SENSORS: ["RGB_SENSOR", "DEPTH_SENSOR"]
+# Note:  To train the an RGB only model,
+# you may need to use 8 processes with 4 mini batches,
+# If so, the number of updates should be cut in half
+SENSORS: ["DEPTH_SENSOR"]
 CHECKPOINT_FOLDER: "data/new_checkpoints"
 NUM_UPDATES: 270000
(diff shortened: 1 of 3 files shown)


saketd403:
Hi, 
       When I run this with new config file, the code still does not perform well for RGB input. Can you tell me is there something that I am doing wrong while trying to configure the code for getting RGB input? I have configured the code for getting depth as the only input and it worked well.

Thanks,
Saurabh

habitat_baselines/config/pointnav/ppo_pointnav.yaml
```diff
@@ -1,26 +1,35 @@
...
+# Note:  Hyperparameters have been changed slightly from
+# the paper to allow for things to easily run on 1 GPU
+
+BASE_TASK_CONFIG_PATH: "configs/tasks/pointnav_gibson.yaml"
 TRAINER_NAME: "ppo"
 ENV_NAME: "NavRLEnv"
 SIMULATOR_GPU_ID: 0
 TORCH_GPU_ID: 0
-VIDEO_OPTION: ["disk", "tensorboard"]
+# VIDEO_OPTION: ["disk", "tensorboard"]
```
mathfac: Please, remove or leave it uncommented.

erikwijmans: I left it there for easy video generation for people, but having it be the default when there are 994 episodes to evaluate on is not great :)

mathfac: Let's add comment, something like "Enable video generation for debugging purposes."

Issue #277, opened 2020-01-17
How to enable physics when initialising the environment class ?


ronjamoller:
I am working on a setup that requires inserted objects so i need the physics_enabled flag and i can get it to run if i set up the simulation myself. However, i have not found a way to set up an environment in such a way that the simulation that it owns has physics enabled. From what i can gather there seems to be no physics flag in the environment config and i have been unsuccessful in reinitializing the env._sim._sim to include physics as well. Is there something that i'm missing ?
Thank you

smorad:
This is relevant to me. Is there a way to enable the physics sim via config?

ronjamoller:
I have not found a way to do this correctly, my current solution is a workaround: 
env = habitat.Env(config=config)
env._sim._sim.reconfigure(sim_cfg)
that is pretty much the thing i talked about in the above post - you initialize the environment normally (config, see habitat api demo from the habitat docs page)  and then reinitialize the internal simulation using a simulation configuration (sim_cfg, there are examples i think in the habitat-sim demo on the habitat doc page and in the example/demo_runner in the example folder of habitat-sim). My main issue was that there is no consistency in terms of naming - for example in one config the camera might be called rgb, in another it is called rgb_sensor so if you get naming mismatches during the reinitialization check if you can for example rename keys in you sim_config dict to match up with what the environment config expects. I highly suspect that i am using a very hacky solution and while i can run the env and walk around in it and insert objects i have not trained with it yet, so i cannot guarantee anything. Also make sure to pull the newest release, there was quite a change. best of luck :)

smorad:
Great, I'd like to patch it correctly. Will hopefully submit a patch to add sim options to the API config tonight.

PR #289, opened 2020-01-30
Pass physics option to sim

smorad:
See #277 for some discussion.

This new feature allows for habitat-api to start a physics-enabled `sim`. Tested by inserting objects from the habitat API. Ran `mypy` and `black` before commit.

### DIFF
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -249,6 +249,10 @@ _C.SIMULATOR.HABITAT_SIM_V0.GPU_DEVICE_ID = 0
 _C.SIMULATOR.HABITAT_SIM_V0.GPU_GPU = False
 # Whether or not the agent slides on collisions
 _C.SIMULATOR.HABITAT_SIM_V0.ALLOW_SLIDING = True
+_C.SIMULATOR.HABITAT_SIM_V0.ENABLE_PHYSICS = False
+_C.SIMULATOR.HABITAT_SIM_V0.PHYSICS_CONFIG_FILE = (
+    "./data/default.phys_scene_config.json"
+)
 # -----------------------------------------------------------------------------
 # PYROBOT
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mathfac:
Thank you @smorad and @ronjamoller for this first step to support physics in Habitat-API/Tasks as well. Looks good to me! Will wait for @aclegg3 accept as well.

aclegg3:
> Thank you @smorad and @ronjamoller for this first step to support physics in Habitat-API/Tasks as well. Looks good to me! Will wait for @aclegg3 accept as well.
> Let's add a smoke test to make sure physics setup is working: `test/test_habitat_physics.py` ([for motivation physics Python test in HSIM](https://github.com/facebookresearch/habitat-sim/blob/master/tests/test_physics.py)).

Agreed, I'm happy to accept this change. Thanks for working on this. 👍

mathfac:
@smorad let us know if it's ok to add a smoke test to make sure physics setup is working: test/test_habitat_physics.py (for motivation physics Python test in HSIM). Otherwise, someone else can do that.

smorad:
@mathfac Unfortunately, I'll be pretty busy this weekend. Hopefully I can write up a test sometime next week.

ronjamoller:
hi @smorad im coming back to this thread because i want to train in environments with inserted objects, do you by any chance still have the code of the object insertion via the api you mentioned in the first comment ? I would like to avoid using my sim-based workaround if a better and "proper" version exists. Thank you!

smorad:
@ronjamoller this change was merged into master. Make sure you update to master. Then you can set `SIMULATOR.HABITAT_SIM_V0.PHYSICS_CONFIG_FILE`

ronjamoller:
Hi, thank you for getting back to me so quickly - my issue is not activating physics options/setting the file, that works nicely thank you - i am interested in how to insert an object using the env class (im assuming thats what you meant when you said "Tested by inserting objects from the habitat API"). I only know/found examples with sim and while i can use env._sim._sim.add_object(obj_index), i dont think thats proper usage of the api but maybe i misunderstood.

smorad:
That's exactly how I do it. I believe the index refers to the ordering of the objects in the `PHYSICS_CONFIG_FILE`. The API itself does not appear to offer that functionality, you need to do it in the sim.

ronjamoller:
Okay, thank you !

PR #290, opened 2020-01-30
Added interface to calculate geodesic distance to multiple goals

mathfac:
## Motivation and Context
Added interface to calculate geodesic distance to multiple points. The functionality is super convenient for multi target navigation and speed up metrics calculation process. 
- We still support single point functionality through API and are fully backward compatible.
- Switched existing code to multi goal function usage using a single element list.

## How Has This Been Tested
Added relevant test.

## Types of changes

### DIFF
--- a/habitat/datasets/pointnav/pointnav_generator.py
+++ b/habitat/datasets/pointnav/pointnav_generator.py
@@ -40,7 +40,7 @@ def is_compatible_episode(
     if np.abs(s[1] - t[1]) > 0.5:  # check height difference to assure s and
         #  t are from same floor
         return False, 0
-    d_separation = sim.geodesic_distance(s, t)
+    d_separation = sim.geodesic_distance(s, [t])
     if d_separation == np.inf:
         return False, 0
     if not near_dist <= d_separation <= far_dist:
--- a/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Union
 
 import numpy as np
 from gym import spaces
@@ -281,10 +281,19 @@ class HabitatSim(Simulator):
         self._update_agents_state()
 
     def geodesic_distance(self, position_a, position_b):
-        path = habitat_sim.ShortestPath()
+        path = habitat_sim.MultiGoalShortestPath()
         path.requested_start = np.array(position_a, dtype=np.float32)
-        path.requested_end = np.array(position_b, dtype=np.float32)
+        if isinstance(position_b[0], List) or isinstance(
+            position_b[0], np.ndarray
+        ):
+            path.requested_ends = np.array(position_b, dtype=np.float32)
+        else:
+            path.requested_ends = np.array(
+                [np.array(position_b, dtype=np.float32)]
+            )
+
         self._sim.pathfinder.find_path(path)
+
         return path.geodesic_distance
 
     def action_space_shortest_path(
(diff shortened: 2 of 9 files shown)


habitat/datasets/pointnav/pointnav_generator.py
```diff
@@ -40,7 +40,7 @@ def is_compatible_episode(
     if np.abs(s[1] - t[1]) > 0.5:  # check height difference to assure s and
         #  t are from same floor
         return False, 0
-    d_separation = sim.geodesic_distance(s, t)
+    d_separation = sim.geodesic_distance(s, [t])
```
erikwijmans: The `[t]` seems unnecessary, it should work with just `t` right?

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -281,9 +281,16 @@ def reconfigure(self, config: Config) -> None:
         self._update_agents_state()
 
     def geodesic_distance(self, position_a, position_b):
-        path = habitat_sim.ShortestPath()
-        path.requested_start = np.array(position_a, dtype=np.float32)
-        path.requested_end = np.array(position_b, dtype=np.float32)
+        path = habitat_sim.MultiGoalShortestPath()
```
erikwijmans: Still needs the `path.requested_start = np.array(position_a, dtype=np.float32)`

habitat/datasets/pointnav/pointnav_generator.py
mathfac: Yes, but we want to promote usage of `[t]` for future code.

test/test_habitat_sim.py
```diff
@@ -82,3 +82,22 @@ def test_sim_no_sensors():
...
     sim.reset()
     sim.close()
+
+
+def test_sim_geodesic_distance():
+    config = get_config()
+    if not os.path.exists(config.SIMULATOR.SCENE):
+        pytest.skip("Please download Habitat test data to data folder.")
+    sim = make_sim(config.SIMULATOR.TYPE, config=config.SIMULATOR)
+    sim.seed(0)
```
mathfac: Thank you, @mpiseno! Now, we can write test like these using `seed` and be user there will be no false positives.

test/test_habitat_sim.py
```diff
@@ -82,3 +82,28 @@ def test_sim_no_sensors():
...
+    navigable_points = [sim.sample_navigable_point() for _ in range(10)]
+    assert np.isclose(
+        sim.geodesic_distance(navigable_points[0], navigable_points[1]),
+        1.3849650,
+    ), "Geodesic distance or sample navigable points mechanism has been changed."
+    assert np.isclose(
+        sim.geodesic_distance(navigable_points[0], navigable_points[1:]),
+        1.3849650,
+    ), "Geodesic distance or sample navigable points mechanism has been changed."
+    assert sim.geodesic_distance(
```
mathfac: @erikwijmans, this test is failing and maybe surfacing an issue in MultiGoal calculations or current usage of the functionality:
```
E       AssertionError: Geodesic distance for multi target setup isn't equal to separate single target calls.
E       assert 1.3849650621414185 == [0.6194838285446167]
```

erikwijmans: Interesting that it is off by that much.  Unfortunately the multi-goal version won't be perfectly accurate as the closest target is selected in an intermediate step and before things are completely optimized, so it can select the wrong one.  In the little bit of testing I did, it was never off by that much, but I may have overfit to my scene/test cases. Guess we should use the version that is slower but exact (do a full dijkstra expand, get all paths, run the min over that, it will still be faster than running the N calls to A*, but slower than the single A* call done currently).

mathfac: Yes, to use this function reliably make sense to do it slower, but more accurate. For now I will comment out the last test and leave it for you to enable, once MultiGoal evaluation will be changed.

erikwijmans: Changed the multigoal to min(A*) in this PR: https://github.com/facebookresearch/habitat-sim/pull/465

Same functionality as you have here, but considerably faster as it's pushed all the way down in the stack.

mathfac:
Rebased on latest sim changes, switched back to Multigoal usage.

PR #309, opened 2020-02-20
De-duplicate goals for objectnav

erikwijmans:
## Motivation and Context

The object nav datasets are huge, both on disk and in memory.  This is largely due to the goals being duplicated across different episodes to the same object category.  This PR resolves this by de-duplicating the goals.

Script to dedup current episodes: https://gist.github.com/erikwijmans/49925738798c5ced3d50e12e3d13b491

## How Has This Been Tested

It loads the episodes!

## Types of changes

### DIFF
--- a/habitat/datasets/object_nav/object_nav_dataset.py
+++ b/habitat/datasets/object_nav/object_nav_dataset.py
@@ -30,7 +30,26 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
     content_scenes_path: str = "{data_path}/content/{scene}.json.gz"
 
     def to_json(self) -> str:
+        for ep in self.episodes:
+            ep.goals = str(ep.goals[0].object_id)
+
+        self.goals_per_category = {}
+        for i, ep in enumerate(self.episodes):
+            obj_id = str(ep.goals[0].object_id)
+            if obj_id not in self.goals_per_category:
+                self.goals_per_category[obj_id] = ep.goals
+
+            self.episodes[i].goals = obj_id
+
         result = DatasetFloatJSONEncoder().encode(self)
+
+        for i in range(len(self.episodes)):
+            self.episodes[i].goals = self.goals_per_category[
+                self.episodes[i].goals
+            ]
+
+        del self.goals_per_category
+
         return result
 
     def from_json(
@@ -63,6 +82,19 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
             self.category_to_scene_annotation_category_id.keys()
         ), "category_to_task and category_to_mp3d must have the same keys"
 
+        goals_by_category = deserialized["goals_by_category"]
+
+        for k, v in goals_by_category.items():
+            for i in range(len(v)):
+                v[i] = ObjectGoal(**v[i])
+
+                for vidx, view in enumerate(v[i].view_points):
+                    view_location = ObjectViewLocation(**view)
+                    view_location.agent_state = AgentState(
+                        **view_location.agent_state
+                    )
+                    v[i].view_points[vidx] = view_location
+
         for episode in deserialized["episodes"]:
             episode = NavigationEpisode(**episode)
 
@@ -74,15 +106,7 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
 
                 episode.scene_id = os.path.join(scenes_dir, episode.scene_id)
 
-            for i in range(len(episode.goals)):
-                episode.goals[i] = ObjectGoal(**episode.goals[i])
-
-                for vidx, view in enumerate(episode.goals[i].view_points):
-                    view_location = ObjectViewLocation(**view)
-                    view_location.agent_state = AgentState(
-                        **view_location.agent_state
-                    )
-                    episode.goals[i].view_points[vidx] = view_location
+            episode.goals = goals_by_category[episode.goals]
 
             if episode.shortest_paths is not None:
                 for path in episode.shortest_paths:
--- a/habitat/datasets/object_nav/object_nav_dataset.py
+++ b/habitat/datasets/object_nav/object_nav_dataset.py
@@ -41,14 +41,16 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
 
         goals_by_category = dict()
         for i, ep in enumerate(dataset["episodes"]):
-            goals_key = "{}_{}".format(
-                ep["scene_id"], ep["goals"][0]["object_id"]
-            )
+            dataset["episodes"][i]["object_category"] = ep["goals"][0][
+                "object_category"
+            ]
+            ep = ObjectGoalNavEpisode(**ep)
+
+            goals_key = ep.goals_key
             if goals_key not in goals_by_category:
-                goals_by_category[goals_key] = ep["goals"]
+                goals_by_category[goals_key] = ep.goals
 
             dataset["episodes"][i]["goals"] = []
-            dataset["episodes"][i]["goals_key"] = goals_key
 
         dataset["goals_by_category"] = goals_by_category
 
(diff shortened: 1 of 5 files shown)


habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -30,7 +30,26 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
     content_scenes_path: str = "{data_path}/content/{scene}.json.gz"
 
     def to_json(self) -> str:
+        for ep in self.episodes:
+            ep.goals = str(ep.goals[0].object_id)
+
+        self.goals_per_category = {}
+        for i, ep in enumerate(self.episodes):
+            obj_id = str(ep.goals[0].object_id)
```
mathfac: `ep.goals[0]` should vail because of line 34.

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -30,7 +30,26 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
     content_scenes_path: str = "{data_path}/content/{scene}.json.gz"
 
     def to_json(self) -> str:
+        for ep in self.episodes:
+            ep.goals = str(ep.goals[0].object_id)
```
mathfac: Maybe better key can be `{ep.scene}_{ep.goals[0].category}`, then dataset with multiple scenes won't collide.

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -63,6 +79,25 @@ def from_json(
             self.category_to_scene_annotation_category_id.keys()
         ), "category_to_task and category_to_mp3d must have the same keys"
 
+        if not "goals_by_category" in deserialized:
+            if len(deserialized["episodes"]) == 0:
+                return
+            else:
```
mathfac: I was expecting the code will create `goals_by_category` if it's not provided and make it new format.

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -74,15 +109,7 @@ def from_json(
...
-            for i in range(len(episode.goals)):
-                episode.goals[i] = ObjectGoal(**episode.goals[i])
-
-                for vidx, view in enumerate(episode.goals[i].view_points):
-                    view_location = ObjectViewLocation(**view)
-                    view_location.agent_state = AgentState(
-                        **view_location.agent_state
-                    )
-                    episode.goals[i].view_points[vidx] = view_location
+            episode.goals = goals_by_category[episode.goals]
```
mathfac: Maybe avoid using `goals` field and re-create key to avoid confusions:
```suggestion
            episode.goals = goals_by_category[goals_key(episode)]
```

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -63,6 +79,25 @@ def from_json(
             self.category_to_scene_annotation_category_id.keys()
         ), "category_to_task and category_to_mp3d must have the same keys"
 
+        if not "goals_by_category" in deserialized:
+            if len(deserialized["episodes"]) == 0:
+                return
+            else:
+                raise RuntimeError("Episodes have no goals")
+
+        goals_by_category = deserialized["goals_by_category"]
```
mathfac: Maybe:
```suggestion
        self.goals_by_category = deserialized["goals_by_category"]
```

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -30,7 +30,23 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
     content_scenes_path: str = "{data_path}/content/{scene}.json.gz"
 
     def to_json(self) -> str:
+        self.goals_per_category = {}
```
mathfac: If we will have this code in `from_json`, then `self.goals_per_category` will be save here automatically.

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -30,7 +30,23 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
...
+        for i, ep in enumerate(self.episodes):
+            goals_id = "{}_{}".format(ep.scene_id, ep.goals[0].object_id)
+            if goals_id not in self.goals_per_category:
+                self.goals_per_category[goals_id] = ep.goals
+
+            self.episodes[i].goals = goals_id
```
mathfac: Key can be recreated from episode data:
```suggestion
            self.episodes[i].goals = []
```

habitat/tasks/nav/object_nav_task.py
```diff
@@ -23,6 +23,15 @@
 )
 
 
+@attr.s(auto_attribs=True, kw_only=True)
+class ObjectGoalNavEpisode(NavigationEpisode):
+    r"""ObjectGoal Navigation Episode
+
+    :param goals_key: Key to retrieve goals with
+    """
+    goals_key: str = None
```
mathfac: Optional?

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -26,13 +30,58 @@ class ObjectNavDatasetV1(PointNavDatasetV1):
     """
     category_to_task_category_id: Dict[str, int]
     category_to_scene_annotation_category_id: Dict[str, int]
-    episodes: List[NavigationEpisode]
+    episodes: List[ObjectGoalNavEpisode]
     content_scenes_path: str = "{data_path}/content/{scene}.json.gz"
+    goals_by_category: Dict[str, List[ObjectGoal]]
+
+    @staticmethod
+    def dedup_goals(dset: Dict[str, Any]) -> Dict[str, Any]:
```
mathfac: you're doing change in place on `dset`, no need to return it.

mathfac:
```suggestion
    def dedup_goals(dataset: Dict[str, Any]) -> Dict[str, Any]:
```

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -74,27 +133,17 @@ def from_json(
...
 
             if episode.shortest_paths is not None:
                 for path in episode.shortest_paths:
                     for p_index, point in enumerate(path):
-                        point = {
-                            "action": point,
-                            "rotation": None,
-                            "position": None,
-                        }
+                        if isinstance(point, int) or point is None:
```
mathfac: We switching to action be str.
```suggestion
                        if isinstance(point, int) or point is None or point is str:
```

habitat/tasks/nav/object_nav_task.py
mathfac: We need this key only once. I would use function/labda instead and avoid creating it.

erikwijmans: You need it all the time.  We need to save out the `goals_key` to disk.  The de-duplicated dataset should be saved to disk for the best performance benefits (for reference, running de-dup takes 30 minutes, so it really should be done just once).

habitat/datasets/object_nav/object_nav_dataset.py
erikwijmans: The change in-place is done on purpose.  It makes a considerable difference on the time it takes to construct the list of episodes if you do this change in-place first.

habitat/datasets/object_nav/object_nav_dataset.py
```diff
@@ -63,8 +114,18 @@ def from_json(
             self.category_to_scene_annotation_category_id.keys()
         ), "category_to_task and category_to_mp3d must have the same keys"
 
-        for episode in deserialized["episodes"]:
-            episode = NavigationEpisode(**episode)
+        if len(deserialized["episodes"]) == 0:
```
mathfac: Do we really need to drop stop it here? I can imagine that some data that could left after deleting episodes won't be read.

mathfac:
Looks good to go. One last piece worth to check how it will work with `CONTENT_SCENES`.

habitat/datasets/object_nav/object_nav_dataset.py
erikwijmans: If there aren't any episodes, there aren't any goals, so stopping here makes sense.

PR #310, opened 2020-02-21
Split on val

erikwijmans:
## Motivation and Context

We currently have logic in habitat-baselines that relies on each sub-process having unique episodes during evaluation, however, or current scene splitting logic does **not** work for evaluation, thus evaluation will sometimes not run on all the scenes!

Also added a progress bar, they are nice and helps verify that this is working as intended.

## How Has This Been Tested

Tests for splitting logic.  

Evaluated a random checkpoint and it gets through all 994 Gibson pointnav eval episodes, every time. Without this fix, it consistently doesn't.

## Types of changes


habitat/core/dataset.py
```diff
@@ -65,6 +67,25 @@ class Dataset(Generic[T]):
...
 
+    @staticmethod
+    def _scene_from_episode(episode: T) -> str:
+        r"""Helper method to get the scene name from an episode.  Assumes
+        the scene_id is formated /path/to/<scene_name>.<ext>
+        """
+        return os.path.splitext(os.path.basename(episode.scene_id))[0]
+
+    @classmethod
+    def get_scenes_to_load(cls, config: Config) -> List[str]:
```
mathfac: Duplicating `scene_ids ` property: map(dataset.scene_ids, Dataset. _scene_from_episode). We should have one method for that maybe with param or combine 2 methods like above.

habitat/datasets/eqa/mp3d_eqa_dataset.py
```diff
@@ -57,6 +58,14 @@ def __init__(self, config: Config = None) -> None:
         with gzip.open(config.DATA_PATH.format(split=config.SPLIT), "rt") as f:
             self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
 
+        if ALL_SCENES_MASK not in config.CONTENT_SCENES:
+            scenes_to_load = set(config.CONTENT_SCENES)
+            self.episodes = [
```
mathfac: As I understand this part of code filters episode form scenes that were not specified, but loaded in main dataset file. Maybe, use `dataset.filter` here.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -51,10 +51,25 @@ def get_scenes_to_load(cls, config: Config) -> List[str]:
...
+        ):
+            return cls._get_scenes_from_folder(
+                content_scenes_path=dataset.content_scenes_path,
+                dataset_dir=dataset_dir,
+            )
+        else:
+            # Load the full dataset, things are split into separate files
+            cfg.CONTENT_SCENES = [ALL_SCENES_MASK]
+            dataset = cls(cfg)
+            scenes = {
```
mathfac: `dataset.scene_ids` maybe?

habitat/datasets/vln/r2r_vln_dataset.py
```diff
@@ -58,6 +45,14 @@ def __init__(self, config: Optional[Config] = None) -> None:
         with gzip.open(dataset_filename, "rt") as f:
             self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
 
+        if ALL_SCENES_MASK not in config.CONTENT_SCENES:
+            scenes_to_load = set(config.CONTENT_SCENES)
+            self.episodes = [
```
mathfac: `dataset.filter` here.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -84,19 +99,32 @@ def __init__(self, config: Optional[Config] = None) -> None:
...
-        for scene in scenes:
-            scene_filename = self.content_scenes_path.format(
-                data_path=dataset_dir, scene=scene
+        if os.path.exists(
+            self.content_scenes_path.split("{scene}")[0].format(
+                data_path=dataset_dir
             )
-            with gzip.open(scene_filename, "rt") as f:
-                self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
+        ):
```
mathfac: criptic.

test/test_pointnav_dataset.py
```diff
@@ -109,6 +109,54 @@ def test_multiple_files_pointnav_dataset():
     check_json_serializaiton(partial_dataset)
 
 
+@pytest.mark.parametrize("split", ["train", "val"])
+def test_dataset_splitting(split):
```
mathfac: If that makes sense to have one test for dataset configs as params. Then we can make sure new tasks will be just added.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -84,19 +99,32 @@ def __init__(self, config: Optional[Config] = None) -> None:
...
+            for scene in scenes:
+                scene_filename = self.content_scenes_path.format(
+                    data_path=dataset_dir, scene=scene
+                )
+                with gzip.open(scene_filename, "rt") as f:
+                    self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
+
+        elif ALL_SCENES_MASK not in config.CONTENT_SCENES:
+            scenes_to_load = set(config.CONTENT_SCENES)
+            self.episodes = [
```
mathfac: `dataset.filter`

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -84,19 +99,32 @@ def __init__(self, config: Optional[Config] = None) -> None:
...
+                )
+
+            for scene in scenes:
+                scene_filename = self.content_scenes_path.format(
+                    data_path=dataset_dir, scene=scene
+                )
+                with gzip.open(scene_filename, "rt") as f:
+                    self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
+
+        elif ALL_SCENES_MASK not in config.CONTENT_SCENES:
```
mathfac: without `elif`, to be consistent we should filter episodes from `non-allowed` scenes anyway.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -84,19 +99,32 @@ def __init__(self, config: Optional[Config] = None) -> None:
...
-            scene_filename = self.content_scenes_path.format(
-                data_path=dataset_dir, scene=scene
+        if os.path.exists(
+            self.content_scenes_path.split("{scene}")[0].format(
+                data_path=dataset_dir
             )
-            with gzip.open(scene_filename, "rt") as f:
-                self.from_json(f.read(), scenes_dir=config.SCENES_DIR)
+        ):
+            scenes = config.CONTENT_SCENES
```
mathfac: This part of code became hard to understand. Several overloaded terms like `scenes to load` / `content_scenes` and etc. As we are adding filtering, maybe that worth to change `CONTENT_SCENES` to something else, like `SCENES_TO_LOAD` or `ALLOWED_SCENES`. Up to you, not sure how large is refactoring.

Let's add some additional variables to make the code tell a story:
```
separate_scene_files_exist =  os.path.exists(self.content_scenes_path.split("{scene}")[0].format(
                data_path=dataset_dir
            )
if separate_scene_files_exist:
...
```

habitat/core/dataset.py
erikwijmans: The `.scene_ids` property is a little different, it returns the full scene paths, not just the scene name.  We could change that property, but I wasn't sure what the implications of doing that was.

test/test_pointnav_dataset.py
erikwijmans: Different datasets will have different splits available and will be different sizes.  For PointNav and ObjectNav, it only makes sense to load a few scenes and split on those, while for EQA and VLN, it is feasible to load all of train and split that, so that singular test will likely end up having way too many flags and conditions and people will have a harder time adding to it.

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -51,10 +50,26 @@ def get_scenes_to_load(cls, config: Config) -> List[str]:
...
+                data_path=dataset_dir
+            )
         )
+        if has_individual_scene_files:
+            return cls._get_scenes_from_folder(
+                content_scenes_path=dataset.content_scenes_path,
+                dataset_dir=dataset_dir,
+            )
+        else:
+            # Load the full dataset, things are split into separate files
```
mathfac: Looks like this branch is true when there is a single file per split.

test/test_pointnav_dataset.py
mathfac: Got it, thank you for the clarification.

habitat/core/dataset.py
```diff
@@ -65,6 +69,37 @@ class Dataset(Generic[T]):
...
+        return sorted(list(scenes))
+
+    @classmethod
+    def build_content_scenes_filter(cls, config) -> Callable[[T], bool]:
+        scenes_to_load = set(config.CONTENT_SCENES)
+
+        def _filter(ep: T) -> bool:
+            return (
+                ALL_SCENES_MASK in scenes_to_load
+                or cls._scene_from_episode(ep) in scenes_to_load
```
mathfac:
```suggestion
                or cls._scene_from_scene_path(ep.scene_id) in scenes_to_load
```

habitat/datasets/pointnav/pointnav_dataset.py
```diff
@@ -51,10 +50,26 @@ def get_scenes_to_load(cls, config: Config) -> List[str]:
...
+            cfg.CONTENT_SCENES = [ALL_SCENES_MASK]
+            dataset = cls(cfg)
+            scenes = {
```
mathfac: Then we can do:
```suggestion
            scenes = list(map(dataset.scene_ids, cls._scene_from_scene_path))
```

habitat/core/dataset.py
```diff
@@ -65,6 +69,37 @@ class Dataset(Generic[T]):
...
+        assert cls.check_config_paths_exist(config)
+        dataset = cls(config)
+        scenes = {
+            cls._scene_from_episode(episode) for episode in dataset.episodes
+        }
+
+        return sorted(list(scenes))
+
+    @classmethod
+    def build_content_scenes_filter(cls, config) -> Callable[[T], bool]:
```
mathfac: Doc string required for the core class.

habitat/core/dataset.py
mathfac: More detailed doc string required for the core class.

habitat/core/dataset.py
```diff
@@ -65,6 +69,37 @@ class Dataset(Generic[T]):
     """
     episodes: List[T]
 
+    @staticmethod
+    def _scene_from_episode(episode: T) -> str:
```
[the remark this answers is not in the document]
mathfac: Let's make it public method as the method is useful even outside of class.

habitat/core/dataset.py
```diff
@@ -65,6 +69,37 @@ class Dataset(Generic[T]):
...
+        the scene_id is formated /path/to/<scene_name>.<ext>
+        """
+        return os.path.splitext(os.path.basename(episode.scene_id))[0]
+
+    @classmethod
+    def get_scenes_to_load(cls, config: Config) -> List[str]:
+        r"""Return a sorted list of scenes
+        """
+        assert cls.check_config_paths_exist(config)
+        dataset = cls(config)
```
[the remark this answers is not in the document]
erikwijmans: Yeah, it's expensive (well as expensive as loading the dataset, which may be fairly cheap!), but necessary for correct evaluation.

PR #336, opened 2020-03-17
DD-PPO stats devices

erikwijmans:
## Motivation and Context

A little bug from #332.  It seems like the 1-process dummy distributed training on the CI doesn't enforce the same requirements on tensors that actually distributed training does.  A note for the future: The CI won't catch everything for distributed training nor with testing with just 1 process!

## How Has This Been Tested

With 8 processes :-)

## Types of changes


erikwijmans:
We'd need to have multiple GPUs on the CI node to do proper testing of this I think :(

PR #352, opened 2020-03-31
Cache the path object for faster pathfinding

erikwijmans:
## Motivation and Context

Make geodesic distance computation faster by caching the habitat-sim `path` object on the episode.  Caching this path object gives two improvements, 1) no more need to convert List[List[float]] to a 2D numpy array all the time (which is slow!) and 2) goes with https://github.com/facebookresearch/habitat-sim/pull/555 to enable caching of the shortest path finding setup, which can end up being the bottleneck in a lot of cases.

## How Has This Been Tested

Existing tests still pass.

## Types of changes


dhruvbatra:
CC: @shivanshpatel35 @msavva

erikwijmans:
Added a check to see if we actually need to recompute the distance to goals, this gives another nice speedup.

erikwijmans:
I am seeing ~100FPS on my test setup (up from the 25 FPS earlier today) using the latest iteration of this + the latest iteration of https://github.com/facebookresearch/habitat-sim/pull/555

habitat/core/dataset.py
```diff
@@ -59,6 +59,18 @@ class Episode:
         default=None, validator=not_none_validator
     )
     info: Optional[Dict[str, str]] = None
+    _shortest_path_cache: Any = attr.ib(init=False, default=None)
```
mathfac, line 62: As I understand we are going to hide `shortest_path_cache` in Episode for ease of use. While it won't be visible during serialization.

habitat/tasks/nav/nav.py
```diff
@@ -863,29 +863,34 @@ def _euclidean_distance(self, position_a, position_b):
...
-        )
+            self._agent_episode_distance += self._euclidean_distance(
+                current_position, self._previous_position
+            )
 
-        self._previous_position = current_position
+            self._previous_position = current_position
 
-        self._metric = distance_to_target
+            self._metric = distance_to_target
```
mathfac, line 893: How `distance_to_target` is calculated when cache is None?

habitat/core/dataset.py
erikwijmans: Yep.  Putting it on the episode seemed the cleanest, but we can't serialize the cache.

habitat/tasks/nav/nav.py
erikwijmans: Made the clearer.  It also looks like `self._agent_episode_distance` is a hold-over from SPL?

PR #361, opened 2020-04-03
SoftSPL

abhiskk:
## Motivation and Context

SoftSPL: SPL with a softer success criterion

## How Has This Been Tested

The SoftSPL metric has been added to the test-config.


habitat/tasks/nav/nav.py
```diff
@@ -430,7 +430,7 @@ def reset_metric(self, *args: Any, episode, task, **kwargs: Any):
         task.measurements.check_measure_dependencies(
             self.uuid, [DistanceToGoal.cls_uuid]
         )
-        self.update_metric(*args, episode=episode, task=task, **kwargs)
+        self.update_metric(episode=episode, task=task, *args, **kwargs)
```
abhiskk, line 433: THis modification makes the positioning of `args, kwargs` in function signature consistent with the common python conventions.

habitat/tasks/nav/nav.py
```diff
@@ -509,6 +509,41 @@ def update_metric(
...
+    Similar to SPL with a relaxed soft-success criteria. Instead of a boolean
+    success is now calculated as 1 - (ratio of distance covered to target).
+    """
+
+    def _get_uuid(self, *args: Any, **kwargs: Any):
+        return "softspl"
+
+    def update_metric(self, episode, *args: Any, **kwargs: Any):
+        current_position = self._sim.get_agent_state().position.tolist()
+        distance_to_target = self._sim.geodesic_distance(
```
mathfac, line 525: @abhiskk we should use `DISTANCE_TO_GOAL` measure here as we do in SPL.

mathfac: `distance_to_target` became more general and works for multiple goals and view points. Here is the example from SPL metric:
https://github.com/facebookresearch/habitat-api/blob/6091d0aedd41d348824252d74119d5d0d3355b8e/habitat/tasks/nav/nav.py#L478-L510

habitat/tasks/nav/nav.py
```diff
@@ -509,6 +509,41 @@ def update_metric(
...
+            current_position, episode.goals[0].position
+        )
+
+        ep_soft_success = (
```
erikwijmans, line 529: Should this be `max(1 - distance_to_target / self._start_end_episode_distance, 0)`, that way the metric is in `[0, 1]`?

habitat/tasks/nav/nav.py
```diff
@@ -509,6 +509,41 @@ def update_metric(
...
 
+@registry.register_measure
+class SoftSPL(SPL):
+    r"""Soft SPL
+
+    Similar to SPL with a relaxed soft-success criteria. Instead of a boolean
+    success is now calculated as 1 - (ratio of distance covered to target).
+    """
+
+    def _get_uuid(self, *args: Any, **kwargs: Any):
```
mathfac, line 520: @abhiskk, looks like `reset` method was lost. Worth to add minimal valuable test so we could caught changes like this.

mathfac: it should require dependency metric:
```
     task.measurements.check_measure_dependencies( 
         self.uuid, [DistanceToGoal.cls_uuid] 
     ) 
```

erikwijmans: I think since it inherits from SPL (and thus gets its reset method), the right thing will already happen?

abhiskk: Yes, because it inherits reset functionality will remain the same.

mathfac: @abhiskk, SPL depends on Success metrics, while SoftSPL isn't.

PR #365, opened 2020-04-07
Added center cropping and resize ops for PPO agents

Skylion007:
## Motivation and Context

Allows people to train PPO agents with non-square screen resolution. Add config options to automatically resize and crop the image in a performant manner. 

Feedback requested on how to best implement this API and configs.

## How Has This Been Tested

It has been tested by testing the utility functions on various local agents.

## Types of changes

## Checklist

### DIFF
--- a/habitat_baselines/common/utils.py
+++ b/habitat_baselines/common/utils.py
@@ -208,13 +208,9 @@ def image_resize_shortest_edge(img, size: int, channels_first: bool = False):
     w *= percent
     h = int(h)
     w = int(w)
-    is_uint8 = img.dtype == torch.uint8
-    if is_uint8:
-        # Interpolate isn't implemented for UINT8
-        img = img / 255.0
-    img = torch.nn.functional.interpolate(img, size=(h, w), mode="area")
-    if is_uint8:
-        img = (255.0 * img).round().byte()
+    img = torch.nn.functional.interpolate(
+        img.float(), size=(h, w), mode="area"
+    ).to(dtype=img.dtype)
     if not channels_first:
         img = img.permute(0, 2, 3, 1)
     if no_batch_dim:
@@ -222,19 +218,29 @@ def image_resize_shortest_edge(img, size: int, channels_first: bool = False):
     return img
 
 
-def center_crop(img, cropx: int, cropy: int):
-    """performs a center
+def center_crop(img, cropx: int, cropy: int, channels_first: bool = False):
+    """Performs a center crop on an image.
 
     Args:
-        img: the array object that needs to be resized
+        img: the array object that needs to be resized (either batched or unbatched)
         size: the size that you want the shortest edge to be resize to
+        channels_first: If it's in NCHW
     Returns:
         the resized array
     """
-    y, x = img.shape[:2]
+    if channels_first:
+        # NCHW
+        y, x = img.shape[-2:]
+    else:
+        # NHWC
+        y, x = img.shape[-3:-1]
+
     startx = x // 2 - (cropx // 2)
     starty = y // 2 - (cropy // 2)
-    return img[starty : starty + cropy, startx : startx + cropx]
+    if channels_first:
+        return img[..., starty : starty + cropy, startx : startx + cropx]
+    else:
+        return img[..., starty : starty + cropy, startx : startx + cropx, :]
 
 
 def apply_ppo_data_augs(
--- a/habitat_baselines/common/utils.py
+++ b/habitat_baselines/common/utils.py
@@ -13,11 +13,8 @@ import numpy as np
 import torch
 import torch.nn as nn
 
-from habitat.core.utils import try_cv2_import
 from habitat.utils.visualizations.utils import images_to_video
 from habitat_baselines.common.tensorboard_utils import TensorboardWriter
-
-cv2 = try_cv2_import()
 
 
 class Flatten(nn.Module):
@@ -179,16 +176,30 @@ def generate_video(
         )
 
 
-def image_resize_shortest_edge(img: np.ndarray, size: int):
+def image_resize_shortest_edge(img, size: int, channels_first: bool = False):
     """Resizes an img so that the shortest side is length of size.
 
     Args:
-        img: the array object that needs to be resized
+        img: the array object that needs to be resized (HWC) or (NHWC)
         size: the size that you want the shortest edge to be resize to
+        channels_first: a boolean that specifies the img is (CHW) or (NCHW)
     Returns:
-        the resized array
+        The resized array as a torch tensor.
     """
-    h, w = img.shape[:2]
+    no_batch_dim = len(img.shape) == 3
+    if len(img.shape) != 3 and len(img.shape) != 4:
+        raise NotImplementedError()
+    img = _to_tensor(img)
+    if no_batch_dim:
+        img = img.unsqueeze(0)  # Adds a batch dimension
+    if channels_first:
+        # NCHW
+        h, w = img.shape[-2:]
+    else:
+        # NHWC
+        h, w = img.shape[1:3]
+        img = img.permute(0, 3, 1, 2)
+
     if w > h:
         percent = size / h
     else:
@@ -197,9 +208,17 @@ def image_resize_shortest_edge(img: np.ndarray, size: int):
     w *= percent
     h = int(h)
     w = int(w)
-    img = cv2.resize(img, (h, w), interpolation=cv2.INTER_AREA)
-    if len(img.shape) == 2:
-        img = np.expand_dims(img, axis=-1)
+    is_uint8 = img.dtype == torch.uint8
+    if is_uint8:
+        # Interpolate isn't implemented for UINT8
+        img = img / 255.0
+    img = torch.nn.functional.interpolate(img, size=(h, w), mode="area")
+    if is_uint8:
+        img = (255.0 * img).round().byte()
+    if not channels_first:
+        img = img.permute(0, 2, 3, 1)
+    if no_batch_dim:
+        img = img.squeeze(dim=0)  # Removes the batch dimension
     return img
 
 
--- a/habitat_baselines/common/utils.py
+++ b/habitat_baselines/common/utils.py
@@ -6,6 +6,7 @@
 
 import copy
 import glob
+import numbers
 import os
 from collections import defaultdict
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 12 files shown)


Skylion007:
There is one test that is currently failing, but it seems to be caused by a YAML config change not related to this PR? Should we bring the AGENT_PPO agent_config inline with the rest of Habitat?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +177,55 @@ def generate_video(
...
+    h, w = img.shape[:2]
+    if w > h:
+        percent = size / h
+    else:
+        percent = size / w
+    h *= percent
+    w *= percent
+    h = int(h)
+    w = int(w)
+    img = cv2.resize(img, (h, w), interpolation=cv2.INTER_AREA)
```
erikwijmans, line 200: Why not do this on the GPU with https://pytorch.org/docs/stable/nn.functional.html?highlight=interpolate#torch.nn.functional.interpolate ?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +174,77 @@ def generate_video(
...
+        percent = size / w
+    h *= percent
+    w *= percent
+    h = int(h)
+    w = int(w)
+    is_uint8 = img.dtype == torch.uint8
+    if is_uint8:
+        # Interpolate isn't implemented for UINT8
+        img = img / 255.0
+    img = torch.nn.functional.interpolate(img, size=(h, w), mode="area")
```
erikwijmans, line 215:
```suggestion
    img = torch.nn.functional.interpolate(img.float(), size=(h, w), mode="area").to(dtype=img.dtype)
```

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -44,6 +45,24 @@ def __init__(self, config=None):
         self.envs = None
         if config is not None:
```
erikwijmans, line 46: This will need to be `self.config` or be done before the call to `super().__init__(config)`

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -56,8 +67,21 @@ def __init__(
         spatial_size=128,
         make_backbone=None,
         normalize_visual_inputs=False,
+        force_input_size=(256, 256),
     ):
         super().__init__()
+        self.force_input_size = force_input_size
+        observation_space = copy.deepcopy(observation_space)
```
erikwijmans, line 74: Could we make a base class that takes `observation_space` and `force_input_size` and produces `self.observation_space` (then both SimpleCNN and ResNet can share this logic).  That base class could also have a `.forward` method that does the resize.

This could be done in a latter PR of course (there is more unification that could be done than just this).

Skylion007: Done

habitat_baselines/config/default.py
```diff
@@ -144,7 +144,10 @@ def get_config(
 
         for config_path in config_paths:
             config.merge_from_file(config_path)
-
+    if opts:
+        for k, v in zip(opts[0::2], opts[1::2]):
```
mathfac, line 148: What is the logic behind this change?

habitat_baselines/common/utils.py
```diff
@@ -53,6 +56,51 @@ def forward(self, x):
         return CustomFixedCategorical(logits=x)
 
 
+class ResizeCenterCropper(nn.Module):
+    def __init__(
+        self, force_input_size: Optional[tuple], channels_first: bool = True
```
mathfac, line 61: `force_input_size` name is a little bit confusing. Is that input size for transformation or size of input to the next level. Maybe, use naming similar to [torchvision.transforms.CenterCrop]:(https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.CenterCrop)
`size (sequence or python:int) – Desired output size of the crop. If size is an int instead of sequence like (h, w), a square crop (size, size) is made.`

Regarding `channels_first` maybe call it like `NCHW` and add docstring. Because, channels are not really first and to be more explicit.

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+    if no_batch_dim:
+        img = img.unsqueeze(0)  # Adds a batch dimension
+    if channels_first:
+        # NCHW
+        h, w = img.shape[-2:]
+    else:
+        # NHWC
+        h, w = img.shape[-3:-1]
+        if len(img.shape) == 4:
+            img = img.permute(0, 3, 1, 2)
```
mathfac, line 250: Why do we need to do this permutations?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+        else:
+            img = img.permute(0, 1, 4, 2, 3)
+
+    if w > h:
+        percent = size / h
```
mathfac, line 255: what is `percent` here?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+    Args:
+        img: the array object that needs to be resized (either batched or unbatched)
+        size: the size that you want the shortest edge to be resize to
+        channels_first: If it's in NCHW
+    Returns:
+        the resized array
+    """
+    if channels_first:
+        # NCHW
+        y, x = img.shape[-2:]
```
mathfac, line 287:
```suggestion
        h, w = img.shape[-2:]
```

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+    else:
+        # NHWC
+        y, x = img.shape[-3:-1]
```
mathfac, line 290: Maybe call it h, w:
```suggestion
        h, w = img.shape[-3:-1]
```

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+            img = img.permute(0, 2, 3, 1)
+        else:
+            img = img.permute(0, 1, 3, 4, 2)
+    if no_batch_dim:
+        img = img.squeeze(dim=0)  # Removes the batch dimension
+    return img
+
+
+def center_crop(img, cropx: int, cropy: int, channels_first: bool = False):
+    """Performs a center crop on an image.
```
mathfac, line 276: Should it be `crop_size` not x, y?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+
+    startx = x // 2 - (cropx // 2)
+    starty = y // 2 - (cropy // 2)
+    if channels_first:
+        return img[..., starty : starty + cropy, startx : startx + cropx]
+    else:
+        return img[..., starty : starty + cropy, startx : startx + cropx, :]
+
+
+def apply_ppo_data_augs(
```
mathfac, line 300: This function isn't used.

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+                    obs,
+                    center_crop_size,
+                    center_crop_size,
+                    channels_first=channels_first,
+                )
+            observations[k] = obs
+    return observations
+
+
+def overwrite_gym_box(box: Box, shape: tuple) -> Box:
```
mathfac, line 323: Maybe `overwrite_gym_box_height_width`.

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+    # if len(shape) < len(box.shape):
```
mathfac, line 324: Do we still need this comments?

habitat_baselines/common/utils.py
```diff
@@ -174,3 +222,111 @@ def generate_video(
...
+    shape = list(shape) + list(box.shape[len(shape) :])
+    low = box.low
```
mathfac, line 326: Maybe shorter form:
`low = box.low if np.isscalar(low) else np.min(low)`

habitat_baselines/common/utils.py
```diff
@@ -53,6 +56,51 @@ def forward(self, x):
...
+                    )
+        self.observation_space = observation_space
+        return observation_space
+
+    def forward(self, input: torch.Tensor):
+        if self.force_input_size is None:
+            return input
+
+        return center_crop(
+            image_resize_shortest_edge(
```
mathfac, line 93: How to disable `image_resize_shortest_edge` functionality in current setup?

test/test_baseline_agents.py
```diff
@@ -25,28 +26,39 @@
...
-        if input_type in ["depth", "rgbd"]:
-            config_env.SIMULATOR.AGENT_0.SENSORS += ["DEPTH_SENSOR"]
-        config_env.freeze()
-        del benchmark._env
-        benchmark._env = habitat.Env(config=config_env)
-        agent_config.INPUT_TYPE = input_type
-
-        agent = ppo_agents.PPOAgent(agent_config)
-        habitat.logger.info(benchmark.evaluate(agent, num_episodes=10))
+        for resolution in [256, 384]:
```
mathfac, line 40: Use `resolution` for `@pytest.mark.parametrize` like [here](https://github.com/facebookresearch/habitat-api/blob/master/test/test_baseline_trainers.py#L29).

habitat_baselines/common/utils.py
Skylion007: Because PyTorch only accepts NCHW channel order for that function.

habitat_baselines/config/default.py
Skylion007: It's impossible to overwrite the BASE_TASK_CONFIG from the command line without since the BASE_TASK_CONFIG is used before its args are overwritten by the command line. Likewise, moving the code for that to this point would make it impossible to overwrite TASK_CONFIG variables from the command line. As such, BASE_TASK_CONFIG must be extracted and overwritten and then the remaining config parameters can be overwritten.

habitat_baselines/common/utils.py
Skylion007: It also resizes the input the though, it doesn't just center crop. I also kept it like so because at least for center cropping I wrote the code such that ....CHW or ...HWC (regardless of the number of channels) were supported. The channels_first is a convention from Tensorflow/Keras.

habitat_baselines/common/utils.py
Skylion007: its the percentage resize to maintain the proper aspect ratio.

habitat_baselines/common/utils.py
Skylion007: It can also overwrite channels if specified though or any other aspects of shape.

habitat_baselines/common/utils.py
Skylion007: There isn't a way, that's why it's ResizeCenterCropper not just CenterCropper. Currently there is no way to disable it.

habitat_baselines/common/utils.py
Skylion007: It only crops along height, width.

test/test_baseline_agents.py
Skylion007: That would require reconstructing the agent and benchmark for every iteration? I think there is a reason this is already done in a loop before I added the code (it would make the test a lot longer).

habitat_baselines/common/utils.py
Skylion007: I changed to channels_list since that is true regardless of the channels number.

habitat_baselines/common/utils.py
```diff
@@ -174,3 +226,92 @@ def generate_video(
         tb_writer.add_video_from_np_images(
             f"episode{episode_id}", checkpoint_idx, images, fps=fps
         )
+
+
+def image_resize_shortest_edge(img, size: int, channels_last: bool = False):
```
mathfac, line 231: Maybe:
```suggestion
def image_resize_shortest_edge(img: np.array, size: int, channels_last: bool = False):
```

mathfac, line 231: Please, specify type for img.
```suggestion
def image_resize_shortest_edge(img: torch.Tensor, size: int, channels_last: bool = False):
```

habitat_baselines/common/utils.py
```diff
@@ -174,3 +226,92 @@ def generate_video(
...
+        if len(img.shape) == 4:
+            img = img.permute(0, 3, 1, 2)
+        else:
+            img = img.permute(0, 1, 4, 2, 3)
+    else:
+        # NCHW
+        h, w = img.shape[-2:]
+
+    # Percentage resize
+    if w > h:
```
mathfac, line 260: maybe
`percent = size / max(h, w)`

habitat_baselines/common/utils.py
```diff
@@ -174,3 +226,92 @@ def generate_video(
...
+        percent = size / h
+    else:
+        percent = size / w
+    h *= percent
```
mathfac, line 264: Then do it in one line `h = int(h * percent)`.

habitat_baselines/common/utils.py
mathfac: @Skylion007, for each `img.permute` can we add inline comments like `# NHWC = >NCHW`. Then it will be easier to support the code.

habitat_baselines/common/utils.py
Skylion007: it's shortest resize so it would percent = size / min(h,w)

test/test_baseline_agents.py
mathfac: Makes sense, do you want to test when `h <> w`?

habitat_baselines/common/utils.py
mathfac: Yes, you're right.

Skylion007:
@mathfac it's not noticeably affected and it shouldn't be as all those operations are performed on the GPU. (Cropping is a straight up array view and resizing uses an optimized GPU kernel to compute it.

mathfac:
@Skylion007, looks like habitat baseline config part is missing to configure ResizeCrop.

Skylion007:
> @Skylion007, looks like habitat baseline config part is missing to configure ResizeCrop.

We don't even have a good way to deserialize arbitrary transforms for this yet and walking the config down to that scope would require significant code changes at the moment so I plan on leaving it hard coded for now.

mathfac:
Tested it for ObjectNav:
For 415232 frames:
With resize Depth only 640x480: env-time: 2513.411s	pth-time: 3503.996s
Without resize  Depth only 1x1: env-time: 1744.576s	pth-time: 3602.428s
**+44% env-time**. I think it acceptable with resolution change. Merging.

Issue #369, opened 2020-04-14
EmbodiedQA baseline in Habitat


mukulkhanna:
## ❓ Questions and Help

Hello everyone,

I have been working on implementing a primarily IL-based EQA baseline in Habitat with guidance from @dhruvbatra.

I have chosen to implement an EQA baseline using the approach shared in EQA's first paper (CVPR 2018; [link](https://embodiedqa.org/paper.pdf)) and have used MP3D scene and task datasets for the same.

The proposed approach includes 4 stages of training - 

1. Training a feature extractor (#430)
2. Training the VQA model (#487)
3. Training the NAVigation model (#539)
4. Training for EQA by finetuning the NAV model

I have been able to complete the first 3 stages. The forked repo can be found [here](https://github.com/mukulkhanna/habitat-api/tree/eqa-1) (eqa-1 branch). More information about the implementation can be found in the [README](https://github.com/mukulkhanna/habitat-api/blob/eqa-1/habitat_baselines/il/README.md).

Due to the infeasibility of loading each scene for each episode during training, one of the key design decisions was to cache each episode's images beforehand by storing them on the disk. Even though this might not seem to be the most elegant approach, loading each scene for each episode will severly tank training speed. Also, since all of the first 3 trainers are going to use this data, caching by writing to the disk seemed to be the best option.

I would like the community to share their feedback about the whole implementation. Since there are multiple levels, we can start of with the first two (feature extractor and vqa). Please let me know if I can share any more information about the aforementioned.

Thank you.

mathfac:
Hi @mukulkhanna,
That looks great, please, send it as draft PR that we can give more grounded to code feedback.
Regarding loading data to the disk, that may makes sense for IL. Usually we load one house per worker and switch scenes only after all episodes from current scenes were passed.

mukulkhanna:
@mathfac Thank you for the response. I will open a PR for that. 

Should I close this issue or let it be open till the PR is resolved?

> Usually we load one house per worker and switch scenes only after all episodes from current scenes were passed.

I thought of using that method too, but in a single-worker IL setting like mine, I felt that would impinge on the training process. I believe it would be better if the scenes (and corresponding frames) are shuffled; that should avoid overfitting and generalise better.

rpartsey:
All the PR's mentioned in the scope of this Issue have already been merged.

PR #374, opened 2020-04-16
EmbodiedQA baseline implementation

mukulkhanna:
## Motivation and Context

This is to implement an Embodied QA baseline in Habitat using the approach shared in EQA's first paper (CVPR 2018; [link](https://embodiedqa.org/paper.pdf)), using the Matterport 3D dataset. 

### Corresponding Pull requests

- EQA-CNN encoder trainer **(merged)**: #430  
- VQA trainer **(merged)**: #487 
- NAV PACMAN trainer **(merged)** : #539

For more information about the implementation, please refer to #369 and this [README](https://github.com/mukulkhanna/habitat-api/blob/vqa/habitat_baselines/il/README.md).

## How Has This Been Tested

By running all trainers manually. Most baseline_trainer tests don't apply. Would need help in curating relevant test cases.

## Types of changes

## Checklist

Please let me know if I can share any more information about the aforementioned.

Thank you.


habitat_baselines/il/data/edfe_data.py
```diff
@@ -0,0 +1,224 @@
...
+        for idx, pos in enumerate(pos_queue):
+
+            observation = self.env.sim.get_observations_at(
+                pos.position, pos.rotation
+            )
+
+            depth = observation["depth"]
+            rgb = observation["rgb"]
+            seg = observation["semantic"]
+            seg = seg % 40
```
erikwijmans, line 99: This likely isn't correct.  The semantic sensor produces instances segmentation.  If you want semantic segmentation, you'll need to map object ID's to semantic label ID.

habitat_baselines/il/data/edfe_data.py
```diff
@@ -0,0 +1,224 @@
...
+                "ep_" + episode_id + "_{0:0=3d}".format(idx),
+            )
+            seg_frame_path = os.path.join(
+                self.dataset_path.format(split=split, type="seg"),
+                "ep_" + episode_id + "_{0:0=3d}".format(idx),
+            )
+
+            cv2.imwrite(rgb_frame_path + ".jpg", rgb)
+            cv2.imwrite(depth_frame_path + ".jpg", depth * 255)
+            np.savez_compressed(seg_frame_path, seg)
```
erikwijmans, line 122: I highly recommending putting these into a database like https://www.h5py.org/ or https://lmdb.readthedocs.io/en/release/

Reading out of databases like this is considerably faster than reading from thousands of files on disk.

mukulkhanna: Alright, I'll make the changes. Thank you

habitat_baselines/il/data/edfe_data.py
mukulkhanna: Thank you @erikwijmans for pointing this out.

With help from [this issue](https://github.com/facebookresearch/habitat-sim/issues/263), I am able to extract the semantic label ids using this -

```
scene = self.env.sim.semantic_annotations()

instance_id_to_label_id = {int(obj.id.split("_")[-1]): obj.category.index() for obj in scene.objects}
self.mapping = np.array([ instance_id_to_label_id[i] for i in range(len(instance_id_to_label_id)) ])
seg = np.take(self.mapping, observation['semantic'])
```

I see that there are a total of 42 categories (semantic label ids) across all the objects in the MP3D scene dataset-
|Category ID   | Category name   |
|:-:|:-:|
| -1  | '' |
| 0  | 'void'  |
| 1  | 'wall'  |
| ...  |  ... |
| 40  | 'misc'  |

Just wanted to confirm if it'd be okay to fuse the first two categories (ids -1 and 0) into one and specify 41 classes for the task of learning the segmentation map?

erikwijmans: Yeah, that sounds fine.

habitat_baselines/il/data/edfe_data.py
mukulkhanna: @erikwijmans 

There are a total of 360k images (100k * 3 train + 20k * 3 test) that need to be saved, before training starts. Currently, the dataset takes about 5.1 GB disk space (4.1 GB train + 1 GB test). 

Even though reading out files from an h5 database will be fast, saving the whole dataset into train.h5 and test.h5 will likely have the below problems - 
 
1. Take much more disk space (even gzip compressed ones)
2. While writing to H5 format - we can't load all the (360k) frames into memory before writing them out to the h5 file , and appending images to the h5 database, episode by episode will be quite memory expensive too.

Could you help me with what can be done? Thank you

erikwijmans: You can use PIL + LMDB to keep things compressed.  LDMB is a key-value storage and will just store arbitrary binary. I know you can use PIL to write to/read from https://docs.python.org/3/library/io.html#io.BytesIO (cv2 may also be able to do this, but you need to be careful with rgb vs. bgr).  jpeg is pretty slow to decode however, so storing uncompressed is common.

For h5, you can pre-allocate everything and just write directly into the h5 file for each image.  No need to sore everything in memory and write all at once.

habitat_baselines/il/data/edfe_data.py
```diff
@@ -52,7 +53,25 @@ def __init__(self, env, config, mode="train"):
             for each scene > load scene in memory > save frames for each
             episode corresponding to each scene
             """
-            print("Saving rgb, seg, depth frames to disk.")
+            print("Saving rgb, seg, depth data to database.")
+
+            train_lmdb = lmdb.open(
```
mukulkhanna, line 58: @erikwijmans LMDB code added. Please review, thank you.

habitat_baselines/il/data/edfe_data.py
mukulkhanna: @erikwijmans 

Thank you for the explanation. I have made the suggested changes in the latest commit-

- fix for semantic segmentation annotations
- LMDB database added

If the LMDB implementation looks good, I'll make the same changes for the VQA, NAV trainer datasets.

habitat_baselines/il/data/edfe_data.py
```diff
@@ -0,0 +1,220 @@
...
+                    self.save_frames(random_pos)
+
+            print("EDFE database ready!")
+
+            if self.mode == "train":
+                self.lmdb_txn = self.train_txn
+            elif self.mode == "val":
+                self.lmdb_txn = self.val_txn
+        else:
+            lmdb_env = lmdb.open(self.dataset_path.format(split=self.mode))
```
erikwijmans, line 97:
```suggestion
            lmdb_env = lmdb.open(self.dataset_path.format(split=self.mode), readonly=True, lock=False)
```

I have found that some systems need `lock=False` to support concurrent loading from many workers smoothly.

habitat_baselines/il/trainers/edfe_trainer.py
```diff
@@ -0,0 +1,249 @@
...
+        edfe_dataset = EDFEDataset(env, config)
+
+        train_loader = DataLoader(
+            edfe_dataset, batch_size=config.IL.EDFE.batch_size, shuffle=True
+        )
+
+        logger.info("train_loader has %d samples" % len(edfe_dataset))
+
+        model = MultitaskCNNOutput(num_classes=41)
+        model.double().train().cuda()
```
erikwijmans, line 58: Just saw this -- why are you doing everything in double precision?  Consumer GPUs generally don't support double precision well and are really slow with it.

mukulkhanna: @erikwijmans Thank you for pointing this out. The model's input happened to be in double precision, which raised an error about mismatch of weight type (float) and input type (double). Instead of changing the input data type, I had casted the model to double type. 

I have made the suggested changes and removed the conversion to double in the latest commit. Thank you

habitat_baselines/il/data/edfe_data.py
mukulkhanna: Sorry I missed this suggestion somehow. Have committed it now. Thank you

habitat_baselines/il/data/edfe_data.py
mukulkhanna: @erikwijmans I was implementing the LMDB dataset for the VQA trainer. I noticed that storing uncompressed numpy arrrays will take about 10x more storage. More storage was okay for the smaller EDFE dataset. But the dataset required for VQA and NAV trainers is huge (avg 95 images per episode, for ~13k episodes = ~1250k images). The normal jpeg dataset saved on disk occupied ~37 GB disk space. The LMDB dataset, however, has taken up ~100 GB for storing episode frames from just 20 scenes (out of 67 train+test scenes). So it is likely to take >300 GB storage all in all.

Just for your information - with the basic file system (jpeg encoded images), the network took about 4 minutes per epoch on a batch size of 50. I think that's pretty reasonable.

Please share your thoughts on the same. Thank you

habitat_baselines/common/base_trainer.py
```diff
@@ -212,3 +220,490 @@ def _pause_envs(
             batch,
             rgb_frames,
         )
+
+
+class BaseILTrainer(BaseTrainer):
```
erikwijmans, line 225: Please put this in a separate file.

habitat_baselines/common/utils.py
```diff
@@ -228,6 +228,42 @@ def generate_video(
...
+    images = []
+
+    for img_tensor in tensor:
+        image = img_tensor.permute(1, 2, 0).cpu().numpy() * 255
+        images.append(image)
+
+    return images
+
+
+def tensor_to_rgb_images(tensor: torch.Tensor) -> List[np.ndarray]:
```
erikwijmans, line 247:
```suggestion
def tensor_to_brg_images(tensor: torch.Tensor) -> List[np.ndarray]:
```
We keep images in RGB format, OpenCV is in BGR.  Looks like this is returning things in BGR.

habitat_baselines/config/eqa/il_edfe.yaml
```diff
@@ -0,0 +1,28 @@
+BASE_TASK_CONFIG_PATH: "configs/tasks/eqa_mp3d.yaml"
+TRAINER_NAME: "edfe"
```
erikwijmans, line 2: I am not sure what EDFE stands for.  Can we just call this EQA?

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,544 @@
...
+                self.disk_cache_exists = True
+                print("Disk cache exists.")
+
+        if not self.disk_cache_exists:
+            """
+            for each scene > load scene in memory > save frames for each
+            episode corresponding to each scene
+            """
+
+            print("Disk cache not present / incomplete.")
```
erikwijmans, line 125: Use logger instead of print

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,544 @@
...
+
+        for idx, pos in enumerate(pos_queue):
+            observation = self.env.sim.get_observations_at(
+                pos.position, pos.rotation
+            )
+            img = observation["rgb"]
+            frame_path = os.path.join(
+                episode_frames_path, "{0:0=3d}".format(idx)
+            )
+            cv2.imwrite(frame_path + ".jpg", img)
```
erikwijmans, line 218:
```suggestion
            cv2.imwrite(frame_path + ".jpg", img[..., ::-1])
```
OpenCV uses BGR.

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def get_frames(self, frames_path, num=0):
+        r"""Fetches frames from disk.
+        """
+        frames = []
+        for img in sorted(os.listdir(frames_path))[-num:]:
+            img_path = os.path.join(frames_path, img)
+            img = cv2.imread(img_path)
```
erikwijmans, line 226:
```suggestion
            img = cv2.imread(img_path)[..., ::-1]
```

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,544 @@
...
+        if self.input_type == "pacman":
+
+            cnn_kwargs = {
+                "num_classes": 40,
+                "pretrained": True,
+                "checkpoint_path": config.EDFE_CKPT_PATH,
+            }
+            self.cnn = MultitaskCNN(**cnn_kwargs)
+            self.cnn.eval()
+            self.cnn.cuda()
```
erikwijmans, line 78: Use `to(device=self.device)` (or similar) instead of `cuda()`.  That way it will work on just CPU also.

habitat_baselines/il/metrics.py
```diff
@@ -0,0 +1,119 @@
...
+            current_stats.append(self.metrics[i])
+
+        self.stats.append(copy.deepcopy(current_stats))
+
+    def get_stat_string(self, mode=1) -> str:
+
+        stat_string = ""
+
+        for k, v in self.info.items():
+            stat_string += "[%s:%s]" % (k, v)
```
erikwijmans, line 69: Use the python3 style string formatting please.

habitat_baselines/il/models/models.py
```diff
@@ -0,0 +1,780 @@
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.autograd import Variable
```
erikwijmans, line 6: No `Variable`, that has been long deprecated.

habitat_baselines/il/models/models.py
```diff
@@ -0,0 +1,780 @@
...
+        initrange = 0.1
+        self.embed.weight.data.uniform_(-initrange, initrange)
+
+    def forward(self, x):
+        N, T = x.size()
+        idx = torch.LongTensor(N).fill_(T - 1)
+
+        # Find the last non-null element in each sequence
+        x_cpu = x.data.cpu()
+        for i in range(N):
```
erikwijmans, line 329: No for loop, this will be unnecessarily slow.  You can just do `idx = (x != self.NULL).long.sum(-1) - 1`

habitat_baselines/il/models/models.py
```diff
@@ -0,0 +1,780 @@
...
+
+        assert planner_hidden_index.max().item() <= planner_states.size(1)
+
+        planner_hidden_index = (
+            planner_hidden_index.contiguous()
+            .view(N_p, planner_hidden_index.size(1), 1)
+            .repeat(1, 1, planner_states.size(2))
+        )
+
+        # controller_hidden_in = planner_states.cpu().gather(
```
erikwijmans, line 727: Remove commented out code.

habitat_baselines/il/trainers/eqa_trainer.py
```diff
@@ -0,0 +1,52 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from habitat import logger
+from habitat_baselines.common.base_trainer import BaseILTrainer
+from habitat_baselines.common.baseline_registry import baseline_registry
+from habitat_baselines.common.tensorboard_utils import TensorboardWriter
+
+
+@baseline_registry.register_trainer(name="eqa")
+class EQATrainer(BaseILTrainer):
```
erikwijmans, line 14: Is this used anywhere?

habitat_baselines/il/trainers/nav_trainer.py
```diff
@@ -0,0 +1,561 @@
...
+from habitat_baselines.common.baseline_registry import baseline_registry
+from habitat_baselines.common.tensorboard_utils import TensorboardWriter
+from habitat_baselines.il.data.data import EQADataset
+from habitat_baselines.il.metrics import NavMetric
+from habitat_baselines.il.models.models import (
+    MaskedNLLCriterion,
+    NavPlannerControllerModel,
+)
+
+os.environ["CUDA_LAUNCH_BLOCKING"] = "1"
```
erikwijmans, line 27: Remove this

habitat_baselines/il/trainers/nav_trainer.py
```diff
@@ -0,0 +1,561 @@
...
+
+
+@baseline_registry.register_trainer(name="nav")
```
erikwijmans, line 30: NavTrainer is a confusing name for this since it is tied to EQA/PACMAN.

habitat_baselines/il/trainers/nav_trainer.py
```diff
@@ -0,0 +1,561 @@
...
+                    controller_outs = controller_outs.cuda()
+                    controller_action_lengths = (
+                        controller_action_lengths.cuda()
+                    )
+                    controller_masks = controller_masks.cuda()
+
+                    (
+                        planner_action_lengths,
+                        perm_idx,
+                    ) = planner_action_lengths.sort(0, descending=True)
```
erikwijmans, line 140: I assume sorting is done here for using https://pytorch.org/docs/stable/nn.html?highlight=pack_padded#torch.nn.utils.rnn.pack_padded_sequence ?  If so, that doesn't requiring sorting anymore, so this is unneeded!

habitat_baselines/il/trainers/nav_trainer.py
```diff
@@ -0,0 +1,561 @@
...
+        )
+
+        logger.info("train_loader has %d samples" % len(nav_dataset))
+
+        q_vocab_dict, _ = nav_dataset.get_vocab_dicts()
+
+        model_kwargs = {"q_vocab": q_vocab_dict}
+        model = NavPlannerControllerModel(**model_kwargs)
+
+        planner_lossFn = MaskedNLLCriterion()
```
erikwijmans, line 74:
```suggestion
        planner_loss_fn = MaskedNLLCriterion()
```

habitat_baselines/il/trainers/nav_trainer.py
```diff
@@ -0,0 +1,561 @@
...
+        controller_lossFn = MaskedNLLCriterion()
```
erikwijmans, line 75:
```suggestion
        controller_loss_fn = MaskedNLLCriterion()
```

habitat_baselines/config/eqa/il_edfe.yaml
mukulkhanna: EDFE is an Encoder-Decoder based Feature Extractor - a pretrained version of which is used for extracting features from images for the VQA trainer, the NAV trainer (PACMAN) and the RL-based joint fine-tuning trainer (which I thought of as calling as the EQA trainer). I have provided more information about the role of the three trainers [here](https://github.com/mukulkhanna/habitat-api/blob/eqa-1/habitat_baselines/il/README.md).

In my understanding, a total of four trainers would be required - EDFE (feat extractor), VQA, NAV/PACMAN and the (to-be created) EQA trainer.

habitat_baselines/il/data/edfe_data.py
mukulkhanna: @erikwijmans Could you please share your thoughts on the previous comment?

habitat_baselines/config/eqa/il_edfe.yaml
erikwijmans: If I was a person looking at this for the first time, and I saw EDFE I would have no idea what that is or that it is related to EQA.  What about `eqa-cnn-pretrain`?  Or `semseg-ae-depth-cnn-pretrain` if you want to decouple it from EQA.

mukulkhanna: I'll go with `eqa-cnn-pretrain` then. Thanks

habitat_baselines/il/trainers/eqa_trainer.py
mukulkhanna: Not yet. I had made a template for the future where I'd put the joint trainer.
I can remove it if you like.

mukulkhanna:
> Left an initial set of comments. If possible, can you break up this into 2 or 3 different PRs? Keeping track of 4k lines of code at once is going to be hard.

Thanks for the comments @erikwijmans. I have updated the code as per most of the suggestions. 

I have also created a PR for the `eqa-cnn-pretrain` trainer implementation at #430. I think we can start with this. I can create PRs for other trainers as we go along.

Thank you.

habitat_baselines/il/data/edfe_data.py
erikwijmans: Storing things as jpeg in a database is fine, but storing things as individual images is really inefficient, especially if the access will be done via a networked filesystem as is the case on the HPC clusters that are common in DL.   An option going forward will be this: https://github.com/tmbdev/webdataset as it is also likely getting integrated into pytorch: https://github.com/pytorch/pytorch/issues/38419 and is easier to work with than lmdb.

mathfac:
Hi @mukulkhanna!  What functionality of the current PR left comparing to separate PR's you have already merged?

mukulkhanna:
Hi @mathfac. Only the RL-based finetuning of the NAV-PACMAN model (joint with VQA module) for EQA is left for completing the overall implementation.

We can close this PR because it serves no actual purpose other than being just a tracker for EQA sub-tasks.

PR #392, opened 2020-05-10
add utility func for computing local state change

Xiaoming-Zhao:
## Motivation and Context

This PR follows the discussion in #390, adding a utility function for computing state change w.r.t the local coordinate system.

Actually, the code discussed in the issue is not correct. Please consider the situation where the agent only rotates without any movement. In such a situation, position change should be zeros. However, the code in #390 will give us a non-zero result. This PR corrects the position change part.

I think a more preferred logic is to compose a homography/transformation matrix which transforms points from **global coordinate system** to **local coordinate system** for every timestep. Then we use two homography/transformation matrices from consecutive timesteps to construct the matrix, which transforms points between two consecutive **local coordinate system**. After changing the rotation sub-matrix to a quaternion representation, we reduce to the current version of the code. 

## How Has This Been Tested

Test various situations and verify the results.

## Types of changes

New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/habitat/utils/geometry_utils.py
+++ b/habitat/utils/geometry_utils.py
@@ -44,6 +44,14 @@ def quaternion_from_two_vectors(v0: np.array, v1: np.array) -> np.quaternion:
     return np.quaternion(s * 0.5, *(axis / s))
 
 
+def quaternion_xyzw_to_wxyz(v: np.array):
+    return np.quaternion(v[3], *v[0:3])
+
+
+def quaternion_wxyz_to_xyzw(v: np.array):
+    return np.quaternion(*v[1:4], v[0])
+
+
 def quaternion_to_list(q: np.quaternion):
     r"""Creates coeffs in [x, y, z, w] format from quaternions
     """
--- a/habitat/utils/geometry_utils.py
+++ b/habitat/utils/geometry_utils.py
@@ -7,6 +7,9 @@
 import numpy as np
 import quaternion
 
+from habitat.core.simulator import AgentState
+from habitat.tasks.utils import quaternion_from_coeff, quaternion_rotate_vector
+
 EPSILON = 1e-8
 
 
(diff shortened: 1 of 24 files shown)


Skylion007:
Make sure to install commit hooks so it runs flake8 and the other linters before committing so you don't fail these linter tests in the future.

Skylion007:
Not sure the function should live in nav.py. It seems like it could be useful more generically than navigation. Thoughts @erikwijmans ? Maybe in geometry_utils?

erikwijmans:
Thank you for submitting a PR!  Agreed with @Skylion007, moving this to `habitat-api/utils/geometry_utils.py` would be better.  This method also needs a doc string explaining the input/output and why you may want to use it.

Xiaoming-Zhao:
It seems like there is a circular import between `habitat/utils/geometry_utils.py` and `habitat/core/simulator.py` since we need the definition of `AgentState` in this utility function. Do we really need to put this function into `habitat/utils/geometry_utils.py`? Or some suggestions about removing the requirement of `AgentState`? Thanks

Skylion007:
You could just make it take and return a position / rotation tuple. No
reason it has to be an AgentState explicitly.

Xiaoming-Zhao:
I think several (if not all) functions from `habitat/tasks/utils.py` may be moved to `utils/geometry_utils.py` to make it more self-contained. Currently, there are two duplicated functions, namely `quaternion_to_list_xyzw` and `quaternion_to_list`. I think it may cause some conflicts if changing the name in-place. But I personally prefer `quaternion_to_list_xyzw` since it is more precise.

Xiaoming-Zhao:
@erikwijmans Thanks a lot for your clarification. I have made all computations consistent with the `xyzw` format.

habitat/utils/geometry_utils.py
```diff
@@ -51,6 +53,81 @@ def quaternion_wxyz_to_xyzw(v: np.array):
...
+    Returns:
+        np.array: The rotated vector
+    """
+    vq = np.quaternion(0, 0, 0, 0)
+    vq.imag = v
+    return (quat * vq * quat.inverse()).imag
+
+
+def agent_state_target2ref(
+    ref_agent_state: List, target_agent_state: List
```
Skylion007, line 84: Typing seems a bit too strict here, won't this error if fed in an np.array or a tuple which should also both be valid?

Xiaoming-Zhao: Hi, I think np.array may not be a good idea since the lengths of `position` and `rotation` are different. If we want to specify the np.array, an alternative way is to separate a single argument `agent_state` into two, namely `position` and `rotation`. How do you like this?

Skylion007: Oh, right forgot they had different lengths. Just make it a ```Union[List,Tuple]```

Skylion007:
Note: @Xiaoming-Zhao Changed the function and removed the duplicated functions in #398 . 

My bad, should have noticed the duplicated functions / checked our conventions on rotation, position order before merging.

Xiaoming-Zhao:
> Note: @Xiaoming-Zhao Changed the function and removed the duplicated functions in #398 .
> 
> My bad, should have noticed the duplicated functions / checked our conventions on rotation, position order before merging.

@Skylion007 Sorry for the inconvenience. Thanks a lot for your cleanup.

PR #393, opened 2020-05-11
Build api doc in CI

dhruvbatra:
## Motivation and Context

Addresses: 
https://github.com/facebookresearch/habitat-sim/pull/615#discussion_r422681983

## How Has This Been Tested

Same piece of code as on h-sim CI config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,7 +215,6 @@ jobs:
               cd docs
               conda install -y -c conda-forge doxygen==1.8.16
               conda install -y  jinja2 pygments docutils
-              conda install -y -c conda-forge pelican
               mkdir -p ../build/docs
               ./build-public.sh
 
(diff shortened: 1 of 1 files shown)


erikwijmans:
Hmm, looks like habitat-api documentation building requires habitat-sim docs to be built (which makes sense, they cross link between each other).

Adding 

```bash
pushd ../../habitat-sim

python setup.py install --headless

pushd docs

mkdir -p ../build/docs
./build-public.sh

popd
popd
```

before the current call to `./build-public.sh` for habitat-api should fix.  The rebuild is because we are using python=3.8 for the doc build.

dhruvbatra:
Thanks for catching that. But perhaps it's cleaner to create a separate job for building habitat-sim docs (like in the habitat-sim repository)? So that we disentangle what is causing tests to fail.

dhruvbatra:
How's this?

erikwijmans:
Yeah, that seems better!  Looks like habitat-api's CI isn't setup to install bullet however, so the build with bullet is failing.

mosra:
I don't think it's needed to build the full sim docs here, only download the `objects.inv` file from a previous build. Testing a thing right now, will report back when the CI stops failing for me.

mosra:
Opened a PR to your PR: #396

dhruvbatra:
@mathfac -- this is ready for review and merging.

PR #396, opened 2020-05-11
Build doc in CI, on Python 3.6, taking a shortcut and not building sim docs

mosra:
## Motivation and Context

Took #393 and made a bunch of changes:

- Python 3.8 is no longer needed, addressed with facebookresearch/habitat-sim#627
- A full build of habitat-sim docs isn't needed, only the `inv` file generated by those which we can download from elsewhere (and if it's stale it's not an issue either, as the docs from here aren't uploaded or used in any way)
- Latex is needed

What would be ideally done is downloading the `inv` file not from aihabitat.org but directly from the habitat-website repo (to avoid the CI jobs failing when the site is not up, when certs expire etc). I wasn't able to convince github to give me that via a plain curl command, would probably need some access tokens etc. Feel free to iterate on this further.

## How Has This Been Tested

Builds green :heavy_check_mark: https://circleci.com/gh/facebookresearch/habitat-api/1855

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,6 +200,13 @@ jobs:
               . activate habitat; cd habitat-api
               python setup.py develop --all
 
+              # Download sim inventory for crosslinking (no need to build
+              # the whole sim docs for that)
+              # TODO: take it from github.com/facebookmicrosites/habitat-website
+              #   instead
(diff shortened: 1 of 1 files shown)


dhruvbatra:
Changes look great!

PR #405, opened 2020-05-20
Seed fix for dataset

dexter1691:
## Motivation and Context
Right now the seed doesn't get fixed before the dataset object is created resulting in unpredictable order of episodes. Setting random and np.random seed before the `make_dataset` fixes it. 

## How Has This Been Tested
Manually tested that list of episodes is the same across runs. 

## Types of changes
- Fixing the seeds in env_utils.py before the dataset is created. 

## Checklist

### DIFF
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -40,6 +40,7 @@ _C.ENVIRONMENT.ITERATOR_OPTIONS.NUM_EPISODE_SAMPLE = -1
 _C.ENVIRONMENT.ITERATOR_OPTIONS.MAX_SCENE_REPEAT_EPISODES = -1
 _C.ENVIRONMENT.ITERATOR_OPTIONS.MAX_SCENE_REPEAT_STEPS = int(1e4)
 _C.ENVIRONMENT.ITERATOR_OPTIONS.STEP_REPETITION_RANGE = 0.2
+_C.ENVIRONMENT.ITERATOR_OPTIONS.SEED = _C.SEED
 # -----------------------------------------------------------------------------
 # TASK
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mathfac:
Thank you for opening this PR and improving reproducibility. To make the change more consistent for other usecases I would add random seed as part of episode iterator arguments [here](https://github.com/facebookresearch/habitat-api/blob/1b0af67bb24c1dd65214a482f2296e651762f474/habitat/core/dataset.py#L338) and pass SEED from config there.

erikwijmans:
I like the spirit of adding the random seed to the config, but it seems like a band-aid not a fix.  There will slowly become (even more) of a creep of "yes but did you set ALL the random seeds?".

In terms of band-aids, I think just adding `self.seed(config.SEED)` here: https://github.com/erikwijmans/habitat-api-private/blob/tips-and-tricks-pointnav/habitat/core/env.py#L72 would do it?

That won't fix a theoretical issues where dataset loading itself is randomized, @dexter1691's original band-aid would have fixed that, but at a narrower scope.  The alternative would be to have env pass the top level seed to it the iterator.

dexter1691:
>There will slowly become (even more) of a creep of "yes but did you set ALL the random seeds?".
exactly! I didn't do this earlier for the same reason. And it seems kinda odd, we are setting global seeds and now if the global seed and the dataset specific seeds are different, we will not know what the seed is. 

>AFAIK, all dataset loading is deterministic currently
Yeah, it's deterministic right now. The episode iterator is the only part where randomized is used and that doesn't get called until after environment is created.

erikwijmans:
> Hmm, fixing the seed before any of the environments / dataset are created will fix the global seed for everything right?  

Yeah, what I meant by narrower scope is that it’ll fix it for that code path to spin the environments up, but no others.

dexter1691:
Actually calling self.seed() before the dataset episodes are fetched won't work since seed() also runs self._sim.seed() and and self._task.seed() but sim and task haven't been initialized yet. I think the latest commit fixes all the above concerns.

dexter1691:
@erikwijmans: you will have to merge this :) I don't have write access.

erikwijmans:
Oh yeah :)

PR #418, opened 2020-06-18
One less bridge

erikwijmans:
## Motivation and Context

Use inheritance instead of a member variable for the instance of habitat-sim.  This alleviates the needs for `_sim._sim` to get to methods where a bridge hasn't been built.

The one downside is that inheritance won't allow both habitat_sim.Simulator and HabitatSimulator to have their existing `self.config`.  The one in for HabitatSimulator was renamed `hab_config`.

## How Has This Been Tested

The existing tests

## Types of changes

## Checklist


mathfac:
Really like the change! Can we use `hab_config => habitat_config` 4 characters more , but better readability. Thank you!

Issue #404, opened 2020-05-19
Implement Registry for policies for Habitat Baselines


mathfac:
## 🚀 Feature

Implement Registry for policies([similar that we have for baselines](https://github.com/facebookresearch/habitat-api/blob/master/habitat_baselines/common/baseline_registry.py)) for Habitat Baselines than we can avoid patterns like [this](https://github.com/jacobkrantz/VLN-CE/blob/master/vlnce_baselines/dagger_trainer.py#L254).

## Motivation
Often we want to swap policy architecture, but use the same trainer.

Skylion007:
Closed by #440

PR #440, opened 2020-08-05
Implemented PolicyRegistry

rpartsey:
## Motivation and Context

Often we want to swap policy architecture but use the same trainer.
See issue #404.

## How Has This Been Tested
Tests and pre-commit hooks passed successfully.

## Types of changes

## Checklist


habitat_baselines/common/policy_registry.py
```diff
@@ -0,0 +1,45 @@
...
+
+from typing import Optional
+
+from habitat.core.registry import Registry
+
+
+class PolicyRegistry(Registry):
+    @classmethod
+    def register_policy(cls, to_register=None, *, name: Optional[str] = None):
+        r"""Register a RL policy to registry with key 'name'.
```
Skylion007, line 27: The doc strings are unfortunately not in the right format for our documentation system. Please see the `:param:` syntax we use elsewhere.

habitat_baselines/common/policy_registry.py
```diff
@@ -0,0 +1,45 @@
...
+
+        """
+        from habitat_baselines.rl.ppo.policy import Policy
+
+        return cls._register_impl(
+            "policy", to_register, name, assert_type=Policy
+        )
+
+    @classmethod
+    def get_policy(cls, name):
```
Skylion007, line 41:
```suggestion
    def get_policy(cls, name : str):
```

rpartsey:
**To discuss:** How to properly handle instantiating policies dynamically

1. Come up with a standard interface for policies?
2. Use similar approach as `make_task`?
3. Other ...

habitat_baselines/common/policy_registry.py
Skylion007, line 41: Also should have a doc string.

habitat_baselines/common/policy_registry.py
rpartsey: Thanks!
Will fix that
I formatted docstring the same way as in the [BaselineRegistry](https://github.com/facebookresearch/habitat-api/blob/7f2121e1a31d304b5e897c8a97929de44b4f02aa/habitat_baselines/common/baseline_registry.py#L29).

Skylion007: Actually, you might be right that we may not build docs for habitat_baselines, or habitat-api. Perhaps we should consider doing so if we don't.

rpartsey: @Skylion007  Updated!

habitat_baselines/common/policy_registry.py
rpartsey: Updated
**Note:** other `get_somethint` methods in the Registry and BaselineRegistry don't have type annotations.

rpartsey: Updated
**Note:** other `get_somethint` methods in the Registry and BaselineRegistry don't have docstrings.

Skylion007: Can't hurt 👍

habitat_baselines/common/policy_registry.py
```diff
@@ -0,0 +1,65 @@
...
+
+-   Register a policy: ``@policy_registry.register_policy``
+"""
+
+from typing import Optional
+
+from habitat.core.registry import Registry
+
+
+class PolicyRegistry(Registry):
```
erikwijmans, line 26: This doesn't need to be a separate class.  The two methods can be added to the existing BaselineRegistry.  The `"policy"` arg used bellow acts as an additional key so it won't conflict with anything else there.

rpartsey: Done!

rpartsey:
I'd also suggest creating a new issue to standardize the docstrings, type annotations, and methods definition order of `Registry` and `BaselineRegistry` classes.

What do you think?

Skylion007:
Sounds good!

habitat_baselines/common/policy_registry.py
rpartsey: @Skylion007 would you mind resolving this change request?
I've already updated code according to your comments

habitat_baselines/config/test/ppo_imagenav_test.yaml
```diff
@@ -13,6 +13,9 @@ LOG_INTERVAL: 100
 CHECKPOINT_INTERVAL: 1
 
 RL:
+  POLICY:
+    name: "PointNavBaselinePolicy"
```
mathfac, line 17: Please, add default "POLICY.name" value to the default config [here](https://github.com/facebookresearch/habitat-api/blob/master/habitat_baselines/config/default.py#L51). Then we can drop adding policy in part of configs. I would set default to `PointNavBaselinePolicy` as PPO trainer is default one.

habitat_baselines/rl/ppo/policy.py
```diff
@@ -88,8 +89,11 @@ def forward(self, x):
         return self.fc(x)
 
 
+@baseline_registry.register_policy
 class PointNavBaselinePolicy(Policy):
-    def __init__(self, observation_space, action_space, hidden_size=512):
+    def __init__(
+        self, observation_space, action_space, hidden_size=512, **kwargs
```
mathfac, line 95: Maybe, worth to add ` **kwargs` to [PointNavResNetPolicy](https://github.com/facebookresearch/habitat-api/blob/master/habitat_baselines/rl/ddppo/policy/resnet_policy.py#L34).

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -59,10 +59,17 @@ def _setup_actor_critic_agent(self, ppo_cfg: Config) -> None:
         """
         logger.add_filehandler(self.config.LOG_FILE)
 
-        self.actor_critic = PointNavBaselinePolicy(
+        policy = baseline_registry.get_policy(self.config.RL.POLICY.name)
+        self.actor_critic = policy(
             observation_space=self.envs.observation_spaces[0],
             action_space=self.envs.action_spaces[0],
             hidden_size=ppo_cfg.hidden_size,
+            rnn_type=self.config.RL.DDPPO.rnn_type,
```
mathfac, line 67: I had a second look on this design choice, accessing `RL.DDPPO` looks strange here, as it's different baseline. Let's remove new unrelated arguments and let's the run fails if wrong Policy was specified.

habitat_baselines/config/test/ppo_imagenav_test.yaml
rpartsey: Done!

habitat_baselines/rl/ppo/policy.py
rpartsey: Done!

habitat_baselines/rl/ppo/ppo_trainer.py
rpartsey: Done!

habitat_baselines/common/baseline_registry.py
```diff
@@ -61,5 +64,42 @@ def register_env(cls, to_register=None, *, name: Optional[str] = None):
     def get_env(cls, name):
         return cls._get_impl("env", name)
 
+    @classmethod
+    def register_policy(cls, to_register=None, *, name: Optional[str] = None):
+        r"""Register a RL policy to registry with :p:`name`.
```
mathfac, line 69:
```suggestion
        r"""Register a RL policy with :p:`name`.
```

PR #442, opened 2020-08-06
Use habitat-sim's topdown view

erikwijmans:
## Motivation and Context

Habitat-sim can generate the topdown occupancy grid by itself, so refactor things here to a) use that and b) allow the code here to be used to visualization the top-down map without needing to spin-up habitat-api (or even a simulator, just the pathfinder is needed).

I also de-duplicated some code that was in both habitat-baselines and the shortest path follower example.

@aclegg3 Let me know if there is something else from https://github.com/facebookresearch/habitat-sim/pull/718 that you need here.

## How Has This Been Tested

The shortest path follower example produces nicer looking outputs as now the map can be higher res!

## Types of changes


erikwijmans:
> Looks good, thanks for doing this conversion! 😁
> 
> Minor example fix to add height required. I use the scene bounding box to find a guess for the ground floor, but maybe you have another method.

Yeah... getting a height is a tricky thing.  The scene bounding box will be mostly correct for getting the ground floor, but getting the other level surfaces.....

aclegg3:
> > Looks good, thanks for doing this conversion! 😁
> > Minor example fix to add height required. I use the scene bounding box to find a guess for the ground floor, but maybe you have another method.
> 
> Yeah... getting a height is a tricky thing. The scene bounding box will be mostly correct for getting the ground floor, but getting the other level surfaces.....

Yeah, for the tests ground floor is ok I imagine, or a golden number for scenes we expect. 

For real applications I guess we could write a utility function to estimate this by binning NavMesh triangle areas into vertical discrete sections. Could then visualize the range and select floors from some subset of bins or something. 🤷‍♂️  Otherwise users would need to define this manually for their data.

PR #445, opened 2020-08-07
habitat-lab tutoriall

abhiskk:
TODO: revert the pre-commit config editing, currently this is required because pre-commit was creating a cyclic file modification loop.

PR #449, opened 2020-08-13
Interactive task tutorial

dexter1691:
## Motivation and Context

Adding the tutorial for interactive task

## How Has This Been Tested

By running the interactive task ipython notebook and the corresponding python script

## Types of changes

## Checklist


Skylion007:
Great, can you add a quick test of the python script similar to #445 ? You can just take the test/test_examples.py script and change the params of the test function to your script.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+import numpy as np
+import torch
+
+import habitat
+from habitat import Config, Env, RLEnv, VectorEnv, make_dataset
+from habitat.config import get_config
+from habitat_baselines.config.default import get_config as get_baseline_config
+
+# only run this if in colab environment
+if "google.colab" in sys.modules:
```
Skylion007, line 2145: No, we should definitely test this locally in the CI too... Why is this in the if statement like this. Unit testing this part is an important part of the script.

dexter1691: But running the whole training process will take a while!

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+        )
+    )
+
+
+try:  # Got to make initialization idiot proof
+    env.close()
+except NameError:
+    pass
+
+env = habitat.Env(config)
```
Skylion007, line 1382: Since this is all one cell, maybe you should use with `habitat.Env(config) as env:` to demonstrate good practice on closing the environment when you are done with it. Should also make the colab more stable by closing the env on errors

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+    vut.display_video(
+        "data/videos/episode=0-ckpt=0-object_to_goal_distance=0.25-agent_to_object_distance=0.90-episode_success=1.00.mp4"
+    )
+
+# %%
+# %load_ext tensorboard
+# %tensorboard --logdir data/tb
+
+# %%
+if os.path.exists("configs/rl.yaml"):
```
Skylion007, line 2166: Why not just remove the files as soon as you are done with them instead of waiting til then end?

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
Skylion007: @dexter1691 Can we short circuit locally? Like just run 1 or two iterations? And have that number be higher in the notebook?

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+    print(
+        "Gripped Object: {}, Distance To Object: {}, Distance To Goal: {}".format(
+            obs["gripped_object_id"],
+            metrics["agent_to_object_distance"],
+            metrics["object_to_goal_distance"],
+        )
+    )
+
+
+try:  # Got to make initialization idiot proof
```
Skylion007, line 1377: If below is the only place you construct an env, using the with syntax will make this part of the code unnecessary.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+# %% [markdown]
+# ## Implement a hard-coded and an RL agent
+#
+#
+
+# %%
+# @title Load the `RearrangementTask` in Habitat-Lab and run a hard-coded agent
+import habitat
+
+config_file = """
```
Skylion007, line 1329: You could parameterize these config files with format strings and have the user specify the values with sliders / other widgets if you wanted to. Just like we do in the interactivity tutorial. 

You can also contruct these configs as recursive python directionaries. They are yamls after all. You can generate how the dictionaries by running yaml.loads

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
dexter1691: Yeah, I guess we can do that. 👍

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+# %%
+# @title Spawn the agent at a pre-defined location
+
+try:  # Got to make initialization idiot proof
+    sim.close()
+except NameError:
+    pass
+
+cfg.sim_cfg.default_agent_id = 0
+sim = habitat_sim.Simulator(cfg)
```
Skylion007, line 327: You should be careful the sim is closed before the env is opened. Remember, env will implicitly open a sim object and crash anyhow

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2168 @@
...
+
+
+# %%
+# @title Util functions to visualize observations
+# @markdown - `make_video_cv2`: Renders a video from a list of observations
+# @markdown - `simulate`: Runs simulation for a given amount of time at 60Hz
+# @markdown - `simulate_and_make_vid` Runs simulation and creates video
+
+
+def make_video_cv2(
```
Skylion007, line 117: convert to new utility function once it lands

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
Skylion007: and in that case vut.is_notebook() might be a better way to check

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
Skylion007: you should be more worried about closing sim objects than env objects it looks like.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
dexter1691: Yeah, I was wondering about this. Initially I thought, I will just write it to a yaml file to tell the user that look you can just change the config YAML file and load it back. But this looks more ugly. Will change it .

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2120 @@
...
+baseline_config.RL.SUCCESS_MEASURE = "object_to_goal_distance"
+baseline_config.RL.REWARD_MEASURE = "object_to_goal_distance"
+baseline_config.RL.GRIPPED_SUCCESS_REWARD = 2.5
+
+baseline_config.freeze()
+random.seed(baseline_config.TASK_CONFIG.SEED)
+np.random.seed(baseline_config.TASK_CONFIG.SEED)
+torch.manual_seed(baseline_config.TASK_CONFIG.SEED)
+
+trainer = RearrangementTrainer(baseline_config)
```
Skylion007, line 2114: Really surpried this didn't error, still might be best to if __name__ == '__main__': guard here

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2120 @@
...
+
+import numpy as np
+import torch
+
+import habitat
+from habitat import Config, Env, RLEnv, VectorEnv, make_dataset
+from habitat.config import get_config
+from habitat_baselines.config.default import get_config as get_baseline_config
+
+# only run this if in colab environment
```
Skylion007, line 2076: IS this comment still valid?

Skylion007:
@dexter1691 Here is the temporary link to the GIthub hosted colab for testing btw: https://colab.research.google.com/github/dexter1691/habitat-api/blob/tutorial/examples/tutorials/colabs/Habitat_Interactive_Tasks.ipynb

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2120 @@
...
+
+
+# %%
+# @title Dataset class to read the saved dataset in Habitat-API.
+# @markdown To read the saved episodes in Habitat-API, we will extend
+# @markdown the `Dataset` class and the `Episode` base class. It will help
+# @markdown provide all the relevant details about the episode through
+# @markdown a consistent API to all downstream tasks.
+
+# @markdown - We will first create a `RearrangementEpisode` by extending
```
Skylion007, line 522: This text seems like it's been messily wrapped around by Black, you can make it look better by disabling black on these blocks with fmt:off and fmt:on

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2120 @@
...
+# %%
+# @title Spawn the agent at a pre-defined location
+
+try:  # Got to make initialization idiot proof
+    sim.close()
+except NameError:
+    pass
+
+cfg.sim_cfg.default_agent_id = 0
+sim = habitat_sim.Simulator(cfg)
```
Skylion007, line 316: This occurs to me what if you `with sim` over multiple cells, will it automatically close when it leaves scope, if so, this may be a clean way of coding it.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2120 @@
...
+config.DATASET.DATA_PATH = (
+    "data/datasets/rearrangement/coda/v1/{split}/{split}.json.gz"
+)
+config.DATASET.TYPE = "RearrangementDataset-v0"
+config.freeze()
+
+dataset = RearrangementDatasetV0(config.DATASET)
+
+# check if the dataset got correctly deserialized
+assert len(dataset.episodes) == 1
```
Skylion007, line 644: You may want to add error messages to these asserts by adding an extra arg

.circleci/config.yml
```diff
@@ -157,6 +158,16 @@ jobs:
                 unzip habitat-test-scenes.zip
                 rm habitat-test-scenes.zip
               fi
+      - run:
+          name: Download coda scene
+          command: |
+            if [ ! -f ./habitat-sim/data/scene_datasets/coda/coda.glb ]
+            then
+              cd habitat-sim
+              wget --no-check-certificate 'https://docs.google.com/uc?export=download&id=1Pc-J6pZzXEd8RSeLM94t3iwO8q_RQ853' -O coda.zip
```
Skylion007, line 167: OKay, this could break things down the line for sure... We should fix this and properly host it.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
Skylion007: As an update, it should have landed.

.circleci/config.yml
dexter1691: Yes. Can you help me upload this scene to a  Fb's s3 bucket?

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
dexter1691: I have modified the make_video_cv2 for my usecase.

.circleci/config.yml
Skylion007: Yeah sure. Also you should remove __MAC_OSX subfiles from the zip. No reason to include them.

Skylion007:
Getting a weird bug in colab... Exception: Version mismatch: this is the 'cffi' package version 1.14.1, located in '/usr/local/lib/python3.6/dist-packages/cffi/api.py'.  When we import the top-level '_cffi_backend' extension module, we get version 1.11.5, located in '/usr/local/lib/python3.6/dist-packages/_cffi_backend.cpython-36m-x86_64-linux-gnu.so'.  The two versions should be equal; check your installation.

Skylion007:
Grr... seems like I might need to mess with  the install script again...

Skylion007:
The tensorboard also works but I am not seeing any output on it. And rerunning rm -rf data/tb throws an error when run multiple times so you should fix that.

examples/tutorials/colabs/Habitat_Interactive_Tasks.ipynb
```diff
@@ -0,0 +1,2228 @@
...
+   "outputs": [],
+   "source": [
+    "# @title Installation { display-mode: \"form\" }\n",
+    "# @markdown (double click to show code).\n",
+    "\n",
+    "!curl -L https://raw.githubusercontent.com/facebookresearch/habitat-sim/master/examples/colab_utils/colab_install.sh | NIGHTLY=true bash -s\n",
+    "%cd /content\n",
+    "\n",
+    "!gdown --id 1Pc-J6pZzXEd8RSeLM94t3iwO8q_RQ853\n",
+    "!unzip /content/coda.zip -d /content/habitat-sim/data/scene_datasets"
```
Skylion007, line 41: You need to have a flag to allow overwriting or skipping the previous unzip, otherwise it pauses execution and waits for user input

.circleci/config.yml
dexter1691: Great! I have removed the subfiles from the latest version of the zip. It's here: https://drive.google.com/file/d/1Pc-J6pZzXEd8RSeLM94t3iwO8q_RQ853/view?usp=sharing

Skylion007:
@dexter1691 Still not seeing any output on Tensorboard, shouldn't the directory not be deleted until the end of the cell? Does the colab work for you when you run it at the link provided above?

dexter1691:
Yeah. I am able to see the tensorboard output. Maybe you need to click "refresh" button on the tensorboard. The directory is getting deleted before each run so that's fine.

dexter1691:
I also realized that the error due to cffi is probably because colab loads its own version at run-time which gets overwritten. I was able to consistently run the notebook when I restart the runtime after installation step.

Skylion007:
> I also realized that the error due to cffi is probably because colab loads its own version at run-time which gets overwritten. I was able to consistently run the notebook when I restart the runtime after installation step.
Hmm, interesting, I was going to try to pip uninstall cffi to match the appropriate version. (You actually have to uninstall it twice!!!). Does pip uninstalling and then reinstalling the appropriate version work?

I think I am not able to see the Tensorboard output due to the cffi error.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2077 @@
...
+if "google.colab" in sys.modules:
+    os.environ["IMAGEIO_FFMPEG_EXE"] = "/usr/bin/ffmpeg"
+
+repo = git.Repo(".", search_parent_directories=True)
+dir_path = repo.working_tree_dir
+data_path = os.path.join(dir_path, "data")
+output_directory = "data/tutorials/output/"  # @param {type:"string"}
+output_path = os.path.join(dir_path, output_directory)
+
+if not os.path.exists(output_path):
```
Skylion007, line 95: NIT: you don't need to create output is make_video is false? So don't

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2077 @@
...
+
+# %%
+import os
+import time
+from collections import defaultdict, deque
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+import torch
+import tqdm
```
Skylion007, line 1563: You don't seem to actually use this import anywhere? And if you did, you should consider importing tqdm.auto instead as that would select the write progressbar depending on if its python or ipython

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2077 @@
...
+
+                    return
+
+                # episode continues
+                elif len(self.config.VIDEO_OPTION) > 0:
+                    frame = observations_to_image(observations[0], infos[0])
+                    rgb_frames[0].append(frame)
+
+
+# %%
```
Skylion007, line 2018: Why is this a separate cell, you could merge it with the next cell if you wanted to.

Skylion007:
@dexter1691 I found the issue. It's an issue with conda not properly downgrading with CFFI from 1.14 to 1.11. Forcing reinstalling it solves the issue.

Skylion007:
Okay  facebookresearch/habitat-sim#736  is merged in so you may need to update any absolute paths in the script but the CFFI bug should be fixed now.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
dexter1691: I wanted to keep the tensoboard output different from the logs which get generated in the next cell

Skylion007:
So I gave it a run earlier, and it looked like it mostly worked, but the trainer.eval() function never completed. Any idea why? was that fixed in the latest commit?

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2078 @@
...
+            self.set_translation(object_pos, object_id)
+            if isinstance(object_rot, list):
+                object_rot = quat_from_coeffs(object_rot)
+
+            object_rot = quat_to_magnum(object_rot)
+            self.set_rotation(object_rot, object_id)
+
+            self.set_object_motion_type(MotionType.STATIC, object_id)
+
+            # Recompute the navmesh after placing all the objects.
```
mathfac, line 909: Wrong intent, should be moved on one tab left.

examples/tutorials/nb_python/Habitat_Interactive_Tasks.py
```diff
@@ -0,0 +1,2078 @@
...
+
+        if len(existing_object_ids) > 0:
+            for obj_id in existing_object_ids:
+                self.remove_object(obj_id)
+
+        self.sim_object_to_objid_mapping = {}
+        self.objid_to_sim_object_mapping = {}
+
+        if objects is not None:
+            object_template = objects["object_template"]
```
mathfac, line 886: I would expect `objects` is a list, but here it's just one object. Little confusing.

dexter1691: Yeah, for the purpose of this tutorial, I was just doing one object. The point was that these can easily be generalized to more complex scenarios but I wanted to keep all downstream reward computation simple.

dexter1691:
@Skylion007 @mathfac  The latest commit seems to fix all the issues on the colab around CFFI. And a single run end-to-end is working fine for me.

PR #451, opened 2020-08-14
add profiling_utils.py

eundersander:
## Motivation and Context

profiling_utils.py adds helper functions for profiling. Profilers like Nvidia Nsight can capture the time spent in annotated ranges.

I introduce the `HABITAT_PROFILING` environment variable to selectively enable these and future profiling-related stuff. 

Currently, `range_push`/`range_pop` are thin wrappers over the NVTX versions. In the future, we might add a different backend that doesn't depend on Nvidia or that works across multiple machines.

Example usage for marking ranges. For convenience, there are three syntax options; they're all functionally equivalent:
```
# use the decorator to mark an entire function as a range
@profiling_utils.RangeContext("my_function")
def my_function():
    # use range_push/range_pop explicitly
    range_push("create and sort widgets")
    create_widgets()
    sort_widgets()
    range_pop()
    # use RangeContext and a with statement
    with profiling_utils.RangeContext("print and save widgets"):
        print_widgets()
        save_widgets()
```
Example of capturing an Nsight Systems profile:
```
apt-get install nvidia-nsight
export HABITAT_PROFILING=1
path/to/nvidia/nsight-systems/bin/nsys profile --sample=none --trace=cuda,nvtx
--trace-fork-before-exec=true --output=my_profile python my_program.py
# look for my_profile.qdrep in working directory
```

## How Has This Been Tested

I added some test ranges to PPO and DDPPO train scripts. I tested the scripts with and without `HABITAT_PROFILING`, on my Fedora desktop and my devfair machine.

I added `test_profiling_utils.py`, although this only tests that the code runs without error (not behavior correctness). See comments in the file.

## Types of changes

## Checklist

### DIFF
--- a/habitat/utils/profiling_utils.py
+++ b/habitat/utils/profiling_utils.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+r"""Helper functions for profiling. Profilers like Nvidia Nsight can capture the
+time spent in annotated ranges.
+
+Example usage for marking ranges. For convenience, there are three syntax
+options; they're all functionally equivalent:
+
+# use the decorator to mark an entire function as a range
+@profiling_utils.RangeContext("my_function")
+def my_function():
+
+    # use range_push/range_pop explicitly
+    range_push("create and sort widgets")
+    create_widgets()
+    sort_widgets()
+    range_pop()
+
+    # use RangeContext and a with statement
+    with profiling_utils.RangeContext("print and save widgets"):
+        print_widgets()
+        save_widgets()
+
+Example of capturing an Nsight Systems profile:
+apt-get install nvidia-nsight
+export HABITAT_PROFILING=1
+path/to/nvidia/nsight-systems/bin/nsys profile --sample=none --trace=cuda,nvtx
+--trace-fork-before-exec=true --output=my_profile python my_program.py
+# look for my_profile.qdrep in working directory
+"""
+import os
+from contextlib import ContextDecorator
+
+from habitat.core.logging import logger
+
+env_var = os.environ.get("HABITAT_PROFILING", "0")
+enable_profiling = env_var != "0"
+if enable_profiling:
+    logger.info("HABITAT_PROFILING={}".format(env_var))
+    logger.info(
+        "profiling_utils.py range_push/range_pop annotation is enabled"
+    )
+    from torch.cuda import nvtx
+
+
+def range_push(msg: str):
+    r"""Annotates the start of a range for profiling. Requires HABITAT_PROFILING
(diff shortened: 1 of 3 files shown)


habitat/utils/profiling_utils.py
```diff
@@ -0,0 +1,50 @@
...
+Example of capturing an Nsight Systems profile:
+apt-get install nvidia-nsight
+export HABITAT_PROFILING=1
+path/to/nvidia/nsight-systems/bin/nsys profile --sample=none --trace=cuda,nvtx
+--trace-fork-before-exec=true --output=my_profile python my_program.py
+# look for my_profile.qdrep in working directory
+"""
+import os
+
+env_var = os.environ.get("HABITAT_PROFILING", "0")
```
eundersander, line 27: For a another example of using environment variables in Hab-Lab, look at `SLURM_JOB_ID` in `ddp_utils.py`.

habitat/utils/profiling_utils.py
```diff
@@ -0,0 +1,50 @@
...
+enable_profiling = env_var != "0"
+if enable_profiling:
+    print("HABITAT_PROFILING={}".format(env_var))
```
Skylion007, line 30: This should probably use the logger instead of printing here, no?

habitat/utils/profiling_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    print("profiling_utils.py range_push/range_pop annotation is enabled")
+    from torch.cuda import nvtx
+
+
+def range_push(msg: str):
```
Skylion007, line 35: @eundersander The proposed usage is actually not ideal. I would much more recommend creating a context object and using `__enter__` and `__exit__`. That would be much more pythonic and much more resilient to errors.

eundersander: Hey, a context object is a great suggestion. I will add it here, built on top of range_push/range_pop.

I personally find context objects annoying to use for this kind of thing, because they require you to indent huge blocks of code just to add them, creating diff spam and making rapid iteration difficult. But to each their own; it's user preference. I agree they are less error-prone.

Skylion007: @eundersander You are right, the best thing here would be a function decorator!

Skylion007: @eundersander Extend https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator to use it as both!

habitat/utils/profiling_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def range_pop():
+    r"""Annotates the end of a range for profiling. See also range_push."
+    """
+    if enable_profiling:
+        nvtx.range_pop()
+
+
+class RangeContext(ContextDecorator):
```
Skylion007, line 67: This needs a brief docstring explaining it's usage too, right? So that IDEs will load the explanation of what it does? Otherwise looks good.

Issue #452, opened 2020-08-14
Standard interface for policy instantiation


rpartsey:
## 🚀 Feature

Standard interface for policy instantiation.

## Motivation

As described in https://github.com/facebookresearch/habitat-lab/issues/404 it would be nice to be able to change the trainer's policy architecture. The motivation for this issue is to implement the standard interface for policy instantiation so that we can swap them dynamically and avoid hardcoded instantiation ([like](https://github.com/facebookresearch/habitat-lab/blob/4e875493769e0c1b756cf54e8ca5ef826d1ab49a/habitat_baselines/rl/ppo/ppo_trainer.py#L63) this and [this](https://github.com/facebookresearch/habitat-lab/blob/4e875493769e0c1b756cf54e8ca5ef826d1ab49a/habitat_baselines/rl/ddppo/algo/ddppo_trainer.py#L73)).

## Additional context

 The main policy registry functionality was implemented here https://github.com/facebookresearch/habitat-lab/pull/440.

Skylion007:
Closed by #453

PR #453, opened 2020-08-14
Standard interface for policy instantiation.

rpartsey:
## Motivation and Context

Standard interface for policy instantiation.
See https://github.com/facebookresearch/habitat-lab/issues/452 for more details. 

## How Has This Been Tested

Tests in the `test` directory passed.

## Types of changes

-  Refactoring 

## Checklist

### DIFF
--- a/habitat_baselines/rl/ppo/policy.py
+++ b/habitat_baselines/rl/ppo/policy.py
@@ -77,6 +77,10 @@ class Policy(nn.Module):
 
         return value, action_log_probs, distribution_entropy, rnn_hidden_states
 
+    @classmethod
+    def from_config(cls, config, envs):
+        raise NotImplementedError
+
 
 class CriticHead(nn.Module):
     def __init__(self, input_size):
@@ -99,6 +103,14 @@ class PointNavBaselinePolicy(Policy):
                 observation_space=observation_space, hidden_size=hidden_size
             ),
             action_space.n,
+        )
+
+    @classmethod
+    def from_config(cls, config, envs):
+        return cls(
+            observation_space=envs.observation_spaces[0],
+            action_space=envs.action_spaces[0],
+            hidden_size=config.RL.PPO.hidden_size,
         )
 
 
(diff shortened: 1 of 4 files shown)


habitat_baselines/rl/ppo/policy.py
```diff
@@ -77,6 +77,10 @@ def evaluate_actions(
 
         return value, action_log_probs, distribution_entropy, rnn_hidden_states
 
+    @classmethod
+    def from_config(cls, config, envs):
```
Skylion007, line 81: Actually upon review, shouldn't Policy be an abstract class like Net is in the same file? We should probably be consistent in that regard.

rpartsey: Thanks for your comment!
In my humble opinion, it is considered abstract "implicitly".
I can add additional annotations as in the `Net` class if needed.
I'd be grateful to hear what other core contributors think as well.

Skylion007: @erikwijmans @mathfac ^

rpartsey: Updated code!

Should the `Policy`'s `forward` method be left as it is?
Neither `PointNavBaselinePolicy` nor `PointNavResNetPolicy` that inherit from `Policy` implement `forward` method.

erikwijmans: Yeah, let's leave forward as it is.  That raise is there because there isn't a forward method and you aren't expected to implement one (what forward should do is kinda undefined for RL so RL policies don't typically have one)

mathfac: Why do we choose `from_config` method instead of a constructor that takes `config, envs`?

rpartsey: In my humble opinion, this option is more flexible. The constructor explicitly says what fields are needed to initialise the object and _from_config_ knows how to "extract" these fields from config end env and passes them to constructor.

There may be cases (hypothetically🙂) when we need to create the object (for example in tests or demo examples) "explicitly" without creating config or env.

mathfac: Explicitly makes sense, but when we pass the config to constructor of particular class that is expected to be node related to particular class and should not include unnecessary parameters. Unfortunately, that not the case here and we pass config node of higher order. That should be good to go.

PR #470, opened 2020-09-01
[WIP] Rearrangement Task

dexter1691:
## Motivation and Context
This PR will add the following for the rearrangement task:
- Dataset Spec & first version of the episodes. 
- New Sensors
- New Measurements
- Grab/Release action using the magic pointer abstraction. 

Creating this PR to keep track of changes and incorporate early feedback. 
## Types of changes
- New Feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -206,6 +206,35 @@ _C.TASK.DISTANCE_TO_GOAL.DISTANCE_TO = "POINT"
 _C.TASK.ANSWER_ACCURACY = CN()
 _C.TASK.ANSWER_ACCURACY.TYPE = "AnswerAccuracy"
 # -----------------------------------------------------------------------------
+# # REARRANGEMENT TASK GRIPPED OBJECT SENSOR
+# -----------------------------------------------------------------------------
+_C.TASK.GRIPPED_OBJECT_SENSOR = CN()
+_C.TASK.GRIPPED_OBJECT_SENSOR.TYPE = "GrippedObjectSensor"
+# -----------------------------------------------------------------------------
+# # REARRANGEMENT TASK ALL OBJECT POSITIONS SENSOR
+# -----------------------------------------------------------------------------
+_C.TASK.OBJECT_POSITION = CN()
+_C.TASK.OBJECT_POSITION.TYPE = "ObjectPosition"
+_C.TASK.OBJECT_POSITION.GOAL_FORMAT = "POLAR"
+_C.TASK.OBJECT_POSITION.DIMENSIONALITY = 2
+# -----------------------------------------------------------------------------
+# # REARRANGEMENT TASK ALL OBJECT GOALS SENSOR
+# -----------------------------------------------------------------------------
+_C.TASK.OBJECT_GOAL = CN()
+_C.TASK.OBJECT_GOAL.TYPE = "ObjectGoal"
+_C.TASK.OBJECT_GOAL.GOAL_FORMAT = "POLAR"
+_C.TASK.OBJECT_GOAL.DIMENSIONALITY = 2
+# -----------------------------------------------------------------------------
+# # OBJECT_DISTANCE_TO_GOAL MEASUREMENT
+# -----------------------------------------------------------------------------
+_C.TASK.OBJECT_TO_GOAL_DISTANCE = CN()
+_C.TASK.OBJECT_TO_GOAL_DISTANCE.TYPE = "ObjectToGoalDistance"
+# -----------------------------------------------------------------------------
+# # OBJECT_DISTANCE_FROM_AGENT MEASUREMENT
+# -----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


habitat/sims/rearrangement/rearrangement_simulator.py
```diff
@@ -0,0 +1,157 @@
...
+from habitat.core.simulator import Config
+from habitat.sims.habitat_simulator.habitat_simulator import HabitatSim
+from habitat_sim.agent import ActionSpec, ActuationSpec
+from habitat_sim.nav import NavMeshSettings
+from habitat_sim.physics import MotionType
+from habitat_sim.utils.common import quat_from_coeffs, quat_to_magnum
+
+
+@registry.register_simulator(name="RearrangementSim-v0")
+class RearrangementSim(HabitatSim):
```
mathfac, line 19: This all should be broken into new members of `HabitatSim` + `RearrangementTask` changes.

habitat/tasks/rearrangement/rearrangement_task.py
```diff
@@ -0,0 +1,392 @@
...
+
+        observations = self._sim._sensor_suite.get_observations(
+            self._sim._prev_sim_obs
+        )
+        return observations
+
+
+@registry.register_task(name="Rearrangement-v0")
+class RearrangementTask(NavigationTask):
+    r"""Embodied Rearrangement Task
```
mathfac, line 384: You can save any additional state as members of `Task` here, instead of HabitatSim. Like if object was gripped or etc.

Skylion007: Agreed, this is a better design way.

habitat/tasks/rearrangement/rearrangement_task.py
```diff
@@ -0,0 +1,392 @@
...
+            object_position = habitat_sim.get_translation(sim_obj_id)
+            sensor_data[obj_id] = self._compute_pointgoal(
+                agent_position, rotation_world_agent, object_position
+            )
+
+        return sensor_data
+
+
+@registry.register_sensor
+class AllObjectGoals(PointGoalSensor):
```
Skylion007, line 272: Why do we inherit from PointGoalSensor, does not make sense to use a higher level abstract sensor? Like a Task_Sensor

Skylion007:
Merge from master to fix the tests failing.

habitat/sims/rearrangement/actions.py
```diff
@@ -0,0 +1,82 @@
...
+import habitat_sim
+from habitat.core.registry import registry
+from habitat.sims.habitat_simulator.actions import (
+    HabitatSimActions,
+    HabitatSimV1ActionSpaceConfiguration,
+)
+from habitat_sim.agent.controls.controls import ActuationSpec
+
+
+def raycast(sim, sensor_name, crosshair_pos=[128, 128], max_distance=2.0):
```
Skylion007, line 21: This utility is great, so great that maybe it should live in a higher level? Also it should be under sims/habitat_sim/ since it relies heavily on habitat_sim specific APIs, right?

habitat/sims/rearrangement/actions.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
```
Skylion007, line 1: Rearrangement is not a sim, you may be extending functionality of Habitat-Sim, but it is not a Sim.

habitat/sims/rearrangement/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    try:
+        import habitat_sim  # noqa: F401
+
+        has_habitat_sim = True
+    except ImportError as e:
+        has_habitat_sim = False
+        habitat_sim_import_error = e
+
+    if has_habitat_sim:
+        from habitat.sims.rearrangement.actions import (  # noqa: F401
```
Skylion007, line 15: You don't need to do the if else check, you can just put these imports into the try as they will all nicely all fail under one control flow and remove the need for the `has_habitat_sim` variable.

habitat/config/default.py
```diff
@@ -218,6 +248,12 @@ def __init__(self, *args, **kwargs):
 _C.SIMULATOR.TURN_ANGLE = 10  # angle to rotate left or right in degrees
 _C.SIMULATOR.TILT_ANGLE = 15  # angle to tilt the camera up or down in degrees
 _C.SIMULATOR.DEFAULT_AGENT_ID = 0
+_C.SIMULATOR.INITIAL_LOOK_DOWN_ANGLE = (
```
Skylion007, line 251: This shouldn't override the defaults globally this will break people's experiments no? It should start at 0 and the task config should override it.

Skylion007, line 251: You can also set this by changing the sensor orientation.

PR #472, opened 2020-09-01
Add UUID config to simulator visual sensors

### LINKED ISSUE #471: Have Default SensorTypes have a YAML Customizable UUIDs
## 🚀 Feature

* Currently, using multiple visual sensors is difficult since each sensor has a UUID that is baked into the default class. It should be able to be overriden by the configs.
## Motivation
It should be able to be overriden for the configs for the default RGBs, Depth Sensors, Semantic etc at the very least. That way, one can define multiple RGB, Depth, or other sensors without having to subclass each sensor.

## Pitch

Being able to define multiple sensors from pure YAML without needing to subclass the model.

Skylion007:
## Motivation and Context

Fixes #471 

## How Has This Been Tested
With existing sensors tests as well as by overriding the UUID

## Types of changes

## Checklist

- [] My change requires a change to the documentation.

### DIFF
--- a/habitat/core/simulator.py
+++ b/habitat/core/simulator.py
@@ -105,10 +105,15 @@ class Observations(dict):
 
 class RGBSensor(Sensor):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.uuid = "rgb"
         super().__init__(*args, **kwargs)
+        if "config" in kwargs:
+            config = kwargs["config"]
+            if hasattr(config, "UUID"):
+                self.uuid = config.UUID
 
     def _get_uuid(self, *args: Any, **kwargs: Any) -> str:
-        return "rgb"
+        return self.uuid
 
     def _get_sensor_type(self, *args: Any, **kwargs: Any) -> SensorTypes:
         return SensorTypes.COLOR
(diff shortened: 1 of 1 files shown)


habitat/core/simulator.py
```diff
@@ -105,10 +105,15 @@ def __init__(
...
+        self.uuid = "rgb"
         super().__init__(*args, **kwargs)
+        if "config" in kwargs:
+            config = kwargs["config"]
+            if hasattr(config, "UUID"):
+                self.uuid = config.UUID
 
     def _get_uuid(self, *args: Any, **kwargs: Any) -> str:
-        return "rgb"
+        return self.uuid
```
mathfac, line 116: This all can be done with:
```suggestion
        return self.uuid
         if hasattr(self.config, "UUID"):
           return config.UUID
         else:
           return "rgb"
```

habitat/core/simulator.py
```diff
@@ -63,7 +63,11 @@ class Sensor:
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.config = kwargs["config"] if "config" in kwargs else None
-        self.uuid = self._get_uuid(*args, **kwargs)
+        if hasattr(self.config, "UUID"):
+            # We allow any sensor config to override the UUID
+            self.uuid = self.config.UUID
+        else:
```
Skylion007, line 69: @mathfac Actually, I am not sure I like this method because calling _get_uuid will return the wrong UUID.

PR #476, opened 2020-09-07
add generate_profile_shell_scripts.py

eundersander:
## Motivation and Context

Reference python script for profiling DDPPO PointNav on the FAIR internal cluster used by the Habitat team. This script can be modified locally to suit your needs, for example, to profile a different program or to profile with different settings.

For an overview of profiling and optimization in Habitat, see:
https://colab.research.google.com/gist/eundersander/b62bb497519b44cf4ceb10e2079525dc/faster-rl-training-profiling-and-optimization.ipynb

## How Has This Been Tested

Ad hoc testing on my Fedora desktop and FAIR cluster.

## Types of changes

## Checklist

### DIFF
--- a/scripts/generate_profile_shell_scripts.py
+++ b/scripts/generate_profile_shell_scripts.py
@@ -33,7 +33,7 @@ if __name__ == "__main__":
     # range is generally a better workflow, but it requires integrating
     # profiling_utils.configure into your train program (beware,
     # profiling_utils.configure is not yet merged into Habitat-sim).
-    do_capture_step_range = False
+    do_capture_step_range = True
 
     if do_capture_step_range:
         # "Step" here refers to however you defined a train step in your train
@@ -42,10 +42,12 @@ if __name__ == "__main__":
         # terms of the time spent in various parts of your program. Early train
         # steps may suffer from poor agent behavior, too-short episodes, etc. If
         # necessary, capture and inspect a very long-duration profile to
-        # determine when your training "settles". DDPPO PointNav empirical test
-        # from Aug 2020: settled around 150-190, after which it was stable out
-        # to ~1400 steps.
-        capture_start_step = 190
+        # determine when your training FPS "settles".
+        # DDPPO PointNav empirical test from Aug 2020, 8 nodes:
+        #   FPS settled at ~190 steps
+        # DDPPO PointNav empirical test from Oct 2020, 2 nodes:
+        #   FPS settled at ~1200 steps
+        capture_start_step = 1200
 
         # If you're focusing on optimizing the train loop body (work that
         # happens consistently every update), you don't need a large number
@@ -83,11 +85,17 @@ if __name__ == "__main__":
         # duration will cause your slurm job to terminate before profiles are
         # saved. A much-too-large duration may result in a longer wait time
         # before slurm starts your job.
-        # DDPPO PointNav empirical test from Aug 2020:
+        # DDPPO PointNav empirical test from Aug 2020, 8 nodes:
         #   startup time is 2 minutes and 100 steps takes 12 minutes
+        # DDPPO PointNav empirical test from Oct 2020, 2 nodes:
+        #   startup time is 2 minutes and 100 steps takes 5.9 minutes
+        buffered_start_minutes = 10
+        buffered_minutes_per_100_steps = 8
         if do_capture_step_range:
-            slurm_job_termination_minutes = 10 + int(
-                (capture_start_step + num_steps_to_capture) * 15 / 100
+            slurm_job_termination_minutes = buffered_start_minutes + int(
+                (capture_start_step + num_steps_to_capture)
+                * buffered_minutes_per_100_steps
+                / 100
             )
         else:
             slurm_job_termination_minutes = (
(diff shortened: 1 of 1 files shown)


scripts/generate_profile_shell_scripts.py
```diff
@@ -0,0 +1,247 @@
...
+    export_sqlite = True
+
+    # This is the end of the user-facing options, except see slurm_submit_str
+    # below.
+    # =========================================================================
+
+    if do_capture_step_range:
+        program_with_extra_args_str = (
+            program_str
+            + " PROFILING.CAPTURE_START_STEP "
```
eundersander, line 127: I'm adding a PROFILING section to `habitat_baselines/config` in a related PR.
```
_C.PROFILING = CN()
_C.PROFILING.CAPTURE_START_STEP = -1
_C.PROFILING.NUM_STEPS_TO_CAPTURE = -1
```

scripts/generate_profile_shell_scripts.py
```diff
@@ -0,0 +1,247 @@
...
+            + str(capture_start_step)
+            + " PROFILING.NUM_STEPS_TO_CAPTURE "
+            + str(num_steps_to_capture)
+        )
+    else:
+        program_with_extra_args_str = program_str
+
+    if do_capture_step_range:
+        capture_range_args = '--capture-range=nvtx -p "habitat_capture_range" --stop-on-range-end=true'
```
eundersander, line 136: See https://github.com/facebookresearch/habitat-sim/pull/781/files#diff-987330c81014b88396f7d2cd6020b676R130

scripts/generate_profile_shell_scripts.py
```diff
@@ -0,0 +1,247 @@
...
+#SBATCH --job-name=capture_profile
+#SBATCH --output=/checkpoint/%u/jobs/job.%j.out
+#SBATCH --error=/checkpoint/%u/jobs/job.%j.err
+#SBATCH --gpus-per-task 1
+#SBATCH --nodes 8
+#SBATCH --cpus-per-task 10
+#SBATCH --ntasks-per-node 8
+#SBATCH --mem-per-cpu=5GB
+#SBATCH --partition=learnfair
+#SBATCH --time="""
```
eundersander, line 200: I got these slurm parameters from @Skylion007 and I assume they are representative for training DDPPO PointNav.

scripts/generate_profile_shell_scripts.py
```diff
@@ -0,0 +1,247 @@
...
+            + str(slurm_job_termination_minutes)
+            + """:00
+#SBATCH --signal=USR1@300
+#SBATCH --open-mode=append
+export GLOG_minloglevel=2
+export MAGNUM_LOG=quiet
+export MASTER_ADDR=$(srun --ntasks=1 hostname 2>&1 | tail -n1)
+set -x
+srun bash capture_profile_slurm_task.sh
```
eundersander, line 209: If you're wondering why I bother with this nested shell script (instead of invoking nsys directly), it's so that `${SLURM_NODEID}` and other env vars resolve correctly.

scripts/generate_profile_shell_scripts.py
```diff
@@ -0,0 +1,247 @@
...
+#SBATCH --gpus-per-task 1
+#SBATCH --nodes 8
+#SBATCH --cpus-per-task 10
+#SBATCH --ntasks-per-node 8
+#SBATCH --mem-per-cpu=5GB
+#SBATCH --partition=learnfair
+#SBATCH --time="""
+            + str(slurm_job_termination_minutes)
+            + """:00
+#SBATCH --signal=USR1@300
```
erikwijmans, line 203: This should be removed.  Seems like you want `slurm_job_termination_minutes` to be respected and the signal will make the job exit earlier and requeue itself.  Since the `capture_profile_slurm_task.sh` script doesn't have a signal forwarder this signal won't get propagated, that is why it doesn't matter currently.

scripts/generate_profile_shell_scripts.py
erikwijmans: May as well knock this down to 2 nodes.  That'll get scheduled faster and you can submit it to `dev` instead.  NCCL gets a little less efficient once crossing from 1 node to 2 nodes -- although since the FAIR cluster has either 4x or 2x IBs, depending on the node, it doesn't really change things much so you could do just 1.

Skylion007:
@eundersander Can you make sure the generated scripts pass shellcheck? We use it in Habitat-Sim as a pre-commit hook, maybe it's time to add it this repo as well if we continue to add shell scripts.

eundersander:
>Can you make sure the generated scripts pass shellcheck? 
Done. I fixed a few things which are included in the "polish" commit.

scripts/generate_profile_shell_scripts.py
eundersander: One consequence of switching from 8 nodes to 2 is that I now have to run for ~1300 steps instead of ~300. See comments starting at L83.

Skylion007:
Mind dropping in the shellcheck pre-commit hooks, because why not.

scripts/generate_profile_shell_scripts.py
erikwijmans: The `srun` is also added to start 2x8 processes (1 per GPU)!

erikwijmans (approved):

eundersander:
>Mind dropping in the shellcheck pre-commit hooks, because why not.
https://github.com/facebookresearch/habitat-lab/pull/509

PR #477, opened 2020-09-07
add profiling config and ranges for PPO/DDPPO training

eundersander:
## Motivation and Context

This PR adds profiling ranges to PPO/DDPPO training.

It also adds a PROFILING section to our config. In general, we won't author a PROFILING section into our yaml files. We'll just override values via the command line. See https://github.com/facebookresearch/habitat-lab/pull/476/files#diff-9a47df42fe45fb85dac9db9bcaa60eceR124 .

For an overview of profiling and optimization in Habitat, see:
https://colab.research.google.com/gist/eundersander/b62bb497519b44cf4ceb10e2079525dc/faster-rl-training-profiling-and-optimization.ipynb

This PR uses a fairly new `habitat_sim.profiling_utils` API. It's in Hab-sim nightly but not Hab-sim stable. I assume it's safe to merge as long as CI passes. After this merges, Hab-lab users that pull latest will also need to update Hab-sim.

Example output viewed in Nsight GUI tool:
nsight_multithreaded_trace1

Example with CUDA capture:
nsight_multithreaded_trace4

Example with OpenGL capture:
nsight_multithread_trace3

## How Has This Been Tested

Ad hoc testing on my linux desktop and FAIR cluster.

## Types of changes

## Checklist

### DIFF
--- a/habitat/core/env.py
+++ b/habitat/core/env.py
@@ -30,7 +30,7 @@ from habitat.core.simulator import Observations, Simulator
 from habitat.datasets import make_dataset
 from habitat.sims import make_sim
 from habitat.tasks import make_task
-from habitat_sim.utils import profiling_utils
+from habitat.utils import profiling_wrapper
 
 
 class Env:
@@ -365,7 +365,7 @@ class RLEnv(gym.Env):
     def current_episode(self) -> Type[Episode]:
         return self._env.current_episode
 
-    @profiling_utils.RangeContext("RLEnv.reset")
+    @profiling_wrapper.RangeContext("RLEnv.reset")
     def reset(self) -> Observations:
         return self._env.reset()
 
@@ -405,7 +405,7 @@ class RLEnv(gym.Env):
         """
         raise NotImplementedError
 
-    @profiling_utils.RangeContext("RLEnv.step")
+    @profiling_wrapper.RangeContext("RLEnv.step")
     def step(self, *args, **kwargs) -> Tuple[Observations, Any, bool, dict]:
         r"""Perform an action in the environment.
 
--- a/habitat/core/env.py
+++ b/habitat/core/env.py
@@ -30,6 +30,7 @@ from habitat.core.simulator import Observations, Simulator
 from habitat.datasets import make_dataset
 from habitat.sims import make_sim
 from habitat.tasks import make_task
+from habitat_sim.utils import profiling_utils
 
 
 class Env:
@@ -364,6 +365,7 @@ class RLEnv(gym.Env):
     def current_episode(self) -> Type[Episode]:
         return self._env.current_episode
 
+    @profiling_utils.RangeContext("RLEnv.reset")
     def reset(self) -> Observations:
         return self._env.reset()
 
@@ -403,6 +405,7 @@ class RLEnv(gym.Env):
         """
         raise NotImplementedError
 
+    @profiling_utils.RangeContext("RLEnv.step")
     def step(self, *args, **kwargs) -> Tuple[Observations, Any, bool, dict]:
         r"""Perform an action in the environment.
 
--- a/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -33,6 +33,7 @@ from habitat_baselines.utils.common import (
     linear_decay,
 )
 from habitat_baselines.utils.env_utils import construct_envs
+from habitat_sim.utils import profiling_utils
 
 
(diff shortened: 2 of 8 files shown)


habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -180,7 +184,10 @@ def _collect_rollout_step(
 
         t_step_env = time.time()
 
-        outputs = self.envs.step([a[0].item() for a in actions])
+        step_data = [a[0].item() for a in actions]
+        profiling_utils.range_pop()  # compute actions
```
eundersander, line 188: I want my range to end after the call to `item()` but before the call to `envs.step`. I'm relying on `item()` as a blocking/synchronizing call for the preceding GPU inference for computing actions.

habitat/core/env.py
```diff
@@ -20,6 +20,7 @@
 from habitat.datasets import make_dataset
 from habitat.sims import make_sim
 from habitat.tasks import make_task
+from habitat_sim.utils import profiling_utils
```
erikwijmans, line 23: Thus far we have kept habitat-sim as a soft dependency for habitat-lab.  This is currently crucial for the habitat-pyrobot bridge as it lets people not have to worry about figuring out how to compile habitat-sim on the robot platform (and that platform may not have the RAM requirements to compile the python bindings in some cases).

However, I get the desire to share code between habitat-lab and habitat-sim.  Maybe we need to consider having habitat_utils be its own thing?  We could keep that in the habitat-sim repo (to not have a 3rd to manage) and then the CI that does nightly Conda-builds could also upload that to Conda and pip for us since that package would be pure python and platform agnostic?

CC @mathfac

eundersander: Yeah, I tried to bring attention to this a while back. See Slack discussion: https://cvmlp.slack.com/archives/CFN5TAUSD/p1597880436485300 .

eundersander: From that thread, only Dhruv really weighed in:
>Yeah, I think hab-sim being a hard dependency for hab-lab is fine.
But I understand we can't break the pyrobot bridge. I don't know if what I'm doing here is breaking that.

>Maybe we need to consider having habitat_utils be its own thing? We could keep that in the habitat-sim repo (to not have a 3rd to manage) 
This sounds reasonable and is one of the approaches the infra team has previously discussed. So this would be a different bindings binary, set up alongside habitat_sim, but both live in the habitat-sim repo.

What do you want to see for this PR? Change `import habitat_sim` to `import habitat_utils`? Or is that a refactor we can tackle later?

erikwijmans: Yeah, I saw that message and it didn't click that the pyrobot bridge would break by doing that until now.  I am not entirely certain what is the right call for this.

eundersander: One option is to punt on this whole problem for now and just move `profiling_utils` back into Hab-lab. I kinda wanted it available to Hab-sim python code, but that's not super important.

erikwijmans: Yeah, for now it is an easier problem to install habitat-lab to get the profiling tools than it is to make habitat-sim a hard dependency due to pyrobot, but this does bring up a good point of where do we keep common utils.  There haven't be any significant ones yet, so the little ones have just gotten duplicated (see the `quat_rotate_vector` function)

aclegg3: Maybe we want to create a unified `habitat_utils` module (for python only) in habitat-sim repo outside of the other source and place all utils which are not C++ dependent here. Then folks could clone habitat-sim but would not need to build it to access these.

eundersander: Ok. For the 0.1.6 release, I'm going to leave `profiling_utils` in habitat-sim's utils folder.

After code freeze, I'll investigate setting up `habitat_utils`. @erikwijmans , how do I go about testing the pyrobot bridge or verifying it's still working? Is there CI or a unit test?

Skylion007: Can't we just have a dummy class with no op functions if it fails to import?

Skylion007:
Hold on, it's definitely not true for PyRobot and running Habitat-Sim could be difficult there given that some robots could lack ANY gpu.

habitat/core/env.py
eundersander: After all this time, we still haven't found a good way to do this `habitat_utils`, although we're still discussing.

For now, I'll take @Skylion007 's suggested approach of importing with a no-op fallback when the import fails. In other words, habitat_sim is still optional when using habitat-lab (but you don't get profiling if you don't have habitat-sim installed, which is fine).

erikwijmans (approved):

Issue #367, opened 2020-04-08
Observation transformation registry for Habitat baselines to provide input cropping, resizing, etc.


mathfac:
## 🚀 Feature

Observation transformation registry for Habitat baselines to provide input cropping, resizing, etc.

## Motivation
To make training pipeline agnostic to observation resolution or experiment with grayscale observations instead of RGB that would be great to have a list of supported transformations.

Reference [Resize and crop PR](https://github.com/facebookresearch/habitat-api/pull/365/).

## Pitch
1. Implement Transformations registry [here](https://github.com/facebookresearch/habitat-api/blob/master/habitat_baselines/common/baseline_registry.py).
2. Add nodes to [the habitat baselines config](url):
```
OBSERVATION_TRANSFORMATIONS: ["RESIZE", "CENTER_CROP", "GRAYSCALE"]
TRANSFORMATIONS:
    RESIZE:
        WIDTH: 640
        HEIGHT: 480
   CROP:
        WIDTH: 250
        HEIGHT: 250        
```

3. Add a call of the transformations chain to training and evaluation general code.
4. Transformations will change Observations space for models and should be addressed in the code.

## Alternative way
Same transformation can be implemented in Models hierarchy.

srama2512:
Would it be better to have this in the models hierarchy? The data augmentation must not impact the visualization aspect of the trainer, right?

Skylion007:
@srama2512 I think it might be best to have this as part of the policy instantiation introduced by @rpartsey in #452 and #453 .

srama2512:
I haven't been following the policy registry issue. Just checked it out. That's a brilliant feature! And yes, this would fit perfectly there.

Skylion007:
@srama2512 That's not entirely true, there definitely are some 'data' augmentations you want to visualize. Like creating a fake depth sensor that is just a neural network that is fed RGB input. Seeing the results of that would be desirable, so it's not quite that simple.

srama2512:
@Skylion007  - Wouldn't that just be a part of the model definition though? That is, having a depth-prediction module. In this case, the network would only take an RGB input.

Skylion007:
@srama2512 Your comments would be appreciated on #478

Skylion007:
Closed by #478

PR #478, opened 2020-09-08
Overhaul Observation Transformation API + Better 360 and multisensor support.

Skylion007:
## Motivation and Context

Solves #367 

Okay, this ObservationTransformer API does the following:
* Adds a new observation transformer registry
* Videos generated now show the final output of the network after all the observation transformers have been fed in.
* Still not ideal since Observations may need to be normalized/ denormalized in each operation. Should fix this longterm.
* Implemented a demo cubemap2equirect that stitches together 6 sensors into one using this API. Creating a 360 Sensor.
* Remove unnecessary float conversion in PPO trainer.
* Reads in video input from batched input now.
* ObservationTransformers separated from Policies.
* Introduces a new API to run.py that can feed in a config without writing it to disk.
* Add more documentations
* Add more typing to better document what is going on.
* Adds tests to cover the change.
* Changes the policy fromConfig API to not have access to the full environment.

## How Has This Been Tested
With Pytest and by splitting the ResizeCenterCropper into two PyTorch modules

## Types of changes

## Checklist


habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@baseline_registry.register_obs_transformer()
+class ResizeShortestEdge(ObservationTransformer):
+    def __init__(self, size: int, channels_last: bool = False):
+        r"""An nn module the resizes your the shortest edge of the input while maintaining aspect ratio.
+        This module assumes that all images in the batch are of the same size.
+        Args:
+            size: The size you want to resize the shortest edge to
+            channels_list: indicates if channels is the last dimension
```
mathfac, line 51:
```suggestion
            channels_last: indicates if channels is the last dimension
```

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        super().__init__()
+        self._size = size
```
mathfac, line 54: You can try using `@attr.s(auto_attribs=True, kw_only=True)` to avoid this code.

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+        self.channels_last = channels_last
+
+    def transform_observation_space(
+        self,
+        observation_space: SpaceDict,
+        trans_keys=("rgb", "depth", "semantic"),
```
mathfac, line 60: Makes more sense to have list here.
```suggestion
        trans_keys: List[str]=["rgb", "depth", "semantic"],
```

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+    ):
+        size = self._size
+        observation_space = copy.deepcopy(observation_space)
+        if size:
+            for key in observation_space.spaces:
+                if key in trans_keys:
+                    # In the observation space dict, the channels are always last
+                    h, w = get_image_height_width(
+                        observation_space.spaces[key], channels_last=True
```
mathfac, line 69:
```suggestion
                        observation_space.spaces[key], channels_last=self.channels_last
```

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+                    )
+                    logger.info(
+                        "Center cropping observation size of %s from %s to %s"
+                        % (key, (h, w), size)
+                    )
+
+                    observation_space.spaces[key] = overwrite_gym_box_shape(
+                        observation_space.spaces[key], size
+                    )
+        self.observation_space = observation_space
```
mathfac, line 140: Looks like never used.
```suggestion
```

habitat_baselines/common/obs_transformers.py
Skylion007: Should be a tuple per flake8 not a list.

habitat_baselines/common/obs_transformers.py
Skylion007: Nope, observations are always channels last whether the the transformation is or not.

habitat_baselines/common/obs_transformers.py
Skylion007: @mathfac How do I so while preserving the docstring though?

erikwijmans: The docstring for the constructor is generally put on the class instead of `__init__`

habitat_baselines/config/default.py
```diff
@@ -59,6 +59,16 @@
 _C.RL.POLICY = CN()
 _C.RL.POLICY.name = "PointNavBaselinePolicy"
 # -----------------------------------------------------------------------------
+# OBS_TRANSFORMS CONFIG
+# -----------------------------------------------------------------------------
+_C.RL.POLICY.OBS_TRANSFORMS = CN()
+_C.RL.POLICY.OBS_TRANSFORMS.active = tuple()
```
mathfac, line 65: Maybe more explicit and conventional:
```suggestion
_C.RL.POLICY.OBS_TRANSFORMS.ENABLED_TRANSFORMS = tuple()
```

mathfac (approved):
Checked the logic again. Looks more extendable now! Minor comment.  
Thank you for bringing typing to policies code.

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,199 @@
...
+                    for sensor in self.trans_keys
+                    if sensor in observations
+                }
+            )
+        return observations
+
+    @classmethod
+    def from_config(cls, config: Config, envs):
+        return cls(
+            (
```
mathfac, line 180:
```suggestion
           size = (
```

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+        )
+
+
+def get_active_obs_transforms(
+    config: Config, envs
+) -> List[ObservationTransformer]:
+    active_obs_transforms = []
+    if hasattr(config.RL.POLICY, "OBS_TRANSFORMS"):
+        obs_transform_names = config.RL.POLICY.OBS_TRANSFORMS.active
```
mathfac, line 192:
```suggestion
        obs_transform_names = config.RL.POLICY.OBS_TRANSFORMS.ENABLED_TRANSFORMS
```

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
```
mathfac, line 5:
```suggestion

# LICENSE file in the root directory of this source tree.
```

mathfac:
Do we have tests covering these changes? Thank you

habitat_baselines/config/default.py
Skylion007: @mathfac Should we update POLICY as well? I just wanted to be consistent with our .name and stuff.

habitat_baselines/common/utils.py
```diff
@@ -148,11 +99,7 @@ def batch_obs(
             batch[sensor].append(_to_tensor(obs[sensor]))
 
     for sensor in batch:
-        batch[sensor] = (
```
mathfac, line 151: cc @erikwijmans, is this change safe?

habitat_baselines/rl/ddppo/algo/ddppo_trainer.py
```diff
@@ -188,8 +201,10 @@ def train(self) -> None:
 
         observations = self.envs.reset()
         batch = batch_obs(observations, device=self.device)
+        batch = apply_obs_transforms_batch(batch, self.obs_transforms)
 
-        obs_space = self.envs.observation_spaces[0]
+        # obs_space = self.envs.observation_spaces[0]
+        obs_space = self.obs_space
```
mathfac, line 207: Maybe, Remove both lines, and use `self.obs_space` instead of `obs_space`

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -60,7 +65,16 @@ def _setup_actor_critic_agent(self, ppo_cfg: Config) -> None:
         logger.add_filehandler(self.config.LOG_FILE)
 
         policy = baseline_registry.get_policy(self.config.RL.POLICY.name)
-        self.actor_critic = policy.from_config(self.config, self.envs)
+        observation_space = self.envs.observation_spaces[0]
+        self.obs_transforms = get_active_obs_transforms(self.config)
+        observation_space = apply_obs_transforms_obs_space(
+            observation_space, self.obs_transforms
+        )
+        # for active_obs in self.obs_transforms:
```
mathfac, line 73: Comments left.

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,597 @@
...
+    def from_config(cls, config: Config):
+        return cls(
+            (
+                config.RL.POLICY.OBS_TRANSFORMS.CENTER_CROPPER.HEIGHT,
+                config.RL.POLICY.OBS_TRANSFORMS.CENTER_CROPPER.WIDTH,
+            )
+        )
+
+
+class Cube2Equirec(nn.Module):
```
mathfac, line 193: Need tests for this functionality and let's move to separate PR, as soon as this PR will be ready.

habitat_baselines/rl/ddppo/algo/ddppo_trainer.py
Skylion007: This mutates the observation space in a very specific way though and the original code did not modify the observation space return from the envs so I am not sure that is the proper move.

habitat_baselines/common/utils.py
Skylion007: @mathfac All tests pass so presumably and any RGB input is already converted implicitly to float in the encoders.

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,630 @@
...
+    def __init__(self, equ_h, equ_w, cube_length, CUDA=False):
```
erikwijmans, line 216: CUDA really shouldn't be an arg to `__init__` IMO.  You can do 
```
self.grid  = self.grid.to(batch.device)
self.orientation_mask = self.orientation_mask.to(batch.device)
```
in forward and that will be a no-op after the first call.

habitat_baselines/common/utils.py
erikwijmans: Yeah, it should be fine

Skylion007:
All tests are now passing. Ready to get this merged. Only thing I am unsure of is the ObsTransformer name and if I should move it from the RL.POLICY to just RL.

habitat_baselines/common/obs_transformers.py
```diff
@@ -0,0 +1,623 @@
...
+        return observations
+
+    @classmethod
+    def from_config(cls, config: Config):
+        return cls(config.RL.POLICY.OBS_TRANSFORMS.RESIZE_SHORTEST_EDGE.SIZE)
+
+
+@baseline_registry.register_obs_transformer()
+class CenterCropper(ObservationTransformer):
+    def __init__(
```
mathfac, line 137: We need doc string.

Skylion007:
I am going to move the OBS_TRANSFORMER out of RL.POLICY to RL. and willl add the doc strings.

Skylion007:
I am still not sure what scope to put the config under, thoughts @erikwijmans I could RL.OBS_TRANSFORMER, but I could also see the arguement for RL.POLICY.OBS_TRANSFORMER. It's general enough to just be at the root node too though.

mathfac (approved):

PR #487, opened 2020-09-23
VQA baseline implementation

mukulkhanna:
## Motivation and Context

This is to implement a VQA baseline in Habitat that will eventually be used as a submodule in the subsequent EmbodiedQA implementation. This model utilizes the previously added EQA CNN pretrained encoder as a feature extractor.

Implementation is based on **EmbodiedQA, Das et al, CVPR 2018** ([paper](https://embodiedqa.org/paper.pdf), [code](https://github.com/facebookresearch/EmbodiedQA))

Relevant issue: #369; More information in [README](https://github.com/mukulkhanna/habitat-api/tree/vqa/habitat_baselines/il#visual-question-answering-vqa-model-).

Note: Since the VQA implementation's data loader has a lot in common with that of the (next to come) PACMAN/NAV implementation, I've intended the data loader to be reusable. 

Thanks to @dhruvbatra and @erikwijmans for all the feedback related to this.

## Types of changes

## Checklist


habitat_baselines/il/metrics.py
```diff
@@ -0,0 +1,114 @@
...
+        for i in range(len(self.metric_names)):
+            if self.metrics[i][mode] is not None:
+                stat_string += "[{}:{:.5f}]".format(
+                    self.metric_names[i],
+                    self.metrics[i][mode],
+                )
+
+        return stat_string
+
+    def get_stats(self, mode=1) -> List:
```
Skylion007, line 81: List[] of what classes? Especially important given the lack of docstrings?

habitat_baselines/il/models/models.py
```diff
@@ -5,6 +5,36 @@
 import torch.nn.functional as F
 
 
+def build_mlp(
```
Skylion007, line 8: typehints would be helpful

habitat_baselines/utils/visualizations/utils.py
```diff
@@ -71,3 +72,94 @@ def save_depth_results(
...
+
+    return image
+
+
+def save_vqa_image_results(
+    images_tensor: torch.Tensor,
+    question: str,
+    prediction: str,
+    ground_truth: str,
+    path,
```
Skylion007, line 134:
```suggestion
    path: str,
```

habitat_baselines/utils/common.py
```diff
@@ -307,3 +307,17 @@ def overwrite_gym_box_shape(box: Box, shape) -> Box:
     low = box.low if np.isscalar(box.low) else np.min(box.low)
     high = box.high if np.isscalar(box.high) else np.max(box.high)
     return Box(low=low, high=high, shape=shape, dtype=box.dtype)
+
+
+def get_scene_episode_dict(episodes: List) -> Dict:
```
Skylion007, line 312: Once again, add the holders to List[]

habitat_baselines/il/trainers/vqa_trainer.py
```diff
@@ -0,0 +1,390 @@
...
+            info={"split": "val"},
+            metric_names=[
+                "loss",
+                "accuracy",
+                "mean_rank",
+                "mean_reciprocal_rank",
+            ],
+            log_json=os.path.join(config.OUTPUT_LOG_DIR, "eval.json"),
+        )
+        with torch.no_grad():
```
Skylion007, line 322: You could decorate the whole function with `@torch.no_grad()` if you want.

habitat_baselines/il/metrics.py
```diff
@@ -0,0 +1,114 @@
...
+
+            self.metrics[i][0] = float(self.metrics[i][0])
+            self.metrics[i][1] = float(self.metrics[i][1])
+            self.metrics[i][2] = float(self.metrics[i][2])
+
+            current_stats.append(self.metrics[i])
+
+        self.stats.append(copy.deepcopy(current_stats))
+
+    def get_stat_string(self, mode=1) -> str:
```
Skylion007, line 64:
```suggestion
    def get_stat_string(self, mode: int =1) -> str: 
```

habitat_baselines/il/metrics.py
```diff
@@ -0,0 +1,114 @@
...
+
+import copy
+import json
+from typing import List, Tuple
+
+import numpy as np
+import torch
+
+
+class Metric:
```
Skylion007, line 15: Shouldn't we subclass the exacting TaskMetrics we already have? If that doesn't work, perhaps we need a new abstractclass Metric type. @mathfac thoughts?

habitat/datasets/utils.py
```diff
@@ -85,6 +85,15 @@ def _build(self):
     def idx2word(self, n_w):
         return self.word_list[n_w]
 
+    def token_idx_2_string(self, tokens):
```
Skylion007, line 88:
```suggestion
    def token_idx_2_string(self, tokens: Iterable[int] ) -> str: 
```
You will need to add the corresponding import of Iterable from typing too.

Skylion007 (changes requested):

Skylion007:
Would be great if we merged in #492 first so that we can validate the typing. ;)

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,255 @@
...
+                    else:
+                        pos_queue = episode.shortest_paths[0]
+
+                    self.save_frame_queue(pos_queue, episode.episode_id)
+
+            logger.info("[ Saved all episodes' frames to disk. ]")
+
+            create_tar_archive(
+                self.frame_dataset_path + ".tar", self.frame_dataset_path
+            )
```
erikwijmans, line 119: I think we should remove the contents of `self.frame_dataset_path` after making the tarball?

erikwijmans (commented):
Some suggestions and one comment

mukulkhanna:
Hi @erikwijmans, @mathfac, any further suggestions/comments? Please let me know, thanks!

Skylion007 (approved):

Skylion007:
@mukulkhanna Resolve the last merge conflict and should be good to go.

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,255 @@
...
+                )
+                == 0
+            ):
+                logger.info("[ Tar archive created. ]")
+                logger.info("[ Deleting dataset folder. ]")
+                delete_folder(self.frame_dataset_path)
+
+            logger.info("[ Frame dataset is ready. ]")
+
+        self.env.close()
```
Skylion007, line 129: Instead of explictly causing close, would make sense to use the with habitat.Env context?

habitat_baselines/utils/common.py
```diff
@@ -343,3 +350,84 @@ def overwrite_gym_box_shape(box: Box, shape) -> Box:
...
+        bytes_obj = BytesIO()
+        bytes_obj.write(img_bytes)
+        image = np.array(Image.open(bytes_obj))
+        img = image.transpose(2, 0, 1)
+        img = img / 255.0
+        images.append(img)
+    return (*x[0:3], np.array(images, dtype=np.float32))
+
+
+def create_tar_archive(archive_path: str, dataset_path: str) -> int:
```
Skylion007, line 415: Nit: I would use the tar package built into Python standard library. Better error handling.

habitat_baselines/il/data/data.py
mukulkhanna: done, thanks!

Skylion007:
Hmm, how did this break this test I wonder.

Skylion007:
Ok, 99% sure its us that just broke something with a recent habitat-sim refactor. That test was not being run due to a bug on our end. Seeing how nothing else broke, going to merge this in if there are no objections @mathfac @erikwijmans

Skylion007:
@erikwijmans @mathfac @jturner65 See https://github.com/facebookresearch/habitat-sim/pull/886 for why our tests didn't catch this. 

Thanks for all the work @mukulkhanna , we'll get it merged in ASAP!

Skylion007:
Okay, pulling from master should fix the test.

Skylion007:
This is great! Now, we really should add some tests at some point in the future. Until then, this looks good to me.

habitat_baselines/utils/common.py
Skylion007: You can sort it by passing in a list of files instead of the source_dir etc...

mukulkhanna: Thanks @Skylion007, I had so far been trying the same snippet you've shared, but was struggling in getting the archived files sorted and grouped by the same basename. Adding a list of files works well - pushing code in a while.

habitat_baselines/utils/common.py
```diff
@@ -414,21 +414,15 @@ def img_bytes_2_np_array(
...
-        "name",
-        "-cf",
-        archive_path,
-        dataset_path,
-    ]
-    process = subprocess.run(cmd)
-    return process.returncode
+    logger.info("[ Creating tar archive. This will take a few minutes. ]")
+
+    with tarfile.open(archive_path, "w:gz") as tar:
```
Skylion007, line 423:
```suggestion
    with tarfile.open(archive_path, "w") as tar:
```
This is going to apply gzip compression, which is not exactly equivalent to the code before: https://docs.python.org/3/library/tarfile.html

mukulkhanna: @Skylion007 WebDataset seems to be able to handle gzip compressed archives too. Plus, the dataset takes less space now. I don't see why this needs to be changed.

Skylion007: Just the extension is tar.gz. Should be good for now I suppose. :)

habitat_baselines/il/metrics.py
mathfac: Yes, we should call it or our metrics class differently. Call out metric class: HabitatMetric or TaskMetric and then this class call BaselinesMetric.

mathfac: Doc string needed.

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,254 @@
...
+
+            group_by_keys = filters.Curried(self.group_by_keys_)
+            super().__init__(
+                urls=self.frame_dataset_path + ".tar",
+                initial_pipeline=[group_by_keys()],
+            )
+
+            self.only_vqa_task = config.ONLY_VQA_TASK
+
+            self.scene_episode_dict = get_scene_episode_dict(self.episodes)
```
mathfac, line 77: `self.scene_episode_dict = get_scene_episode_dict(self.episodes)` => `self.scene_episode_dict = self.env._dataset.scene_ids`

habitat_baselines/utils/common.py
```diff
@@ -345,3 +352,78 @@ def overwrite_gym_box_shape(box: Box, shape) -> Box:
     low = box.low if np.isscalar(box.low) else np.min(box.low)
     high = box.high if np.isscalar(box.high) else np.max(box.high)
     return Box(low=low, high=high, shape=shape, dtype=box.dtype)
+
+
+def get_scene_episode_dict(episodes: List[Episode]) -> Dict:
```
mathfac, line 357: Usually episodes are coming from dataset. This function can be replaces by `scene_ids` property:
https://github.com/facebookresearch/habitat-lab/blob/422a60fb025e9bcaf80bc1b6cf24d48d26e10b41/habitat/core/dataset.py#L131-L134

habitat_baselines/il/data/data.py
```diff
@@ -0,0 +1,254 @@
...
+
+            # [TODO] can be done in mp3d_eqa_dataset when loading
+            self.calc_max_length()
+            self.restructure_ans_vocab()
+
+            group_by_keys = filters.Curried(self.group_by_keys_)
+            super().__init__(
+                urls=self.frame_dataset_path + ".tar",
+                initial_pipeline=[group_by_keys()],
+            )
```
mathfac, line 73: Hard to understand what's happening here. Better self descriptive variables needed or comments.

habitat_baselines/utils/common.py
```diff
@@ -345,3 +352,78 @@ def overwrite_gym_box_shape(box: Box, shape) -> Box:
...
+        if episode.scene_id not in scene_ids:
+            scene_ids.append(episode.scene_id)
+            scene_episode_dict[episode.scene_id] = [episode]
+        else:
+            scene_episode_dict[episode.scene_id].append(episode)
+
+    return scene_episode_dict
+
+
+def base_plus_ext(path: str) -> Union[Tuple[str, str], Tuple[None, None]]:
```
mathfac, line 371: Maybe, `base_plus_ext` => `path_to_base_and_ext`.

habitat_baselines/utils/common.py
```diff
@@ -345,3 +352,78 @@ def overwrite_gym_box_shape(box: Box, shape) -> Box:
...
+    path: path with extensions
+    returns: path with all extensions removed
+    """
+    match = re.match(r"^((?:.*/|)[^.]+)[.]([^/]*)$", path)
+    if not match:
+        return None, None
+    return match.group(1), match.group(2)
+
+
+def valid_sample(sample: dict) -> bool:
```
mathfac, line 383: Doesn't look general enough, possibly better live in IL utils.

habitat_baselines/il/README.md
```diff
@@ -65,4 +65,47 @@ After downloading the pre-trained model, it's path needs to be added to the conf
...
+ python -u habitat_baselines/run.py --exp-config habitat_baselines/config/eqa/il_vqa.yaml --run-type eval
+```
+
+Results from evaluation are stored in `data/eqa/vqa/results/val`.
+
+### Example results:
+
+![](https://user-images.githubusercontent.com/24846546/75141155-464bde00-56e8-11ea-9f2e-ca346440e1d2.jpg)
+![](https://user-images.githubusercontent.com/24846546/75141287-8e6b0080-56e8-11ea-8045-b4c4521954b2.jpg)
+
```
mathfac, line 107: Please, add citation reference required if code used for the paper. Thank you!

mathfac:
@mukulkhanna thank you for the contribution and @Skylion007 for the prompt review. Left some comments, feel free to send follow up PR if you find them reasonable.

mukulkhanna:
Sure, thanks @mathfac!

habitat_baselines/il/README.md
mukulkhanna: I've cited the paper and the code for the whole (EQA) implementation at the top of [`habitat_baselines/il/README.md`](https://github.com/mukulkhanna/habitat-lab/blob/vqa/habitat_baselines/il/README.md).

habitat_baselines/utils/common.py
mukulkhanna: There's no IL utils so far, and only this function to add. I could add it to the `data.py` file, or we could let this be where it is now.  

This is a webdataset related util function, and might be useful for other IL trainers like `eqa_cnn_pretrain` (even for future trainers) if we were to port that to webdataset in the future for uniformity across all baselines (currently it uses an lmdb dataset).

habitat_baselines/utils/common.py
mukulkhanna: The `scene_ids` property can help me get unique scene ids, but `get_scene_episode_dict` provides the scene-episode mapping. 

```
{'scene1': [{episode_1},{episode_2}], 'scene2': [{episode_3},{episode_4}]}
```

Maybe we can add this function to the dataset too?

Skylion007: @mukulkhanna Absolutely if you think it's valuable.

PR #510, opened 2020-10-22
[CIRCLECI] Relax waiting timeout for CUDA install

Skylion007:
Allows a step that waits for CUDA installation to have a more relaxed timeout.

## Motivation and Context

Occasionally the CUDA install is too slow and causes a timeout this should solve it. (We could also have the wait loop print things, but since it works occasionally, I suspect relaxing the timeout will be sufficient.)

## Types of changes

- CI


jturner65 (approved):
LGTM!!

Issue #500, opened 2020-10-16
Add Observation Transformers for Equirec2Perspective


Skylion007:
## 🚀 Feature

 360 sensor -> perspective view observation transformer

I would propose two observation transformers for this:
* Equirect -> perspective imgs (an observation transfomer that can extract multiple perspective imgs from an equirect input)
* A special case of this: Equirect -> Cubemap (allows policies trained on cubemaps to be evaluate on real robots with equirect sensors).

## Motivation
* So one useful feature that would make sim2real comparisons is easier if sampling a perspective image out of a 360 camera sensor (assuming equirect inputs). This would allow easier transfer learning / comparisons from real robots that only come with 360 sensors.

This is useful for transferring policies that were trained on perspective views and having them run on robots which may only have a 360 camera.

Implementations to base ours off of:
* https://github.com/fuenwang/PanoramaUtility/blob/master/Utils/Equirec2Cube.py
* https://github.com/haruishi43/equilib/tree/master/equilib/equi2cube/equi2cube_torch
* https://github.com/haruishi43/equilib/tree/master/equilib/equi2pers/equi2pers_torch

Bonus:
* https://github.com/haruishi43/equilib/tree/master/equilib/equi2equi/equi2equi_torch

Please look at the recent PRs #485 and #486 for examples about how to add these observation transformers.

## Additional context
@matsuren Does this task interest you at all? 

The improvements /additions you made with cubemap2equirect and cubemap2fisheye were greatly appreciated.

matsuren:
Hi @Skylion007, 
It sounds good! I'm a little bit concerned about the spatial resolution when you convert an equirectangular image into a perspective image (You might not be able to see the detailed structures in the perspective image). But if you use high-resolution equirectangular images, it should be okay?

I can try to implement this feature to see how it looks when I have time (I'm busy this week, so hopefully in two or three weeks). Also, I'm interested in the integration of cubemap2equirect, cubemap2fisheye, and equirect2perspective, so I can try this as well if it's okay for you.

Thanks,

Skylion007:
@matsuren Yeah, for reference I imagine this being used on output from commodity 360 sensors like a Ricoh Theta (which easily supports 4k) so that should be more than enough for low resolution perspective images. 👍

matsuren:
Hi @Skylion007, I got it! I will ask you if I have any problems regarding the implementation.
Thanks,

matsuren:
Todo list:
[...]

PR #517, opened 2020-11-01
Implementation of Equirec2Perspective

matsuren:
## Motivation and Context
This is a pull request related to #500, which enables us to convert equirectangular images into perspective images.  Also, I refactored `Cube2Equirec` and `Cube2Fisheye` by implementing the `ProjectionConverter` class that converts images into any camera projection models. 

`ProjectionConverter` class requires input camera projection models and output camera projection model. Each camera projection model has two functions namely `projection` that projects 3D points onto the image plane and `unprojection` that project image points onto the unit sphere. I implemented camera projection models for equirectangular images (`EquirecCamera`), fisheye images (`FisheyeCamera`), and perspective images (`PerspectiveCamera`).

Could you take a look at this pull request, @Skylion007?
I'm also thinking about how to implement `Equirec2Cubemap`. It is possible to implement it by creating six `ProjectionConverter`. Or it might be easier to allow `ProjectionConverter` to have multiple output projection models, but it will complicate the `forward` function.  I would appreciate your help!

## How Has This Been Tested
I tested by executing the following code.
```python
import habitat
from habitat_baselines.common.baseline_registry import baseline_registry
from habitat_baselines.utils.common import batch_obs
import math
from copy import deepcopy
import matplotlib.pyplot as plt
import numpy as np
import torch

# New features
from habitat_baselines.common.obs_transformers import ProjectionConverter
from habitat_baselines.common.obs_transformers import EquirectProjection, FisheyeProjection, PerspectiveProjection, CameraProjection

config=habitat.get_config("configs/tasks/pointnav.yaml")
config.defrost()
CAMERA_NUM = 6
orient = [
    [0, math.pi, 0],  # Back
    [-math.pi / 2, 0, 0],  # Down
    [0, 0, 0],  # Front
    [0, math.pi / 2, 0],  # Right
    [0, 3 / 2 * math.pi, 0],  # Left
    [math.pi / 2, 0, 0],  # Up
]
sensor_uuids = []
# Setup six cameras, Back, Down, Front, Left, Right, Up.
if "RGB_SENSOR" in config.SIMULATOR.AGENT_0.SENSORS:
    config.SIMULATOR.RGB_SENSOR.ORIENTATION = orient[0]
    for camera_id in range(CAMERA_NUM):
        camera_template = f"RGB_{camera_id}"
        camera_config = deepcopy(config.SIMULATOR.RGB_SENSOR)
        camera_config.ORIENTATION = orient[camera_id]

        camera_config.UUID = camera_template.lower()
        sensor_uuids.append(camera_config.UUID)
        setattr(config.SIMULATOR, camera_template, camera_config)
        config.SIMULATOR.AGENT_0.SENSORS.append(camera_template)
config.freeze()

# Setup CubeMap2Equirec
obs_trans = baseline_registry.get_obs_transformer("CubeMap2Equirect")
equirec_size = (480, 960)
cube2equirec = obs_trans(sensor_uuids, equirec_size)

# Load embodied AI task (PointNav) and a pre-specified virtual robot
env = habitat.Env(
    config=config
)

observations = env.reset()

print("Environment creation successful")
observations = env.reset()

batch = batch_obs([observations])

copy_batch = deepcopy(batch)

trans_observations = cube2equirec(copy_batch)
equirect = trans_observations['rgb_0']
plt.title("Generated equirectangular image")
plt.imshow(equirect.numpy().squeeze())
plt.tight_layout()
plt.savefig("equirect.png")

###################################
# Example of ProjectionConverter
###################################
# Input equirectangular camera
input_model = EquirectProjection(*equirec_size)


# Fisheye camera
fish_w, fish_h, fish_fov = 512, 512, 220
xi = 0.3
alpha = 0.3

fx = 0.15 * fish_w
fy = fx
cx = fish_w / 2
cy = fish_h / 2
fish_cam = FisheyeProjection(fish_h, fish_w, fish_fov, cx, cy, fx, fy, xi, alpha)

# Normal camera
perspective_cam = PerspectiveProjection(256, 256)

# Zoom camera
perspective_zoom_cam = PerspectiveProjection(256, 256, f=256)

# Tilted camerea
angle_rad = np.deg2rad(45)
rot = np.array(
    [
        [1.0, 0, 0],
        [0.0, np.cos(angle_rad), -np.sin(angle_rad)],
        [0.0, np.sin(angle_rad), np.cos(angle_rad)],
    ]
)
perspective_tilt = PerspectiveProjection(256, 256, R=torch.from_numpy(rot))


# Define converter functions
to_fisheye = ProjectionConverter(input_model, fish_cam)
to_perspective = ProjectionConverter(input_model, perspective_cam)
to_perspective_zoom = ProjectionConverter(input_model, perspective_zoom_cam)
to_perspective_tilt = ProjectionConverter(input_model, perspective_tilt)

convert_funcs = [to_fisheye, to_perspective, to_perspective_zoom, to_perspective_tilt]

# Convert images
fig, ax = plt.subplots(2, 2, figsize=(10,10))
ax = ax.flatten()

input_tensor = equirect.permute((0, 3, 1, 2))  # NHWC => NCHW
sensor_dtype = input_tensor.dtype
input_tensor = input_tensor.float()


for i, to_ in enumerate(convert_funcs):
    out = to_(input_tensor)
    out = out.permute((0, 2, 3, 1))  # NCHW => NHWC
    out = out.to(dtype=sensor_dtype)
    out_viz = out.numpy().squeeze()
    ax[i].imshow(out_viz)
fig.tight_layout()
fig.savefig("converted_img.png")
```
*I modified this example after renaming {Fisheye, ...}Camera to {Fisheye,...}Projection.

Input equirectangular image:
equirec

Output converted images:
converted_img

## Types of changes

- refactoring
- New feature

## Checklist


matsuren:
Also, I added a bug fix for depth map generation in this pull request.
As explained in #500, currently, six depth maps in cubemap are the distance from each image plane, but it should be the distance from the optical center to create depth map for equirectangular and fisheye.

In my implementation, if input tensors have a single channel, they are considered as depth. So, if you have a plan to use grayscale images, it's better to check dtype as well.

**TODO:**  
- [ ] In the current test, I suppose the depth map generation is not included.  So, I will add them later.
- [ ] Add more comments.

Thanks,

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+    optimized for modern PyTorch."""
+
+    def __init__(self, equ_h: int, equ_w: int):
+        """Args:
+        equ_h: (int) the height of the generated equirect
+        equ_w: (int) the width of the generated equirect
+        """
+
+        # Cubemap input
+        rotations = [
```
Skylion007, line 694: This is a good, nice clean way of doing this! Perhaps these should be constants/enums somewhere though?

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
-            raise ValueError("Batch size should be 6x")
+        # Check whether batch size is len(self.input_models) x
+        if batch_size == 0 or batch_size % input_len != 0:
+            raise ValueError(f"Batch size should be {input_len}x")
 
         # to(device) is a NOOP after the first call
         self.grids = self.grids.to(batch.device)
 
+        # Depth conversion
+        if ch == 1 and self.input_zfactors is not None:
```
Skylion007, line 632: Ah, so one bug here! The semantic sensor could also have a 1 channel output (with uints for object categories), so we need some better way of detecting if it's a depth camera. 
1. We could have it be 1 channel + float. 
2. We could also require by convention that all depth_cameras contain the word "depth" in the UUID. Grr... if we still had the sensor class in scope, would be easy to figure what is a depth sensor. 
3. We could reverse map it, but if a fake_depth ObsTransformer comes first (that generates predicted depth), that wouldn't work either.
4. We could have the user specify it.
5. We could have a separate obs transformer for corrected depth (would inherit with a boolean flag set). Needs to be a distinct name to work around current limitation in config system.
Also we may want to add a logger output under INFO about which channels we are adjusting the depth for.
The ordering of solutions here would be 1 > 4 = 5 > 2 > 3 IMO. Happy to hear your thoughts.

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+            self.output_model.img_w,
         ).sum(dim=1)
-        return output  # batch_size // 6, ch, self.equ_h, self.equ_w
+        return output  # batch_size // input_len, ch, output_model.img_h, output_model.img_w
 
-    # Convert input cubic tensor to output equirectangular image
-    def to_equirec_tensor(self, batch: torch.Tensor) -> torch.Tensor:
-        batch_size = batch.size()[0]
+    def to_converted_tensor(self, batch: torch.Tensor) -> torch.Tensor:
+        batch_size, ch = batch.size()[:2]
```
Skylion007, line 621: This assumes channels first, which I suppose is for this, but there should be a comment in this regard.

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+        sin_theta = torch.sin(theta_map)
+        cos_theta = torch.cos(theta_map)
+        sin_phi = torch.sin(phi_map)
+        cos_phi = torch.cos(phi_map)
+        return torch.stack(
+            [cos_phi * sin_theta, sin_phi, cos_phi * cos_theta], dim=-1
+        )
+
+
+class FisheyeCamera(CameraProjection):
```
Skylion007, line 433: Just for the sake of naming,  does this represent the projection for a single fisheye camera, or a dual fisheye camera. There is some overloading of names here and one might want to construct multiple fisheye cameras (say stereo fisheye cameras even!)

habitat_baselines/common/obs_transformers.py
```diff
@@ -468,29 +852,8 @@ def __init__(
...
-        fish_param: List[float],
-    ) -> torch.Tensor:
-        # Project on sphere
-        xyz_on_sphere, fov_mask = self.get_points_on_sphere(
-            fish_h, fish_w, fish_fov, fish_param
-        )
-
-        # Rotate so that each face will be in front of camera
+        # Cubemap input
         rotations = [
```
Skylion007, line 857: We keep reusing this ordering so maybe it should be a constant/enum somewhere too? Also any reasons its np.arrays instead of a native PyTorch array? It could just be one multidimensional torch.Tensor after all instead of list of np.arrays

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+        """
+        super(ProjectionConverter, self).__init__()
+        self.input_models = input_projections
+        self.output_model = output_projection
+
+        # Check if depth conversion is required
+        # If depth is in z value in input, conversion is required
+        self.input_zfactors = self.calculate_zfactor(self.input_models)
+        # If depth is in z value in output, inverse conversion is required
+        output_invzfactors = self.calculate_zfactor([self.output_model])
```
Skylion007, line 573: This is already assumes a list of output_projects , so why not just make it support the List[CameraProject] as outputs as well. :)

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+        theta_map, phi_map = self.get_theta_phi_map(self.img_h, self.img_w)
+        unproj_pts = self.angle2sphere(theta_map, phi_map)
+        # All points in image are valid
+        valid_mask = torch.full(unproj_pts.shape[:2], True, dtype=torch.bool)
+        # Rotate unproj_pts points according to camera rotation
+        if with_rotation:
+            unproj_pts = self.camcoord2worldcoord(unproj_pts)
+        return unproj_pts, valid_mask
+
+    # Get theta and phi map
```
Skylion007, line 413: Just make these comments into actual doc strings, because why not. That way they should up on autocomplete tools for IDEs

habitat_baselines/common/obs_transformers.py
Skylion007: More importantly, that if statement may want to be factored out into an "is_depth" function to make it clear for future maintainers what it is suppose to be doing.

Skylion007 (changes requested):
> Or it might be easier to allow ProjectionConverter to have multiple output projection models, but it will complicate the forward function. 

Definitely would prefer to have support for multiple outputs! It makes sense really (no reason to reuse / unbatch the computation). Also an unfortunate limitation of our current config system for ObsTransformers that we cannot pass in the same class multiple times with different inputs 
[ (we probably could, we would just need to allow duplicate keys / passing them in directly as part of the YAML config. That is to see instead of having an enabled list with unique keys, having their config values passed directly in and multiple repeated configs represent multiple repeated obs_transformers. That would require some rework of the registry though.]

For most cases, we would actually want to have multiple perspective images from a single equirec (like extracting panorma subviews, or subviews to run pre-trained perspective models on). (That's what I hacked up my code base to use for now, but it's not nearly as clean /reusable as this). In that case, maybe one last standard class we should include is something which splits up the equirect into say 12 perspective slices, just to make sure the depth correction works on that case as well.

I love the camera projection abstraction and the code reuse between classes, very nicely done!

One thing that you brought up actually is that Equirect2Cubemap would need another depth map fix as well! Is that implemented as is? Or does the conversion only work 1 way? Another reason would be to have a special Equirect2Cubemap class. Equirec2Cubemap should also correctly reproject the depth the inverse way as well. (This class would be very useful for any sim2real experiments that use a 360 camera).

Well done though! This is a great, clean, almost complete abstraction. 
TLDR:
* Equirect2Pano / Cubemap are needed to show we can handle depth properly going back and forth (for pretrained classifiers / predictors). 
* To that end, we need to support multiple output projections and converting the depth mapping both ways.

Nice to have:
* That depth map unprojection visualization is great! We should definitely include something like that as a unit test. Do you have a script to generate that? It would be a useful to have. Actually having a tiny tool to test these observations transformer may be a good thing to have. Could in inherit from the existing demos in examples or live as a colab notebook.

Let me know how I can further help!

Aaron

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,496 @@ def from_config(cls, config: Config):
...
+        alpha: float,
+        R: Optional[torch.Tensor] = None,
+    ):
+        """Args:
+        img_h: (int) the height of fisheye camera image
+        img_w: (int) the width of fisheye camera image
+        fish_fov: (float) the fov of fisheye camera in degrees
+        cx, cy: (float) the optical center of the fisheye camera
+        fx, fy, xi, alpha: (float) the fisheye camera model parameters
+        R: (torch.Tensor) 3x3 rotation matrix of camera
```
Skylion007, line 459: Should make a note: in Radians

habitat_baselines/common/obs_transformers.py
Skylion007: Ah nvm, saw your post. Yeah, I would say dtype checking is a must since Semantic Segmentation is one of the default sensors in Habitat and was previously supported. Grayscale would be another common use case.

erikwijmans: IMO, 1 is a hard to find bug waiting to happen.  2 is my preference as requiring that `depth` is in the same of a depth sensor isn't an inconvenient limitation.

habitat_baselines/common/obs_transformers.py
matsuren: Thanks! I didn't think of it. I modified it so that it will appear in IDEs.

habitat_baselines/common/obs_transformers.py
matsuren: I added the order `NCHW` in a comment.

habitat_baselines/common/obs_transformers.py
matsuren: I answered [the below](https://github.com/facebookresearch/habitat-lab/pull/517/files/a066699df29dc4b3bdf00f3bcdd0ba1f9c52ddd5#r515642688) about `constants/enums`.

habitat_baselines/common/obs_transformers.py
matsuren: Hmmm, I'm not sure if I understand correctly. But, maybe I should rename it as `FisheyeProjection`? If someones want `DualFishEye`, they can just use `FisheyeProjection` to define it. 
So, maybe `EquirecProjection` and `PerspectiveProjection` are better naming? What do you think?

Speaking of naming, you use `Equirec` for the abbreviation of equirectangular. I prefer to use `Equirect` since Unity uses `Equirect` for equirectangular [here](https://docs.unity3d.com/ScriptReference/RenderTexture.ConvertToEquirect.html), so it's easy to understand for many people.

habitat_baselines/common/obs_transformers.py
matsuren: You mean `fov_rad = self.fish_fov / 180 * np.pi`? I added a comment.

habitat_baselines/common/obs_transformers.py
matsuren: Sure, I will implement `ProjectionConverter` to allow multiple output projections tomorrow (hopefully).

habitat_baselines/common/obs_transformers.py
matsuren: Give me some time to think about it. Maybe I can try 2 in `ObservationTransformer`, and it passes the `is_depth` flag to `ProjectionConverter` to handle depth correctly.

matsuren (commented):
Thank you for your quick response and valuable comments! It's my first time to receive a professional code review, so let me know if I'm doing something wrong (I'm also not familiar with this GitHub review system as well).

I fixed some issues. But I didn't have time to work on heavy ones (allow multiple output projections, etc.). I will work on it tomorrow (hopefully).

Thanks!

habitat_baselines/common/obs_transformers.py
Skylion007: I actually meant the rotation matrix. In general, angular units should always be defined IMO.

habitat_baselines/common/obs_transformers.py
Skylion007: Sounds good. We are probably going to do a bigger refactor of ObsTransformers at some point anyway so more complete naming is better. Rename it to EquirectProjection and PerspectiveProjection. :)

Agree on your first point. Just wanted to make sure there wasn't term overloading. Leave Fisheye Projection as it is.

habitat_baselines/common/obs_transformers.py
Skylion007: As an extension of 2, we could allow the end user to change the isdepth regex if necessary.

habitat_baselines/common/obs_transformers.py
Skylion007: Shouldn't the RotationMat class at least inherit from 'Enum' then? https://docs.python.org/3/library/enum.html

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,113 +218,545 @@ def from_config(cls, config: Config):
...
-    def angle2sphere(
-        self, theta_map: torch.Tensor, phi_map: torch.Tensor
-    ) -> torch.Tensor:
-        sin_theta = torch.sin(theta_map)
-        cos_theta = torch.cos(theta_map)
-        sin_phi = torch.sin(phi_map)
-        cos_phi = torch.cos(phi_map)
-        return torch.stack(
-            [cos_phi * sin_theta, sin_phi, cos_phi * cos_theta], dim=-1
+class _RotationMat(object):
```
Skylion007, line 710: Maybe add a constructor / static factory method for generating these Tensors from arbitrary Euler angles / Radians as well>

Skylion007: In which case, I guess it makes sense to have it be its own class instead of enum.

habitat_baselines/common/obs_transformers.py
matsuren: Hmm? Is the rotation matrix without units? (neither radians nor degrees)

habitat_baselines/common/obs_transformers.py
matsuren: I started to think that maybe we don't need `RotationMat`.  I agree with you that it's convenient to have methods to rotate matrix from arbitrary Euler angles / Radians, but [Rotaion in scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html) should be enough. 

So, maybe only `get_cubemap_projections` like below is enough? (I couldn't think of the situation where only some rotation matrices out of six rotation matrices are used)
```python
def get_cubemap_projections():
    projections = []
    rotations = [
        torch.tensor([[-1, 0, 0], [0, 1, 0], [0, 0, -1]]),
        torch.tensor([[1, 0, 0], [0, 0, 1], [0, -1, 0]]),
        torch.tensor([[1, 0, 0], [0, 1, 0], [0, 0, 1]]),
        torch.tensor([[0, 0, -1], [0, 1, 0], [1, 0, 0]]),
        torch.tensor([[0, 0, 1], [0, 1, 0], [-1, 0, 0]]),
        torch.tensor([[1, 0, 0], [0, 0, -1], [0, 1, 0]]),
    ]
    for rot in rotations:
        cam = PerspectiveProjection(256, 256, R=rot)
        projections.append(cam)
    return projections
```

matsuren:
Hi @Skylion007, 
now `ProjectionConverter` allows you to use multiple output projection models. I uploaded some examples in [gist](https://gist.github.com/matsuren/30706ca9e1778311d5a799e642c65289). Also, the depth is handled correctly based on UUID in `ObservationTransformer`.

The example using `ObservationTransformer` to deal with both image and depth is in [equirect_obs.py](https://gist.github.com/matsuren/30706ca9e1778311d5a799e642c65289#file-cube2equirect_obs-py) in the gist. I use Open3D to visualize depth as point clouds, which I included in this gist.

I tested the projection conversion from input cubemap to equirect images, and also I converted back from the equirect images to cubemap images. The test code is [habitat_cube2equirect2cube_test.py](https://gist.github.com/matsuren/30706ca9e1778311d5a799e642c65289#file-habitat_cube2equirect2cube_test-py).

Input cubemap:
input_cubemap

Equirect image:
equirect_true

Cubemap from equirect image:
output_cubemap_true

It looks fine, but I noticed the input cubemap and output cubemap have a little difference because of `align_corners = True` in `grid_sample`. Based on geometry,  I think `align_corners` should be `False`, but some borderlines appear with  `align_corners = False`, so I set `align_corners = True` to make the lines disappear.
Could you help me out here? Or let's handle this in another pull request later? (since it's getting messy in this pull request)

with `align_corners = False` (maybe borderlines not so noticeable with this resolution):
equirect_false
output_cubemap_false

Also, one more thing I want to make sure is the depth conversion for cubemap. In the current implementation, depth in the output cubemap is not from optical center but from z value. I think this is a natural way to represent depthmap for perspective cameras.

If it's okay, then I will implement `Equirect2Cubemap` as an example of using multiple output projection models and add tests for `Equirect2Cubemap`.

Thanks!

habitat_baselines/common/obs_transformers.py
```diff
@@ -754,15 +754,15 @@ def __init__(self, equ_h: int, equ_w: int):
...
-        output_projection = EquirecCamera(equ_h, equ_w)
-        super(Cube2Equirec, self).__init__(
+        output_projection = EquirectProjection(equ_h, equ_w)
+        super(Cube2Equirect, self).__init__(
             input_projections, output_projection
         )
 
 
-# TODO Measure Inheritance of CubeMap2Equirec + CubeMap2FishEye into same abstract class
+# TODO Measure Inheritance of CubeMap2Equirect + CubeMap2FishEye into same abstract class
```
Skylion007, line 840: Maybe we should make a ProjectionTransformer here and have all these things subclass it.

habitat_baselines/common/obs_transformers.py
```diff
@@ -633,13 +1024,15 @@ def __init__(
...
         target_uuids: Optional[List[str]] = None,
+        depth_key: str = "depth",
     ):
         r""":param sensor: List of sensor_uuids: Back, Down, Front, Left, Right, Up.
         :param fish_shape: The shape of the fisheye output (height, width)
         :param fish_fov: The FoV of the fisheye output in degrees
         :param fish_params: The camera parameters of fisheye output (f, xi, alpha)
         :param channels_last: Are the channels last in the input
         :param target_uuids: Optional List of which of the sensor_uuids to overwrite
+        :param depth_key: If sensor_uuids has depth_key, they are processed as depth
```
Skylion007, line 1035: Technically, it is if it has a depth key substring.

Skylion007 (changes requested):
>> Also, one more thing I want to make sure is the depth conversion for cubemap. In the current implementation, depth in the output cubemap is not from optical center but from z value. I think this is a natural way to represent depthmap for perspective cameras.

As long as it's a reverse operation to go from 'Equirect -> Cubemap -> Equirect' and 'Cubemap -> Equirect -> Cubemap' sounds good to me.

Thanks for the visualization code, is there anyway we could turn it into a Pytest. Obviously, the easiest Pytest would be to test that the two conversions above are approximately the same (maybe with SSIM?) ^.

> It looks fine, but I noticed the input cubemap and output cubemap have a little difference because of align_corners = True in grid_sample. Based on geometry, I think align_corners should be False, but some borderlines appear with align_corners = False, so I set align_corners = True to make the lines disappear.
Could you help me out here? Or let's handle this in another pull request later? (since it's getting messy in this pull request)

I think this has been brought up here: https://github.com/pytorch/pytorch/issues/20785. Its also a little more complicated because setting it correctly uses an optimized NVIDIA implementation vs a custom PyTorch one (or at least used to). I am not sure if the borderlines are artifacts of the NVIDIA implementation or if there is just something we have to do to fix it (like adding / subtracting 0.5 pixels to the grid_sample functions). Feel free to dive into that thread to fix it. Feel free to fix it in a later PR if it's too much work.

habitat_baselines/common/obs_transformers.py
Skylion007: Yeah, but definitely add some comments about which correspond to which direction and that is should BDFLRU ordering for the cubemap.

Skylion007: Also maybe add a docstring / comment that these can be generated from scipy ordering (maybe even with a small rst code snippit).

matsuren: I added scipy code for generating rotation matrices. I hope this is fine.

matsuren:
I added `ProjectionTransformer`, `Equirect2CubeMap`, and test for the class. I feel like there is another way to test `Equirect2CubeMap`, but I don't understand much about habitat-lab system. please change here if you have better ideas.

> Thanks for the visualization code, is there anyway we could turn it into a Pytest. Obviously, the easiest Pytest would be to test that the two conversions above are approximately the same (maybe with SSIM?) ^.

I hope just comparing values is enough. (I don't wanna include Open3D as the dependencies,,,)
 
I will leave `align_corners = True` here since I couldn't figure out how to fix it. Hopefully, someone will fix this problem later.

I guess this pull request is almost done. Thanks!

test/test_baseline_trainers.py
```diff
@@ -140,10 +134,56 @@ def test_cubemap_stiching(test_cfg_path: str, mode: str, camera: str):
...
+        execute_exp(meta_config, mode)
+        # Deinit processes group
+        if torch.distributed.is_initialized():
+            torch.distributed.destroy_process_group()
+
+    elif camera == "cubemap":
+        # 1) Generate an equirect image from cubemap images.
+        # 2) Generate cubemap images from the equirect image.
+        # 3) Compare the input and output cubemap
+        env = habitat.Env(config=config)
```
Skylion007, line 147: Nit but the Env should be closed. May be good to use a with Env context.

test/test_baseline_trainers.py
```diff
@@ -140,10 +134,56 @@ def test_cubemap_stiching(test_cfg_path: str, mode: str, camera: str):
...
+            output_cube = output_cube.float() / 255
+            input_cube = input_cube.float() / 255
+        output_cube = output_cube.permute((0, 3, 1, 2))  # NHWC => NCHW
+        input_cube = input_cube.permute((0, 3, 1, 2))  # NHWC => NCHW
+        apply_blur = torch.nn.AvgPool2d(5, 3, 2)
+        output_cube = apply_blur(output_cube)
+        input_cube = apply_blur(input_cube)
+
+        # Calculate the difference
+        diff = torch.abs(output_cube - input_cube)
```
Skylion007, line 183: This is fine I suppose, but L1 isn't the best distance. Maybe best to compare with SSIM? Otherwise, this diff looks pretty good.

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,166 +218,676 @@ def from_config(cls, config: Config):
...
+        input_projections = get_cubemap_projections()
+
+        # Equirectangular output
+        output_projection = EquirectProjection(equ_h, equ_w)
+        super(Cube2Equirect, self).__init__(
+            input_projections, output_projection
+        )
+
+
+class ProjectionTransformer(ObservationTransformer):
```
Skylion007, line 845: Nice!

Skylion007 (changes requested):
Just make the small changes to the test case and this looks good. BTW, if you are using the bleeding master, we just made a breaking API change on the interface between Habitat-Sim and Habitat-Lab so make sure you have both updated locally if you merge from Master. :)

habitat_baselines/common/obs_transformers.py
```diff
@@ -217,166 +218,676 @@ def from_config(cls, config: Config):
...
             batch,
             self._grids_cache,
             align_corners=True,
             padding_mode="zeros",
         )
         output = output.view(
-            batch_size // 6, 6, ch, self.equ_h, self.equ_w
-        ).sum(dim=1)
-        return output  # batch_size // 6, ch, self.equ_h, self.equ_w
+            self.input_len,
```
Skylion007, line 679: Hold on, does this still work? Was this tested on out output with VectorEnv? Want to make sure swapping the view ordering doesn't break anything. In this case of batch size 1, this should work (1/6 vs 6/1, but want to make sure it still works on higher batch sizes.)

test/test_baseline_trainers.py
Skylion007: Can we actually test a habitat_env wrapped in a vector env? I just want to make sure it doesn't break with a batch size more than 1. :)

Skylion007: You can also use the same input and just batch two separate observations (from two separate positions, I am worried about the cubemap values being mixed up potentially and we should test explicitly for that.

test/test_baseline_trainers.py
matsuren: Sure, I can use SSIM.

habitat_baselines/common/obs_transformers.py
matsuren: I didn't aware of `VectorEnv`. Let me check if it works.

matsuren: I see, I totally misunderstood the order, If there are two batches from two envs `(R_1st, G_1st, B_1st)` and `(R_2nd, G_2nd, B_2nd)`, the order should be `[R_1st, R_2nd, G_1st, G_2nd, B_1st, B_2nd]` not that `[R_1st, G_1st, B_1st, R_2nd, G_2nd, B_2nd]`. I will fix it.

matsuren: Oh I see, I need to fix the order for output,,, Currently, the order is `[R_1st, G_1st, B_1st, R_2nd, G_2nd, B_2nd]`.

test/test_baseline_trainers.py
matsuren: Hold on. Are you sure you want to use SSIM? Currently, there is no dependency on `scikit-image` in habitat-lab. But I wanna use it for [SSIM](https://scikit-image.org/docs/dev/auto_examples/transform/plot_ssim.html).
I'm not sure if it is a good idea to include `scikit-image` just for testing. Also, only the difference caused by conversion should be some blur and a little misalignment because of `align_corners=True`, so I think L1 after downsampling images is enough in this case.

habitat_baselines/common/obs_transformers.py
matsuren: I fixed it. I think it works now.

test/test_baseline_trainers.py
matsuren: Thank you. I added `VectorEnv` in testing.

matsuren:
Thank you for reviewing my source code!
I fixed the order to get it worked with `VectorEnv`. Also, I checked some examples with the latest Habitat-Sim and Habitat-Lab. Thanks!

test/test_baseline_trainers.py
Skylion007: Wow, does OpenCV really not have any easily accessible Python SSIM bindings??
Neither does PyTorch??
Tensorflow does, but that's overkill. 

I guess L1 is fine in this case then.

matsuren:
Thank you for taking care of it!

PR #519, opened 2020-11-04
Add warning for actions that are CUDA tensors

erikwijmans:
## Motivation and Context

Issue that I helped a user debug where they were getting lots of CUDA contexts.  It turns out they were unknowingly sending actions in as CUDA tensors.  This PR adds a warning on that case and adds comments to why actions are getting moved to the CPU.

## How Has This Been Tested

Locally.

## Types of changes

### DIFF
--- a/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -201,7 +201,7 @@ class PPOTrainer(BaseRLTrainer):
         # sent in to env.step(), that will create CUDA contexts
         # in the subprocesses.
         # For backwards compatibility, we also call .item() to convert to
-        # and int
+        # an int
         step_data = [a.item() for a in actions.to(device="cpu")]
         profiling_wrapper.range_pop()  # compute actions
 
@@ -586,7 +586,7 @@ class PPOTrainer(BaseRLTrainer):
             # sent in to env.step(), that will create CUDA contexts
             # in the subprocesses.
             # For backwards compatibility, we also call .item() to convert to
-            # and int
+            # an int
             step_data = [a.item() for a in actions.to(device="cpu")]
 
             outputs = self.envs.step(step_data)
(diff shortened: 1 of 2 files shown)


Skylion007 (approved):
YES YES YES! It's good to add more warnings like this to Habitat-Lab to prevent bad CUDA ctx memory allocations.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -578,7 +582,14 @@ def _eval_checkpoint(
 
                 prev_actions.copy_(actions)  # type: ignore
 
-            outputs = self.envs.step([a[0].item() for a in actions])
+            # NB: Move actions to CPU.  If CUDA tensors are
+            # sent in to env.step(), that will create CUDA contexts
+            # in the subprocesses.
+            # For backwards compatibility, we also call .item() to convert to
+            # and int
```
Skylion007, line 589: typo: an int

Also, technically according to the typing in hablab, it the action doesn't HAVE to be an int does it? It can be a string or anything that corresponds to proper agent config? 

If we enforce it has to be an int, we should adopt the typing in habitat-sim

erikwijmans: It doesn't have to be an int anymore.  That is just for backwards compatibility (it did once upon a time)

Skylion007: Maybe update the comment to reflect that? I think our typing for Simulator in Habitat.core might also need to be updated. :)

Skylion007:
@erikwijmans Maybe we should use the PyTest instafail plugin to get more meaningful and timely output on CircleCI. :)

erikwijmans:
This is just because I forgot to merge with master after the `scene_id` PRs landed.

Skylion007:
LGTM, let's merge. 👍

PR #527, opened 2020-11-10
Enable MYPY for tests and fix typing bugs

Skylion007:
## Motivation and Context

Caught quite a few typing bugs by using mypy and typeguard on the test suite. I was considering enabling TypeGuard for Habitat-Lab, but it seemed a bit to strict and I couldn't remove all the cyclic dependencies easily. 

This does fix quite a few typing bugs though. The other option is that we could keep typeguard, but only enable it for the habitat-sim package.

I also enabled a few new mypy options that caught quite a few typing bugs especially with ObservationTransformers.

Improved testing

## How Has This Been Tested
Pytest

## Types of changes

## Checklist


erikwijmans (approved):

habitat/core/simulator.py
```diff
@@ -6,27 +6,16 @@
...
-    List,
-    Optional,
-    Sequence,
-    Union,
-)
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Union
 
 import attr
+import numpy as np
+import torch
```
mathfac, line 13: @Skylion007, this is the only line dependent on PyTorch in Hab-Lab core. Is there a way we can guard it under TYPE_CHECKING?

Skylion007: Yes.

PR #533, opened 2020-11-20
Orthographic Camera Support

Skylion007:
## Motivation and Context

* It enables orthographic camera support in Habitat-Sim from Habitat-Lab. Adds a new field to Visual Sensors for `SENSOR_SUBTYPE`. Also working on adding a demo to examples/example_visualization.

Need feedback on the API. There is currently no way to set the default zoom in the sensor spec for Habitat_Sim. This may need to be changed.

## How Has This Been Tested
CI and locally

## Types of changes

## Checklist

### DIFF
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -227,8 +227,6 @@ SIMULATOR_SENSOR.HFOV = 90  # horizontal field of view in degrees
 SIMULATOR_SENSOR.POSITION = [0, 1.25, 0]
 SIMULATOR_SENSOR.ORIENTATION = [0.0, 0.0, 0.0]  # Euler's angles
 SIMULATOR_SENSOR.SENSOR_SUBTYPE = "PINHOLE"
-SIMULATOR_SENSOR.ZOOM = 1.0
-
 # -----------------------------------------------------------------------------
 # RGB SENSOR
 # -----------------------------------------------------------------------------
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -226,6 +226,8 @@ SIMULATOR_SENSOR.WIDTH = 640
 SIMULATOR_SENSOR.HFOV = 90  # horizontal field of view in degrees
 SIMULATOR_SENSOR.POSITION = [0, 1.25, 0]
 SIMULATOR_SENSOR.ORIENTATION = [0.0, 0.0, 0.0]  # Euler's angles
+SIMULATOR_SENSOR.SENSOR_SUB_TYPE = "PINHOLE"
+
 # -----------------------------------------------------------------------------
 # RGB SENSOR
 # -----------------------------------------------------------------------------
--- a/habitat/config/default.py
+++ b/habitat/config/default.py
@@ -226,7 +226,8 @@ SIMULATOR_SENSOR.WIDTH = 640
 SIMULATOR_SENSOR.HFOV = 90  # horizontal field of view in degrees
 SIMULATOR_SENSOR.POSITION = [0, 1.25, 0]
 SIMULATOR_SENSOR.ORIENTATION = [0.0, 0.0, 0.0]  # Euler's angles
-SIMULATOR_SENSOR.SENSOR_SUB_TYPE = "PINHOLE"
+SIMULATOR_SENSOR.SENSOR_SUBTYPE = "PINHOLE"
+SIMULATOR_SENSOR.ZOOM = 1.0
 
 # -----------------------------------------------------------------------------
 # RGB SENSOR
--- a/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -22,6 +22,7 @@ from numpy import ndarray
 
 import habitat_sim
 from habitat.core.dataset import Episode
+from habitat.core.logging import logger
 from habitat.core.registry import registry
 from habitat.core.simulator import (
     AgentState,
@@ -66,10 +67,13 @@ def overwrite_config(
             return config
 
     for attr, value in config_from.items():
-        if attr.lower() not in ignore_keys and hasattr(
-            config_to, attr.lower()
-        ):
-            setattr(config_to, attr.lower(), if_config_to_lower(value))
+        low_attr = attr.lower()
+        if low_attr not in ignore_keys and hasattr(config_to, low_attr):
+            setattr(config_to, low_attr, if_config_to_lower(value))
+        else:
+            logger.warn(
+                f"{low_attr} is not found on target {config_to} but is found on {config_from}. Did you make a typo in the YAML config?"
+            )
 
 
 @registry.register_sensor
--- a/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -3,7 +3,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -11,6 +10,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Set,
     Union,
     cast,
 )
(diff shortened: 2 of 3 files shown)


habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -209,6 +220,19 @@ def __init__(self, config: Config) -> None:
             len(self.sim_config.agents[0].action_space)
         )
         self._prev_sim_obs: Optional[Observations] = None
+        for agent_id, _ in enumerate(self.habitat_config.AGENTS):
```
Skylion007, line 223: @jturner65 I am pretty sure it needs to be called during reconfigure as well, but when are the objects deleted from the scene? I think we should merge the concept of object / scene state management from Sokoban to the basic sim class @dexter1691

habitat/config/default.py
```diff
@@ -226,6 +226,7 @@ def __init__(self, *args, **kwargs):
 SIMULATOR_SENSOR.HFOV = 90  # horizontal field of view in degrees
 SIMULATOR_SENSOR.POSITION = [0, 1.25, 0]
 SIMULATOR_SENSOR.ORIENTATION = [0.0, 0.0, 0.0]  # Euler's angles
+SIMULATOR_SENSOR.SENSOR_SUBTYPE = "PINHOLE"
```
jturner65, line 229: This is an enum on the sim side.  Is this getting remapped?

jturner65 (commented):
Question about the SENSOR_SUBTYPE

habitat/config/default.py
Skylion007: Yeah, you can't represent enums in YAML files easily.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -61,9 +65,16 @@ def if_config_to_lower(config):
...
     for attr, value in config_from.items():
-        if hasattr(config_to, attr.lower()):
-            setattr(config_to, attr.lower(), if_config_to_lower(value))
+        low_attr = attr.lower()
+        if has_ignores or low_attr not in ignore_keys:
+            if hasattr(config_to, low_attr):
+                setattr(config_to, low_attr, if_config_to_lower(value))
+            else:
+                logger.warn(
+                    f"{low_attr} is not found on target {config_to} but is found on config_from. Did you make a typo in the YAML config?: {config_from}"
```
jturner65, line 76: Should this not be 
`found on {config_from}`
or else 
`found in source config`
?

jturner65 (approved):
This looks good to me.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -283,6 +333,21 @@ def _update_agents_state(self) -> bool:
 
         return is_updated
 
+    def _load_agent_body_meshes(self) -> None:
```
Skylion007, line 336: We really need to think about this when we start managing object states in episodes. @jturner65 Is there not a cleaner way to load the agent_body_mesh_config? I don't want to load it multiple times if its unnecessary.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -192,74 +203,105 @@ class HabitatSim(habitat_sim.Simulator, Simulator):
...
+            for sensor in sensor_suite.sensors.values():
+                sim_sensor_cfg = habitat_sim.SensorSpec()
+                sim_sensor_cfg.sensor_subtype = getattr(
+                    habitat_sim.SensorSubType,
+                    sensor.config.SENSOR_SUBTYPE.upper(),
+                )
+                overwrite_config(
+                    config_from=sensor.config,
+                    config_to=sim_sensor_cfg,
+                    ignore_keys={
```
Skylion007, line 252: I really want to have warnings about unsupported properties, but I am wondering if there is a way to keep this list form growing too much.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -192,74 +203,105 @@ class HabitatSim(habitat_sim.Simulator, Simulator):
...
-            # accessing child attributes through parent interface
-            # We know that the Sensor has to be one of these Sensors
-            sensor = cast(HabitatSimVizSensors, sensor)
-            sim_sensor_cfg.sensor_type = sensor.sim_sensor_type
-            sim_sensor_cfg.gpu2gpu_transfer = (
-                self.habitat_config.HABITAT_SIM_V0.GPU_GPU
+            sensor_specifications = []
+            for sensor in sensor_suite.sensors.values():
+                sim_sensor_cfg = habitat_sim.SensorSpec()
+                sim_sensor_cfg.sensor_subtype = getattr(
```
Skylion007, line 245: Is this really the best place to convert the STR to an enum?

habitat/sims/habitat_simulator/habitat_simulator.py
jturner65: This could certainly be made quite a bit simpler - all of this could be bundled in simulator.cpp on sim side and could be accessed by a single function call; probably would be useful to have agent configs be supported in the scene instance json.

habitat/sims/habitat_simulator/habitat_simulator.py
jturner65: Could you have a HabitatSim instance (default) dictionary that ties strings to various hab-sim side quantities, such as the enum bindings?

Skylion007: That's still something we have to maintain though. Ideally the conversion could be implicit.

habitat/sims/habitat_simulator/habitat_simulator.py
Skylion007: I really disagree with the later point. Agent configs definitely belong as part of the task definition not the scene config.

jturner65: Fair enough.  At the very least, the attributes for the objects being used to represent the various agents could be included in the Scene Dataset Library configs.  That way the agent configuration object attributes would not need to be specifically loaded when assigning an object to the agent.

habitat/sims/habitat_simulator/habitat_simulator.py
jturner65: I use a few string-keyed, enum-valued const maps in sim code to manage some attributes mappings [(like here)](https://github.com/facebookresearch/habitat-sim/blob/10829a7f6ee660dd001c22a9e4c2089b3d41ddb2/src/esp/metadata/attributes/ObjectAttributes.h#L29).  These could be added to all enums, and if they were then bound to python, they could provide the conversion.  They would still require support of course, but at least it would be restricted to just sim support, which would be necessary anyway if the enum values changed.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -192,74 +203,105 @@ class HabitatSim(habitat_sim.Simulator, Simulator):
...
-        for sensor in _sensor_suite.sensors.values():
-            sim_sensor_cfg = habitat_sim.SensorSpec()
+        agent_configs = []
+        for agent_id, sensor_suite in self.__sensor_suites.items():
+            agent_config = habitat_sim.AgentConfiguration()
             overwrite_config(
-                config_from=sensor.config, config_to=sim_sensor_cfg
+                config_from=self._get_agent_config(agent_id),
+                config_to=agent_config,
+                ignore_keys={"body_mesh_config"},
```
mathfac, line 240: Looks like `ignore_keys` hack mixes Simulator Agent config and Habitat Lab layer of agent specification. While `body_mesh_config` isn't supported as Agent's config part let's have it separately from HSIM config.

habitat/sims/habitat_simulator/habitat_simulator.py
mathfac: all this keys aren't in HSIM spec?

Skylion007: Nope or if they are they are converted in the loop below.

habitat/sims/habitat_simulator/habitat_simulator.py
Skylion007: I don't follow what you mean? The config is part of the AGENT_0 YAML subconfig.

mathfac:
@Skylion007, does this PR still makes sense? What about merge it in several days before we do config refactoring.

PR #535, opened 2020-11-24
Fix more incorrect typehints in Habitat Core

Skylion007:
## Motivation and Context
I verified a bunch of the typehints using typeguard and caught numerous issues in habitat.core . I went ahead fixed most of these issues. I also enabled including habitat_sim in the mypy type checker so we get any errors at the interface between habitat-lab and habitat-sim.

## How Has This Been Tested
Typeguard + MYPY

## Types of changes

## Checklist

### DIFF
--- a/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -4,21 +4,13 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    Optional,
-    Sequence,
-    Union,
-    cast,
-)
+from typing import Any, Dict, List, Optional, Sequence, Union, cast
 
 import numpy as np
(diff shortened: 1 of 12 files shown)


habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -4,21 +4,13 @@
...
-    Union,
-    cast,
-)
+from typing import Any, Dict, List, Optional, Sequence, Union, cast
 
 import numpy as np
 from gym import spaces
 from gym.spaces.box import Box
 from numpy import ndarray
+from torch import Tensor
```
erikwijmans, line 13: When did torch get added as a hard dependency for habitat?

Skylion007: It is for CUDA enabled and noise models. I can soften.

erikwijmans: Yeah, soften would be good.  Don't want people thinking they have to PyTorch as their ML framework to use Habitat :)

erikwijmans (approved):

PR #538, opened 2020-12-02
Collapse PPO and DD-PPO trainers, faster RNN code, and double buffered sampling

erikwijmans:
## Motivation and Context

This contains most of my local improvements to habitat-baselines.  I also collapsed the PPO and DD-PPO trainers as those don't really have a reason to be separate besides history.

Also made some changes to get test time lower.  Working on this was a PITA otherwise.

### Change-log

- Collapse PPO and DD-PPO trainers
- Faster RNN code -- it is definitely faster and can make a noticeable impact during early training (~20% faster in some cases), but good luck reading it :-)
- Rename `NUM_PROCESSES` to `NUM_ENVIRONMENTS`.  The fact that the simulators are in different processes is an implementation detail.  A backwards compatibility check has been added tho.
- Support specifying training length both in terms of number of updates and number of frames
- Specify the number of checkpoints as the number of checkpoints instead of a checkpoint interval
- Introduce a TensorDict class for more cleanly interacting with dictionaries of tensors (potentially recursive).  This also makes `RolloutStorage` about 100x cleaner.
- Store RGB observations as their proper dtype in the rollout storage (this can save a lot of memory)
- Some refactoring of PPOTrainer.train to be less of a script wrapped in a function
- Double buffered sampling.  This can improve performance when simulation time is equal or larger than policy inference time

A note for reviewing:  There are 3 blocks to this that are (mostly) independent conceptually.  The PPOTrainer refactor + double buffered sampling, TensorDict + RolloutStorage refactor, and the RNN code.

To come in a different PR:
- torch.cuda.amp support
- FP16 inference and mixed-precision training support (aka apex.amp O2 mode without apex because it is kinda a pain)

## How Has This Been Tested

Existing tests and new test for the RNN code.
## Types of changes


habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,170 @@
...
+def _to_tensor(v: TensorLike) -> torch.Tensor:
+    if torch.is_tensor(v):
+        return v
+    elif isinstance(v, np.ndarray):
+        return torch.from_numpy(v)
+    else:
+        return torch.tensor(v, dtype=torch.float)
+
+
+class TensorDict(dict):
```
Skylion007, line 28: So extending dict lie this is usually not a great idea. Is there any reason ModuleDict, ParamterDict or https://pypi.org/project/dotmap/ or https://github.com/cdgriffith/Box/tree/master/box or something else? There are a lot of subtleties that are very easy to get wrong with Dot Dictionaries.

The set functions should have an explicit None return type btw.

habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,170 @@
...
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import copy
+from typing import Callable, Dict, Optional, Tuple, Union, overload
+
+import numpy as np
+import torch
+
+TensorLike = Union[
```
Skylion007, line 12: Typing can be simplified a bit with numbers.Real . Also we don't support integer based arrays for numpy (np.int32, np.int64, np.uint8)

habitat_baselines/common/tensor_dict.py
Skylion007, line 28: My want to also extend `Dict[str, Union['TensorDict', Tensor']]`

Skylion007: You could also use TypeVars to allow further generics for FloatTensors etc... https://stackoverflow.com/questions/61112684/how-to-subclass-a-dictionary-so-it-supports-generic-type-hints

habitat_baselines/config/default.py
```diff
@@ -193,5 +200,14 @@ def get_config(
         config.CMD_TRAILING_OPTS = config.CMD_TRAILING_OPTS + opts
         config.merge_from_list(config.CMD_TRAILING_OPTS)
 
+    if config.NUM_PROCESSES != -1:
```
Skylion007, line 203: Use YACS.register_deprecated_keys() instead to handle this for us.

habitat_baselines/config/pointnav/ppo_pointnav_example.yaml
```diff
@@ -12,25 +12,26 @@ VIDEO_DIR: "video_dir"
 # To evaluate on all episodes, set this to -1
 TEST_EPISODE_COUNT: 2
 EVAL_CKPT_PATH_DIR: "data/new_checkpoints"
-NUM_PROCESSES: 1
+NUM_SIMULATORS: 1
 SENSORS: ["RGB_SENSOR", "DEPTH_SENSOR"]
```
Skylion007, line 16: While we are at it, why do we override SENSORS on the DDPPO level. It's a bad idea as the sensorspec really should be defined by the task and this overrides those values.

Skylion007 (changes requested):
NIT: For backwards compatibility, is there anyway we can have a shim for NUM_PROCESSES -> NUM_SIMULATORS. We have the `register_renamed_key` decorator from YACS we could use / adapt as well as the `register_deprecated_key` to help people upgrade their configs as well. I like the TensorDict, but any reason we aren't using one of the  more sophesticated dotmap dict. 

You should also extend Dict[str, ....] so that all inherited methods will have proper typing.

One last thing that is bother is that the DDPPO_CONFIG overwrites SENSORS. No reason for it to this. While we are making breaking changes can we get rid of / at the very least deprecate this too?

habitat_baselines/config/pointnav/ppo_pointnav_example.yaml
erikwijmans: I agree.  This override is done because there isn't a way to change the habitat config from habitat baseline's config (nor is there a good way to do so with the way the configs are structured).

habitat_baselines/common/tensor_dict.py
erikwijmans: I think we do support integer based arrays.  The RGB sensor returns `np.uint8`

habitat_baselines/config/default.py
erikwijmans: This won't work because we are allowing new keys and YACS checks that before checking deprecation.

habitat_baselines/config/pointnav/ppo_pointnav_example.yaml
Skylion007: There is a way to do with command line arguments, which I imagine would be the most common use case in practice.

.circleci/config.yml
```diff
@@ -245,7 +245,7 @@ jobs:
               . activate habitat; cd habitat-lab
               python setup.py develop --all
               export PYTHONPATH=.:$PYTHONPATH
-              python setup.py test
+              python setup.py test --addopts "--cov-report=xml --cov=./"
```
erikwijmans, line 248: @Skylion007 Running codecov locally was really fighting me today so I made it just run on the CI.  I assume there was no reason it was running locally before?

Skylion007: No, not really

Skylion007:
Also to clarify nothing wrong with dotdict, just curious why you didn't use a modern / typed dotdict library to do it properly.

habitat_baselines/config/pointnav/ppo_pointnav_example.yaml
erikwijmans: Yes and no.  I personally like having it in the config and the command line way is unintuitive because you have the TASK_CONFIG thing

erikwijmans:
> Also to clarify nothing wrong with dotdict, just curious why you didn't use a modern / typed dotdict library to do it properly.

I didn't want to add a dependency and it doesn't matter that much.

habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,165 @@
...
+
+import numpy as np
+import torch
+
+TensorLike = Union[torch.Tensor, np.ndarray, numbers.Real]
+DictTree = Dict[str, Union[TensorLike, "DictTree"]]
+TensorIndexType = Union[int, slice, Tuple[Union[int, slice], ...]]
+
+
+def _to_tensor(v: TensorLike) -> torch.Tensor:
```
Skylion007, line 18: Seems unnecessary. How is this different from [torch.as_tensor](https://pytorch.org/docs/stable/generated/torch.as_tensor.html)

erikwijmans: Oh yeah, forgot about that.  Thanks!

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,193 @@
...
-
-                actions_batch.append(self.actions[: self.step, ind])
-                prev_actions_batch.append(self.prev_actions[: self.step, ind])
-                value_preds_batch.append(self.value_preds[: self.step, ind])
-                return_batch.append(self.returns[: self.step, ind])
-                masks_batch.append(self.masks[: self.step, ind])
-                old_action_log_probs_batch.append(
-                    self.action_log_probs[: self.step, ind]
+        if num_simulators % num_mini_batch != 0:
+            warnings.warn(
```
Skylion007, line 184: Why do we using warnings instead of using logger.warn?

habitat_baselines/config/default.py
```diff
@@ -193,5 +203,14 @@ def get_config(
...
+        warnings.warn(
```
Skylion007, line 207: Once again, any reason we aren't using the habitat logger here?

setup.cfg
```diff
@@ -18,17 +18,19 @@ max-line-length = 88
...
+# SIM106 has too many false positives
+# SIM113 has too many false positives
 ignore =
 	A003,
 	C401,C402,C408,
 	SIM105,SIM106
 	R504,
-	W503
+	W503,
+	SIM113
```
Skylion007, line 29: Please put all the SIM ignores on the same line

habitat_baselines/common/rollout_storage.py
erikwijmans: This way it only shows up once.

habitat_baselines/config/default.py
erikwijmans: Same reason

habitat_baselines/common/rollout_storage.py
Skylion007: Does this PR also fix the config printing so all the config printers aren't woven together for all the processes making reading the config from DDPPO really difficult?

habitat/core/vector_env.py
```diff
@@ -75,6 +76,30 @@ def _make_env_fn(
...
+    write_fn: Callable[[Any], None]
+    _read_wrapper: ReadWrapper
+
+    def __call__(self, data: Any) -> None:
+        assert not self._read_wrapper._is_waiting
+        self.write_fn(data)
+        self._read_wrapper._is_waiting = True
+
+
 class VectorEnv:
```
Skylion007, line 103: As an aside, shouldn't VectorEnv by sub-classing the Gym VectorEnv?

Skylion007: Or at the very leasr Habitat.Env or something?

habitat_baselines/common/rollout_storage.py
erikwijmans: That would require not printing the config in `__init__` as we need to have the distributed group initialized to know ranks, and really just rank0 should do the print once.  I can make that change.  My solution is to just not print the config.

habitat/core/vector_env.py
erikwijmans: Seems odd to have the vector version inherit from the single env version given how different the APIs are.  I don't think the vector env passing `isinstance(env, habitat.Env)` makes sense.

Skylion007: Fair enough, but it should inherit from the GymVectorEnv, right?

Skylion007:
Also NIT: are we going to start having the commonly used modules be ScriptModules for the added speed any time soon? That seems like another great speed optimization.

erikwijmans:
> Also NIT: are we going to start having the commonly used modules be ScriptModules for the added speed any time soon? That seems like another great speed optimization.

No, PyTorch JIT is way too unstable.  Requires a very different song and dance to get working correctly for each version.

habitat/core/vector_env.py
erikwijmans: We have a slightly different API, we return a list of each environment's output instead of `obs, rewards, dones, info` to not deeply enforce that pattern (same reason why `habitat.Env` does not inherit from `gym.Env`).

mathfac: Correct, Habitat Lab core was designed to be open to different approaches. That's the reason we don't provide reward as part of the task specification.

Skylion007 (approved):

examples/tutorials/colabs/Habitat_Interactive_Tasks.ipynb
```diff
@@ -1799,9 +1795,8 @@
     "\n",
     "        self._hidden_size = hidden_size\n",
     "\n",
-    "        self.state_encoder = RNNStateEncoder(\n",
-    "            2 * self._n_input_goal,\n",
-    "            self._hidden_size,\n",
+    "        self.state_encoder = build_rnn_state_encoder(\n",
```
mathfac, line 1798: How you tested this notebook? CI or locally?

habitat/core/vector_env.py
```diff
@@ -75,6 +76,30 @@ def _make_env_fn(
     return habitat_env
 
 
+@attr.s(auto_attribs=True, slots=True)
+class ReadWrapper:
```
mathfac, line 80: For such general wrappers doc strings will be useful.

habitat/core/vector_env.py
```diff
@@ -294,63 +317,57 @@ def _spawn_workers(
             ps.daemon = True
             ps.start()
             worker_conn.close()
-        return (
-            [p.recv for p in parent_connections],
-            [p.send for p in parent_connections],
-        )
+
+        read_fns = [ReadWrapper(p.recv) for p in parent_connections]
```
mathfac, line 321: As I understand, you made blocking working process wise. That means you can send another command to the fastest process while others are still calculating the first command.

habitat/core/vector_env.py
erikwijmans: Happy to add doc strings, but these aren't really general purpose, they are just for the vector env implementation

habitat/core/vector_env.py
erikwijmans: Yep, this just makes the tracking of whether or not a process is doing work or is ready for me done per process instead of all processes.  The `_is_waiting` is also used by the write wrapper and read wrapper to make sure the write/read is valid and won't cause otherwise hard to debug bugs.

mathfac: I see. That's really nice.

habitat_baselines/common/base_trainer.py
```diff
@@ -150,13 +150,58 @@ class BaseRLTrainer(BaseTrainer):
...
+        if self.config.NUM_UPDATES != -1:
+            return self.num_updates_done / self.config.NUM_UPDATES
+        else:
+            return self.num_steps_done / self.config.TOTAL_NUM_STEPS
+
+    def is_done(self) -> bool:
+        return self.percent_done() >= 1.0
+
+    def should_checkpoint(self) -> bool:
+        checkpoint_every = 1 / self.config.NUM_CHECKPOINTS
```
mathfac, line 193: Can you elaborate more why you decided to use `NUM_CHECKPOINTS` instead of checkpoint every NUM updates? Thank you!
One use case, when it will require some math to be done is when user want to run "endless" training and stop it when the training converged. Then very large number used for NUM_UPDATES/TOTAL_NUM_STEPS. Those will required to calculate NUM_CHECKPOINTS.

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
+        assert (self._num_envs % self._nbuffers) == 0
 
-        self.masks = torch.zeros(num_steps + 1, num_envs, 1)
+        self.numsteps = numsteps
+        self.steps = [0 for _ in range(self._nbuffers)]
 
-        self.num_steps = num_steps
-        self.step = 0
+    @property
+    def step(self):
```
mathfac, line 86: There can be some confusion between methods `step` we are using for Env/Sim and property here. Maybe, we can use this moment to rename it to `step_id` or etc.

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
     def __init__(
         self,
-        num_steps,
+        numsteps,
         num_envs,
         observation_space,
         action_space,
         recurrent_hidden_state_size,
         num_recurrent_layers=1,
+        is_double_buffered: bool = False,
```
mathfac, line 26: Let's add some doc string about double buffering. That's not the easiest abstraction to get from the code without context.

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
-        for sensor in observations:
-            self.observations[sensor][self.step + 1].copy_(
-                observations[sensor]
-            )
-        self.recurrent_hidden_states[self.step + 1].copy_(
-            recurrent_hidden_states
+        if not self.is_double_buffered:
+            assert buffer_index == 0
+
+        next_step = dict(
```
mathfac, line 107: Maybe, that worth to divide next vs current step parameters using prefix as well `next_observations`, ....

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
+            )
+
+        if len(current_step) > 0:
+            self.buffers.set(
+                (self.steps[buffer_index], env_slice),
+                current_step,
+                strict=False,
+            )
+
+    def advance_rollout(self, buffer_index: int = 0):
```
mathfac, line 143: Method is not used.

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
         if use_gae:
-            self.value_preds[self.step] = next_value
+            self.buffers["value_preds"][self.step] = next_value
             gae = 0
             for step in reversed(range(self.step)):
                 delta = (
-                    self.rewards[step]
-                    + gamma * self.value_preds[step + 1] * self.masks[step + 1]
-                    - self.value_preds[step]
+                    self.buffers["rewards"][step]
```
mathfac, line 157: [no action needed] As this is `common/rollout_storage.py` not sure how common this calculation are for common RL rollout.

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
-                actions_batch.append(self.actions[: self.step, ind])
-                prev_actions_batch.append(self.prev_actions[: self.step, ind])
-                value_preds_batch.append(self.value_preds[: self.step, ind])
-                return_batch.append(self.returns[: self.step, ind])
-                masks_batch.append(self.masks[: self.step, ind])
-                old_action_log_probs_batch.append(
-                    self.action_log_probs[: self.step, ind]
+        if num_simulators % num_mini_batch != 0:
+            warnings.warn(
+                "Number of simulators ({}) is not a multiple of the number of mini batches ({})".format(
```
mathfac, line 188: Possibly, we can add why that `Number of simulators should be a multiple of the number of mini batches`

habitat_baselines/common/rollout_storage.py
```diff
@@ -4,237 +4,196 @@
...
-        Args:
-            t: first dimension of tensor.
-            n: second dimension of tensor.
-            tensor: target tensor to be flattened.
-
-        Returns:
-            flattened tensor of size (t*n, ...)
-        """
-        return tensor.view(t * n, *tensor.size()[2:])
+        for inds in torch.randperm(num_simulators).chunk(num_mini_batch):
```
mathfac, line 192: By the way, why do we want random permutation here?

habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import copy
```
mathfac, line 6:
```suggestion

import copy
```

habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,156 @@
...
+import torch
+
+TensorLike = Union[torch.Tensor, np.ndarray, numbers.Real]
+DictTree = Dict[str, Union[TensorLike, "DictTree"]]
+TensorIndexType = Union[int, slice, Tuple[Union[int, slice], ...]]
+
+
+class TensorDict(Dict[str, Union["TensorDict", torch.Tensor]]):
+    r"""A dictionary of tensors that can be indexed like a tensor or like a dictionary.  Also
+        supports access via dot notation.
```
mathfac, line 20: Is `access via dot notation.` still accurate?

habitat_baselines/config/default.py
```diff
@@ -30,14 +31,17 @@
 _C.VIDEO_DIR = "video_dir"
 _C.TEST_EPISODE_COUNT = -1
 _C.EVAL_CKPT_PATH_DIR = "data/checkpoints"  # path to ckpt or path to ckpts dir
-_C.NUM_PROCESSES = 16
+_C.NUM_SIMULATORS = 16
```
mathfac, line 34: Maybe, `NUM_ENVIRONMENTS` will is more accurate as there is a VecEnv class and Env and Sim wrapper around each simulator.

habitat_baselines/config/default.py
```diff
@@ -30,14 +31,17 @@
...
+_C.NUM_PROCESSES = -1
```
mathfac, line 35:
```suggestion
_C.NUM_PROCESSES = -1 #  deprecate
```

habitat_baselines/config/imagenav/ddppo_imagenav_gibson.yaml
```diff
@@ -43,10 +45,15 @@ RL:
     use_normalized_advantage: False
 
     hidden_size: 512
+
+    # Use double buffered sampling, typically helps
```
mathfac, line 49: Valuable comment, let's move to the doc string then it will be part of the documentation.

habitat_baselines/rl/ddppo/algo/ddp_utils.py
```diff
@@ -29,6 +31,46 @@
 )
 
 
+def is_slurm_job() -> bool:
+    return SLURM_JOBID is not None
+
+
+def is_slurm_batch_job() -> bool:
+    return is_slurm_job() and os.environ.get("SLURM_JOB_NAME", "bash") not in (
```
mathfac, line 39: Maybe, add a small comment or doc string how does it work. Not obvious connection `bash` and batch job property.

habitat_baselines/rl/ddppo/algo/ddp_utils.py
```diff
@@ -29,6 +31,46 @@
...
+def rank0_only() -> bool:
+    ...
+
+
+@overload
+def rank0_only(fn: Callable) -> Callable:
+    ...
+
+
+def rank0_only(fn: Optional[Callable] = None) -> Union[Callable, bool]:
```
mathfac, line 58: Really like the elegance that wrapper bring to our code base. Let's add the docstring that this decorator designed to wrap methods that are allowed called only on master worker if distrib setting is used. Thank you!

habitat_baselines/common/rollout_storage.py
erikwijmans: It's pretty common.  We would need a different data structure for off policy.

habitat_baselines/common/rollout_storage.py
erikwijmans: That way the mini batches are different for each epoch

habitat_baselines/config/imagenav/ddppo_imagenav_gibson.yaml
erikwijmans: Do we have docs for the config?

habitat_baselines/common/rollout_storage.py
mathfac: That means, if we have `num_simulators > num_mini_batch` the way the ids of environments/episodes will change how it's presented in mini_batches.

habitat_baselines/config/default.py
erikwijmans: Yeah, I like that way better

habitat_baselines/common/tensor_dict.py
erikwijmans: Ah yeah, it isn't

habitat_baselines/common/rollout_storage.py
erikwijmans: It is used in ppo_trainer: https://github.com/facebookresearch/habitat-lab/blob/etw-updates/habitat_baselines/rl/ppo/ppo_trainer.py#L504

habitat_baselines/common/rollout_storage.py
erikwijmans: Makes sense

habitat_baselines/common/rollout_storage.py
erikwijmans: It is really just a speed optimization.  It doesn't take any extra memory, it will either speed things up or slow things down.

habitat_baselines/common/rollout_storage.py
erikwijmans: Yep.  Just to clarify, we have always done this, this is just a cleaner and more robust way to do it.

habitat_baselines/config/imagenav/ddppo_imagenav_gibson.yaml
mathfac: No, we don't. But at leas have this repeated where the parameter is used would be great.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -20,29 +19,31 @@ def __init__(self, n_channels: int) -> None:
...
         self._count: torch.Tensor = self._count
-        self._distributed = distrib.is_initialized()
 
     def forward(self, x: Tensor) -> Tensor:
         if self.training:
-            new_mean = F.adaptive_avg_pool2d(x, 1).sum(0, keepdim=True)  # type: ignore
-            new_count = torch.full_like(self._count, x.size(0))
+            n = x.size(0)
+            x_channels_first = (
+                x.transpose(1, 0).contiguous().view(x.size(1), -1)
```
mathfac, line 27: some comment what's happening here will be useful.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -20,29 +19,31 @@ def __init__(self, n_channels: int) -> None:
...
 
-            if self._distributed:
+            if distrib.is_initialized():
                 distrib.all_reduce(new_mean)
                 distrib.all_reduce(new_count)
+                new_mean /= distrib.get_world_size()
 
-            new_mean = new_mean / new_count
-
-            new_var = F.adaptive_avg_pool2d((x - new_mean).pow(2), 1).sum(  # type: ignore
```
mathfac, line 36: Why you decided to get rid of `adaptive_avg_pool2d`?

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -62,7 +63,7 @@ def forward(self, x: Tensor) -> Tensor:
 
             self._count += new_count
 
-        stdev = torch.sqrt(
+        inv_stdev = torch.rsqrt(
             torch.max(self._var, torch.full_like(self._var, 1e-2))
         )
-        return (x - self._mean) / stdev
+        return torch.addcmul(-self._mean * inv_stdev, x, inv_stdev)
```
mathfac, line 69: It doesn't look cleaner with rsqrt and addcmul.

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -1,88 +1,287 @@
...
+) -> Tuple[
+    PackedSequence, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor
+]:
+    r"""Create a PackedSequence input for an RNN such that each
+    set of steps that are part of the same episode are all part of
+    a batch in the PackedSequence.
+
+    Use the returned select_inds and build_rnn_out_from_seq to invert this.
+
+    :param x: A (T*N, -1) tensor of the data to build the PackedSequence out of
```
mathfac, line 165:
```suggestion
    :param x: A (T * N, -1) tensor of the data to build the PackedSequence out of
```

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -1,88 +1,287 @@
...
+    :param not_dones: A (T*N) tensor where not_dones[i] == False indicates an episode is done
```
mathfac, line 166:
```suggestion
    :param not_dones: A (T * N) tensor where not_dones[i] == False indicates an episode is done
```

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -1,88 +1,287 @@
...
+    return (
+        x_seq,
+        rnn_states,
+        select_inds,
+        rnn_state_batch_inds,
+        last_episode_in_batch_mask,
+    )
+
+
+def build_rnn_out_from_seq(
```
mathfac, line 226: public method requires a doc string

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -1,88 +1,287 @@
...
+        torch.masked_select(rnn_state_batch_inds, last_episode_in_batch_mask)
+        .view(1, N, 1)
+        .expand_as(output_hidden_states)
+    )
+    output_hidden_states.scatter_(1, scatter_inds, last_hidden_states)
+
+    return x, output_hidden_states
+
+
+class RNNStateEncoder(nn.Module):
```
mathfac, line 251: Let's add doc string for this class.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -188,283 +402,411 @@ def _collect_rollout_step(
...
-                last_observation,
-                rollouts.recurrent_hidden_states[rollouts.step],
-                rollouts.prev_actions[rollouts.step],
-                rollouts.masks[rollouts.step],
-            ).detach()
+                step_batch["observations"],
+                step_batch["recurrent_hidden_states"],
+                step_batch["prev_actions"],
+                step_batch["masks"],
+            )
```
mathfac, line 528: Curious, why no need in `.detach` anymore?

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -520,27 +867,33 @@ def _eval_checkpoint(
         batch = apply_obs_transforms_batch(batch, self.obs_transforms)
 
         current_episode_reward = torch.zeros(
-            self.envs.num_envs, 1, device=self.device
+            self.envs.num_envs, 1, device="cpu"
```
mathfac, line 870: why do we want to allocate rewards, masks and etc on cpu?

habitat_baselines/common/tensor_dict.py
```diff
@@ -0,0 +1,157 @@
...
+
+import numpy as np
+import torch
+
+TensorLike = Union[torch.Tensor, np.ndarray, numbers.Real]
+DictTree = Dict[str, Union[TensorLike, "DictTree"]]
+TensorIndexType = Union[int, slice, Tuple[Union[int, slice], ...]]
+
+
+class TensorDict(Dict[str, Union["TensorDict", torch.Tensor]]):
```
mathfac, line 20: Can we add some unit test for the class to cover main usage scenarios. Thank you!

mathfac (approved):
@erikwijmans, thank you a lot for such thoughtful changes and I would say redesign. A lot of technical debt is going to be killed with this PR, as well as efficiency and code readability improvements. Left the last comments. Looking forward to merging it.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
erikwijmans: Agreed, but this is faster and and gives extra stability for fp16 since it is a fused op.  I can move this to that PR if you'd prefer

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
erikwijmans: Just .mean() is typically faster and with the channels format it isn't needed

habitat_baselines/common/base_trainer.py
erikwijmans: I found it very annoying to have to calculate how many updates per checkpoint I wanted since the number of frames per update is a factor of number of DD-PPO workers, number of simulators, and rollout length.  I can keep both functionality if you'd like.  I just deleted the other one since I haven't found myself wanting it since I made this change in my fork 8+ months ago.

habitat_baselines/rl/ppo/ppo_trainer.py
erikwijmans: It was never necessary.  This is all run under `with torch.no_grad():` so there is not autograd graph to detach from.

habitat_baselines/rl/ppo/ppo_trainer.py
erikwijmans: The little update operations that we do with them are much faster on the CPU than GPU because the GPU is total overkill for adding a handful over numbers together.

habitat_baselines/common/tensor_dict.py
erikwijmans: Yep

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
Skylion007: Really, it's faster than rsqrt? I thought the hwole point or rsqrt is that it uses dedicated instructions which are faster than sqrt + divide?

mathfac: > Really, it's faster than rsqrt?
@erikwijmans said that rsqrt and addcmul are faster and and gives extra stability for fp16.
I am totally ok to have it in this PR, just was curious.

habitat_baselines/rl/ppo/ppo_trainer.py
mathfac: Got it. Was it measured or based on experience?

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
Skylion007: Ah, I forgot we were dealing with FP16, yeah rsqrt is going to have stability issues with that.

habitat_baselines/rl/ppo/ppo_trainer.py
erikwijmans: Measured.  It made more of a difference with the work with Brennan (this was actually a non-trivial bottleneck there), here it helped just a little bit (couple FPS for pointnav) but hey, why not!

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
erikwijmans: This PR is swapping to using rsqrt because it is more stable (actually that is still done in fp32, but the addcmul is done in fp16 in the fp16 branch)

erikwijmans: I double checked, rsqrt is going to be more stable in fp16 because sqrt is implemented as 1.0 / rsqrt(x) unless Nvidia added a hardware instruction for sqrt recently.

PR #548, opened 2020-12-22
Support properties/members with call_at

erikwijmans:
## Motivation and Context

This has been a long standing TODO

## How Has This Been Tested

With tests.  We had some specific commands for retrieving specific members/properties that are now done with this instead.

## Types of changes

### DIFF
--- a/habitat/core/vector_env.py
+++ b/habitat/core/vector_env.py
@@ -50,14 +50,8 @@ STEP_COMMAND = "step"
 RESET_COMMAND = "reset"
 RENDER_COMMAND = "render"
 CLOSE_COMMAND = "close"
-OBSERVATION_SPACE_COMMAND = "observation_space"
-ACTION_SPACE_COMMAND = "action_space"
-NUMBER_OF_EPISODES_COMMAND = "number_of_episodes"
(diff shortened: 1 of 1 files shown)


habitat/core/vector_env.py
```diff
@@ -302,7 +285,7 @@ def _spawn_workers(
     def current_episodes(self):
         self._is_waiting = True
         for write_fn in self._connection_write_fns:
-            write_fn((EPISODE_COMMAND, None))
+            write_fn((CALL_COMMAND, ("current_episode", None)))
```
mathfac, line 288: Maybe, worth to keep consts `EPISODE_COMMAND` otherwise we have string const that are related to code function names (reflection) buried in the code.

```suggestion
            write_fn((CALL_COMMAND, (EPISODE_COMMAND, None)))
```

mathfac (approved):
Less of if/else code, that great!

habitat/core/vector_env.py
erikwijmans: Done

Skylion007 (approved):

Skylion007:
Looks good to me, let's merge.

PR #557, opened 2021-01-09
Updated baselines

erikwijmans:
## Motivation and Context

Switch to using ResNet18 as the baseline network.  We now know that SimpleCNN is quite bad and sometimes pretty dramatically so.

## How Has This Been Tested

Just changes to default values.

## Types of changes

### DIFF
--- a/habitat_baselines/config/default.py
+++ b/habitat_baselines/config/default.py
@@ -51,13 +51,13 @@ _C.EVAL.USE_CKPT_CONFIG = True
 _C.RL = CN()
 _C.RL.REWARD_MEASURE = "distance_to_goal"
 _C.RL.SUCCESS_MEASURE = "spl"
-_C.RL.SUCCESS_REWARD = 10.0
+_C.RL.SUCCESS_REWARD = 2.5
 _C.RL.SLACK_REWARD = -0.01
 # -----------------------------------------------------------------------------
 # POLICY CONFIG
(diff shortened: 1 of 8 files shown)


Skylion007 (approved):

habitat_baselines/config/default.py
```diff
@@ -101,17 +101,17 @@
...
 _C.RL.PPO.hidden_size = 512
 # -----------------------------------------------------------------------------
 # DECENTRALIZED DISTRIBUTED PROXIMAL POLICY OPTIMIZATION (DD-PPO)
 # -----------------------------------------------------------------------------
 _C.RL.DDPPO = CN()
 _C.RL.DDPPO.sync_frac = 0.6
 _C.RL.DDPPO.distrib_backend = "GLOO"
-_C.RL.DDPPO.rnn_type = "LSTM"
-_C.RL.DDPPO.num_recurrent_layers = 2
-_C.RL.DDPPO.backbone = "resnet50"
```
mathfac, line 114: @erikwijmans it looks like opposite to what is stated in PR description we are downgrading from resnet50 to resnet18 and from LSTM to GRU as default settings. Right?

erikwijmans: The DD-PPO configs still have it set to that, this is changing the defaults for things like `ppo_pointnav.yaml`

PR #578, opened 2021-01-31
Autocast and fp16 mixed training

erikwijmans:
## Motivation and Context

Add support for torch.cuda.amp.autocast and fp16 mixed training.  Faster training with the former and faster training with less memory usage with the latter (albeit both jumps introduce instability, so they may not always work).

## How Has This Been Tested

Added integration tests.  The training with these is stable for PointNav with GPS+Compass with these models.  The stability for other things is unknown.

## Types of changes


erikwijmans:
@mathfac @Skylion007 currently our CI is on quite an old version of PyTorch, autocast just finally become usable for RNN models in 1.7.1, do either of you have issue with updating the CI?

Skylion007:
> 
> 
> @mathfac @Skylion007 currently our CI is on quite an old version of PyTorch, autocast just finally become usable for RNN models in 1.7.1, do either of you have issue with updating the CI?

Nope, go ahead.

habitat_baselines/config/default.py
```diff
@@ -55,6 +55,7 @@
 # REINFORCEMENT LEARNING (RL) ENVIRONMENT CONFIG
 # -----------------------------------------------------------------------------
 _C.RL = CN()
+_C.RL.fp16_mode = "off"
```
Skylion007, line 58: What's the stylgeuide for what should be all CAPS or not. shouldn't it FP16_mode?

habitat_baselines/rl/ddppo/multi_node_slurm.sh
```diff
@@ -18,6 +18,7 @@ MASTER_ADDR=$(srun --ntasks=1 hostname 2>&1 | tail -n1)
 export MASTER_ADDR
 
 set -x
-srun python -u -m habitat_baselines.run \
+srun --ntasks=1 python -u -m habitat_baselines.run \
```
Skylion007, line 21: Is this line change intentional

habitat_baselines/rl/ddppo/multi_node_slurm.sh
```diff
@@ -18,6 +18,7 @@ MASTER_ADDR=$(srun --ntasks=1 hostname 2>&1 | tail -n1)
...
     --exp-config habitat_baselines/config/pointnav/ddppo_pointnav.yaml \
-    --run-type train
+    --run-type eval \
+    RL.fp16_mode mixed RL.DDPPO.train_encoder True
```
Skylion007, line 24: likewise this

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -32,15 +48,21 @@ def forward(self, x: Tensor) -> Tensor:
                 x.transpose(1, 0).contiguous().view(x.size(1), -1)
             )
             new_mean = x_channels_first.mean(-1, keepdim=True)
-            new_count = torch.full_like(self._count, n)
+            new_count = torch.full_like(self._count, n).float()
```
Skylion007, line 51: Use the dtype param for torch.full_like

habitat_baselines/rl/ddppo/multi_node_slurm.sh
erikwijmans: Yeah, thought I pushed the commit reverting these before opening the PR :)

Skylion007 (changes requested):
Some initial nits

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -32,15 +48,21 @@ def forward(self, x: Tensor) -> Tensor:
...
-            new_count = torch.full_like(self._count, n)
+            new_count = torch.full_like(self._count, n, dtype=torch.float32)
 
             if distrib.is_initialized():
-                distrib.all_reduce(new_mean)
                 distrib.all_reduce(new_count)
                 new_mean /= distrib.get_world_size()
 
+                new_mean = new_mean.float()
+                distrib.all_reduce(new_mean)
```
Skylion007, line 58: Since you've updated PyTorch, all these distrib.all_reduce can be made async super easily, and they should for the sake of speed. Just make async_op = True and use the returned async op handles.

habitat_baselines/rl/ppo/ppo.py
```diff
@@ -83,59 +171,71 @@ def update(self, rollouts: RolloutStorage) -> Tuple[float, float, float]:
...
-                    value_pred_clipped = batch["value_preds"] + (
-                        values - batch["value_preds"]
-                    ).clamp(-self.clip_param, self.clip_param)
-                    value_losses = (values - batch["returns"]).pow(2)
-                    value_losses_clipped = (
-                        value_pred_clipped - batch["returns"]
-                    ).pow(2)
-                    value_loss = 0.5 * torch.max(
-                        value_losses, value_losses_clipped
+                with torch.cuda.amp.autocast() if self._fp16_autocast else contextlib.suppress():
```
Skylion007, line 174: Do you really mean to use ```contextlib.supress``` here? Wouldn't ```contextlib.nullcontext``` be better?

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
erikwijmans: If we are trying to eke out a little bit more efficiency here, going to 1 all_reduce call is always better (even with the async ones), so I did that.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -4,18 +4,34 @@
...
+def welford_update(mean, var, count, new_mean, new_var, new_count):
+    m_a = var * count
+    m_b = new_var * new_count
+    M2 = (
+        m_a
+        + m_b
+        + (new_mean - mean).pow(2) * count * new_count / (count + new_count)
+    )
+
+    var = M2 / (count + new_count)
```
mathfac, line 23: Based on [Parallel algorithm ](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) it should be:
```suggestion
    var = M2 / (count + new_count - 1)
```

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -4,18 +4,34 @@
...
 # LICENSE file in the root directory of this source tree.
 
+
 import torch
 from torch import Tensor
 from torch import distributed as distrib
 from torch import nn as nn
 
 
+def welford_update(mean, var, count, new_mean, new_var, new_count):
```
mathfac, line 14: Let's add docs string with link or description of the algorithm. This looks more like  [Parallel algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) not Welford one.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -4,18 +4,34 @@
...
+        + (new_mean - mean).pow(2) * count * new_count / (count + new_count)
+    )
+
+    var = M2 / (count + new_count)
+    mean = (count * mean + new_count * new_mean) / (count + new_count)
+
+    return mean, var, count + new_count
+
+
 class RunningMeanAndVar(nn.Module):
```
mathfac, line 29: Doc string needed, pretty important class to reuse for other observation's normalization.

habitat_baselines/rl/ppo/ppo.py
```diff
@@ -4,20 +4,71 @@
...
 from habitat.utils import profiling_wrapper
 from habitat_baselines.common.rollout_storage import RolloutStorage
+from habitat_baselines.rl.ddppo.algo.ddp_utils import rank0_only
 from habitat_baselines.rl.ppo.policy import Policy
+from habitat_baselines.utils.common import cast_to_float_if_half
 
 EPS_PPO = 1e-5
 
 
+class FP16OptimParamManager:
```
mathfac, line 25: Does it work for PPO only? If not let's move to common and give it a docstring how to use it.

mathfac (commented):
Thank you for the awesome perf change. Let's add "faster training with less memory usage" to PR name.
Do you have rough estimation of possible speed/memory gains?
This PR sacrifice PPO code readability quite a bit. I thought about fp16 decorator or wrapper to encapsulate all the logic in separate class from PPO logic, but too many `if self._fp16*` calls through all the PPO trainer. Open to ideas.

habitat_baselines/config/default.py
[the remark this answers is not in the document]
erikwijmans: This takes 3 different values and yacs won't let things change types :/

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
erikwijmans: Yeah, this is welford's parallel mean/var algorithm (which can also be used for running)

habitat_baselines/config/default.py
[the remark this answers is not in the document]
mathfac: ow, I see. Thought it just 2. Let's add a comment what all 3 modes are.

habitat_baselines/rl/ppo/ppo.py
erikwijmans: It's general-ish.  I made my own as I haven't found one we could rely on as a dependency instead.  fairscale may eventually come to fit that.  My hesitation for moving it to common is that things like this naturally have a lot of sharp edges

erikwijmans:
> Do you have rough estimation of possible speed/memory gains?

It is highly dependent on the exact workload so it's impossible to say anything concrete.  Using mixed can reduce memory usage by as much as 50% (since activations will all be in fp16).  I've seen both autocast and mixed can both increase speed by as much as 2x.  I've also seen autocast and mixed not increase speed at all -- there is overhead involved with both, so if your benefit from using tensor cores is less than or equal to that overhead, it won't matter.

habitat_baselines/rl/ppo/ppo.py
Skylion007: Agree with @erikwijmans Apex is a pain.

Skylion007:
Yeah, the memory saving parts are the more beneficial part of this PR for sure.

Skylion007 (changes requested):
Also you should pull from master to clear the CI error.

habitat_baselines/rl/ddppo/policy/running_mean_and_var.py
```diff
@@ -4,18 +4,34 @@
...
+    m_a = var * count
+    m_b = new_var * new_count
+    M2 = (
+        m_a
+        + m_b
+        + (new_mean - mean).pow(2) * count * new_count / (count + new_count)
+    )
+
+    var = M2 / (count + new_count)
+    mean = (count * mean + new_count * new_mean) / (count + new_count)
```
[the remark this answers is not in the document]
erikwijmans: Oh python, can't do trivial compiler optimizations because it doesn't have one!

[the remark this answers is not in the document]
Skylion007: Tensorflow actually can, but only because the computation graph is static :P

Skylion007:
@erikwijmans Wanted to see if we could this PR merged as well. :)

erikwijmans:
I think I am gonna trim this to just the autocast version but keep this open so people can see an example of fp16 mixed if they want to try it.  As I have played more with fp16 mixed, it can be really hard to make stable so I am not sure if shipping it makes sense.

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -164,7 +166,8 @@ def forward(self, observations: Dict[str, torch.Tensor]) -> torch.Tensor:  # typ
             # permute tensor to dimension [BATCH x CHANNEL x HEIGHT X WIDTH]
             rgb_observations = rgb_observations.permute(0, 3, 1, 2)
             rgb_observations = (
-                rgb_observations.float() / 255.0
+                rgb_observations.to(dtype=next(self.parameters()).dtype)
```
eundersander, line 169: What about the type of the depth observations? If they remain fp32, then I believe the torch.cat below will convert everything back to fp32.

erikwijmans: They are already fp16 when in fp16 mode as they stored as fp16 in the rollout storage for further memory savings.

Also, torch.cat won't cast, it'll error if things are different dtypes (unless something changed...)

PR #580, opened 2021-02-02
Enhance the display of the top down map for whatever floor the agent is currently on

Skylark0924:
## Motivation and Context
Hi there, when I use the habitat simulator for object navigation tasks, I notice that the top-down map provided by the simulator is quite confusing. **Goal bboxes and their view points from other floors are wrongly marked** in the top-down map of the current layer, which is misleading.

## How Has This Been Tested
For example, the original top-down map for finding a `cushion` in `1pXnuDYAj8r` is like this:

However, most of the highlighted goals are actually not on this layer. After adding the height condition in `nav.py`, we can finally get the correct map.

## Types of changes

## Checklist

### DIFF
--- a/habitat/tasks/nav/nav.py
+++ b/habitat/tasks/nav/nav.py
@@ -731,26 +731,34 @@ class TopDownMap(Measure):
     def _draw_goals_view_points(self, episode):
         if self._config.DRAW_VIEW_POINTS:
             for goal in episode.goals:
-                try:
-                    if goal.view_points is not None:
-                        for view_point in goal.view_points:
-                            self._draw_point(
-                                view_point.agent_state.position,
-                                maps.MAP_VIEW_POINT_INDICATOR,
-                            )
-                except AttributeError:
-                    pass
+                if goal.position[1] > \
+                    self._sim.get_agent(0).state.position[1] and \
+                    goal.position[1] < \
+                    self._sim.get_agent(0).state.position[1] + 2:
+                    try:
+                        if goal.view_points is not None:
+                            for view_point in goal.view_points:
+                                self._draw_point(
+                                    view_point.agent_state.position,
+                                    maps.MAP_VIEW_POINT_INDICATOR,
+                                )
+                    except AttributeError:
+                        pass
 
     def _draw_goals_positions(self, episode):
         if self._config.DRAW_GOAL_POSITIONS:
 
             for goal in episode.goals:
-                try:
-                    self._draw_point(
-                        goal.position, maps.MAP_TARGET_POINT_INDICATOR
-                    )
-                except AttributeError:
-                    pass
+                if goal.position[1] > \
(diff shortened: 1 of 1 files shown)


dhruvbatra:
This is great! Thank you for contributing.

Skylion007:
@Skylark0924 please run pre-commit run -a and resolve all existing linter errors.

habitat/tasks/nav/nav.py
```diff
@@ -731,26 +731,34 @@ def _draw_point(self, position, point_type):
...
-                try:
-                    if goal.view_points is not None:
-                        for view_point in goal.view_points:
-                            self._draw_point(
-                                view_point.agent_state.position,
-                                maps.MAP_VIEW_POINT_INDICATOR,
-                            )
-                except AttributeError:
-                    pass
+                if goal.position[1] > \
```
erikwijmans, line 734: Looks like this pattern is repeated a lot, can you factor that out into a helper function?  Something like 

```python
def is_one_same_floor(pt_source, pt_ref, y_tolerance=2.0):
```

Skylark0924: Sure, let me change it.

habitat/tasks/nav/nav.py
```diff
@@ -821,6 +824,14 @@ def _draw_shortest_path(
                 self.line_thickness,
             )
 
+    def _is_on_same_floor(self, pt, ref_floor_pt=None, ceiling_height=2.0):
```
erikwijmans, line 827:
```suggestion
    def _is_on_same_floor(self, height, ref_floor_height=None, ceiling_height=2.0):
```

Since it is just the y component in the function call.  This looks good other than this!

Skylark0924:
> @Skylark0924 please run pre-commit run -a and resolve all existing linter errors.

Hi, I think all the test errors are solved. Is there anything I can help with?

habitat/tasks/nav/nav.py
```diff
@@ -821,6 +824,16 @@ def _draw_shortest_path(
...
             )
 
+    def _is_on_same_floor(
+        self, height, ref_floor_height=None, ceiling_height=2.0
+    ):
+        same_floor = False
+        if ref_floor_height is None:
+            ref_floor_height = self._sim.get_agent(0).state.position[1]
+        if ref_floor_height < height < ref_floor_height + ceiling_height:
+            same_floor = True
```
Skylion007, line 834: This can be simplified to a  single return True and return False

Skylark0924: Hi, I have already changed it. Is there something else?

Skylion007 (approved):

erikwijmans (approved):

mathfac (approved):
Thank you for this amazing fix.

PR #582, opened 2021-02-03
[PERFORMANCE] Pickle5 ConnectionWrapper for VectorEnv at least (2-5% speed up)

Skylion007:
## Motivation and Context

* [PEP 574](https://www.python.org/dev/peps/pep-0574/) introduced a zero-copy pickling method that makes the pickling of Numpy arrays significantly faster. However, this feature is only supported on the latest Pickle protocol (5) which is not used by the default for multiprocessing (or even torch.multiprocessing, although it is often used in wrappers like Queues). 

I ported and slightly modified the torch Connection wrapper code to habitat-lab and improved it by adding a pickle5 backport. The backport port brings this functionality back to Python<3.8 After using this PR, it should remove a copy from NumPy arrays.

Furthermore, this should not affect the pickling of any other parts of the code stack or any other multiprocessing code.

This speed up is likely to be a lot more on higher resolution outputs from habitat-sim as well as the sensor numbers increase.

This change should be fully backwards compatible and speed things up across the board without any code changes for the end user.

## How Has This Been Tested
Local testing shows at least 2-5% speed up in VectorEnv evaluation.  The entire test suite appears now takes 10 seconds less (despite only a fraction of the tests actually using VectorEnv.) I have tested this in Python 3.8 as well as Python 3.6.

As an aside, we should use add tests for pytest-benchmarking on common habitat-lab components like VectorEnv.

## Types of changes

## Checklist

### DIFF
--- a/habitat/utils/pickle5_multiprocessing.py
+++ b/habitat/utils/pickle5_multiprocessing.py
@@ -9,12 +9,21 @@ import sys
 from multiprocessing.connection import Connection
 from multiprocessing.reduction import ForkingPickler as _ForkingPickler
 
+from habitat.core.logging import logger
+
 if sys.version_info[:2] < (3, 8):
     # pickle 5 backport
     try:
         import pickle5 as pickle
     except ImportError:
         import pickle  # type: ignore[no-redef]
+
+        logger.warn(
+            f"""Warning pickle v5 protocol not supported.
+        Falling back to pickle version {pickle.HIGHEST_PROTOCOL}.
+        pip install pickle5 or upgrade to Python 3.8 or greater
+        for faster performance"""
+        )
 
     class ForkingPickler5(pickle.Pickler):
         wrapped = _ForkingPickler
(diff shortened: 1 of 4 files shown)


erikwijmans:
This is great!  I think we could push this further by using CPU shared memory as that would also allow observations, rewards, etc to be batched by default instead of having to batch in habitat_baselines, but that can be done later

habitat/utils/pickle5_multiprocessing.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+from multiprocessing.connection import Connection
+from multiprocessing.reduction import ForkingPickler as _ForkingPickler
+
+if sys.version_info[:2] < (3, 8):
+    # pickle 5 backport
+    try:
+        import pickle5 as pickle
+    except ImportError:
+        import pickle  # type: ignore[no-redef]
```
Skylion007, line 17: @erikwijmans Only thing I am unsure is what to log here: Should we give a warning? And should we use the `habitat.core.logger` or `warning.warn`

erikwijmans: Probably logger.  This will only run once anyways

mathfac (approved):
That's amazing. Thank you for backward compatibility. Let's merge it.

PR #586, opened 2021-02-04
Fix DDP Reduction in PyTorch 1.7 when the model has unused parameters

erikwijmans:
## Motivation and Context

PyTorch 1.7 slightly changed the way Reducer should be used when the model has unused parameters, leading to DD-PPO breaking when the model does, why this is certainly an edge case, it is a problem.  

This PR solve this and attempts to solve future issues by using a wrapper to interface with DDP's forward call instead of using the undocumented Reducer APIs.  While I am not sure if this won't cause it's own issues since it is kinda hacky, this seems like it has the potential to be better.

## How Has This Been Tested

Tested on a model with unused parameters.  I will write a test for this before merging this, but don't have time today.

## Types of changes

### DIFF
--- a/habitat_baselines/rl/ddppo/algo/ddppo.py
+++ b/habitat_baselines/rl/ddppo/algo/ddppo.py
@@ -102,12 +102,15 @@ class DecentralizedDistributedMixin:
                         find_unused_parameters=find_unused_params,
                     )
 
-        self._evaluate_actions = Guard(_EvalActionsWrapper(self.actor_critic), self.device)  # type: ignore
+        self._evaluate_actions_wrapper = Guard(_EvalActionsWrapper(self.actor_critic), self.device)  # type: ignore
 
-    def evaluate_actions(
+    def _evaluate_actions(
         self, observations, rnn_hidden_states, prev_actions, masks, action
     ):
-        return self._evaluate_actions.ddp(
+        r"""Internal method that calls Policy.evaluate_actions.  This is used instead of calling
+        that directly so that that call can be overrided with inheritence
+        """
+        return self._evaluate_actions_wrapper.ddp(
             observations, rnn_hidden_states, prev_actions, masks, action
         )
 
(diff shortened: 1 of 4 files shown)


Skylion007 (approved):

habitat_baselines/rl/ddppo/algo/ddppo.py
```diff
@@ -79,24 +91,25 @@ class Guard:
...
 
-    def before_backward(self, loss: Tensor) -> None:
-        super().before_backward(loss)  # type: ignore
+        self._evaluate_actions = Guard(_EvalActionsWrapper(self.actor_critic), self.device)  # type: ignore
 
-        if self.find_unused_params:
-            self.reducer.prepare_for_backward([loss])  # type: ignore
-        else:
-            self.reducer.prepare_for_backward([])  # type: ignore
+    def evaluate_actions(
```
mathfac, line 107: Doc string will be useful.

habitat_baselines/rl/ppo/ppo.py
```diff
@@ -152,6 +152,13 @@ def update(self, rollouts: RolloutStorage) -> Tuple[float, float, float]:
 
         return value_loss_epoch, action_loss_epoch, dist_entropy_epoch
 
+    def evaluate_actions(
```
mathfac, line 155: Doc string will be useful.

mathfac (approved):

habitat_baselines/rl/ppo/ppo.py
erikwijmans: Add a little doc string and made it internal since it just exists to be overridden with inheritance.

erikwijmans:
@mathfac @Skylion007 does this test look reasonable?  It's a little ugly but I wanted it to be as close to testing the actual code path used in training as possible.

Skylion007 (approved):

mathfac:
Yes, looks good.

PR #594, opened 2021-02-20
[Baselines Performance]Force pytorch to be single threaded

erikwijmans:
## Motivation and Context

PyTorch keeps making more CPU things multi-threaded and that always ends up reducing our perf when training is being done on the GPU, so force it to be single threaded.

## How Has This Been Tested

Torch is single threaded!

## Types of changes

### DIFF
--- a/habitat_baselines/config/default.py
+++ b/habitat_baselines/config/default.py
@@ -44,10 +44,15 @@ _C.LOG_INTERVAL = 10
 _C.LOG_FILE = "train.log"
 _C.FORCE_BLIND_POLICY = False
 _C.VERBOSE = True
-# PyTorch keeps making more CPU things multi-threaded and that
-# always ends up reducing our perf when training is being done on the GPU,
-# so force it to be single threaded.
-_C.FORCE_TORCH_SINGLE_THREADED = True
+# For our use case, the CPU side things are mainly memory copies
+# and nothing of substantive compute. PyTorch has been making
+# more and more memory copies parallel, but that just ends up
+# slowing those down dramatically and reducing our perf.
+# This forces it to be single threaded.  The default
+# value is left as false as it's different than how
+# PyTorch normally behaves, but all configs we provide
+# set it to true and yours likely should too
+_C.FORCE_TORCH_SINGLE_THREADED = False
 # -----------------------------------------------------------------------------
 # EVAL CONFIG
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


habitat_baselines/config/default.py
```diff
@@ -44,6 +44,10 @@
 _C.LOG_FILE = "train.log"
 _C.FORCE_BLIND_POLICY = False
 _C.VERBOSE = True
+# PyTorch keeps making more CPU things multi-threaded and that
+# always ends up reducing our perf when training is being done on the GPU,
+# so force it to be single threaded.
+_C.FORCE_TORCH_SINGLE_THREADED = True
```
Skylion007, line 50: This config should control the number of threads, you could have 0, -1, or -2 be the default behavior and otherwise pass things to PyTorch.

I am little skeptical this is ALWAYS a good default to have, it definitely is for DDPPO, but what about other baselines? Are we sure this threading slow things down even on very HIGH core count machine? Furthermore, this could be bad if you need to evaluate parts of your model on the CPU.

I would have it be strickly overriden in another conf rather than be a default in buried in the config here. It screams a bit of a performance gotcha later down the line.

Skylion007 (changes requested):
Concerned a bit about the config and the defaults for ALL baselines.

habitat_baselines/config/default.py
erikwijmans: I've felt a same way for a long time (which I why I am just doing this now despite knowing about this for a long time). I've now seen so many times when multi-threaded pytorch just makes things way slower, we are talking 2x to 3x performance improvement by making it single threaded.  They seem to be increasing the number of things that are multi-threaded so that makes the handling individual cases untenable.

> Are we sure this threading slow things down even on very HIGH core count machine

Yes, the higher the core count the more important this is.  The biggest thing it is making parallel that shouldn't be are memory copies.  My guess it that this is a good idea for mobile but on x86 machines parallel copies are an awful idea because `memcpy` is a hardware instruction that does direct DRAM to DRAM transfers so making it parallel only serves to add contention on the memory bus.  

> Furthermore, this could be bad if you need to evaluate parts of your model on the CPU

Yes, in that case it would be bad and that is why this is configurable.  We currently don't have anything that does that, so I made the default be correct for our current uses.

> I would have it be strickly overriden in another conf rather than be a default in buried in the config here

Yeah, that sounds fine.

srama2512: I tried this just for fun on my project codebase. It speeds things up by 2-2.5x for a modified DDPPO training :) .

PR #595, opened 2021-02-20
[Baselines Performance] Observations batching cache

erikwijmans:
## Motivation and Context

We currently batch cpu-side tensors by doing cpu-stack -> cuda, however, that cpu -> cuda transfer implicitly results in a copy to pinned host memory (cuda can only copy from pinned host memory).  This PR maintains a pool of correctly sized pinned memory buffers so that we can implicitly stack by copying into the correct buffers and then do the transfer.  This skips a cpu -> pinned host memory copy.

For the standard 256x256 sensor this doesn't really matter, for a 512x512 sensor, this results in a 20 FPS (or around 10%) improvement for PointNav training and I expect that will go up as the number of large sensors increases.

## How Has This Been Tested

Tests pass.

## Types of changes

### DIFF
--- a/habitat_baselines/utils/common.py
+++ b/habitat_baselines/utils/common.py
@@ -22,6 +22,7 @@ from typing import (
     Union,
 )
 
+import attr
 import numpy as np
 import torch
 from gym.spaces import Box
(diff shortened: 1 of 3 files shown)


Skylion007:
Yes! I was also just about to do this. The only minor issue this might cause is the sharedmemory VectorEnv batching we were discussing @erikwijmans , but we can work around this when we get to it and hopefully we can manage have the shared memory write directly to the pinned memory.

habitat_baselines/utils/common.py
```diff
@@ -86,11 +87,57 @@ def linear_decay(epoch: int, total_num_updates: int) -> float:
     return 1 - (epoch / float(total_num_updates))
 
 
+@attr.s(auto_attribs=True)
```
Skylion007, line 90: Should also be a slot class, because why not?

Skylion007 (approved):
Aside fro the minor nit about slot classes, look good to me. 👍

Skylion007:
This will also really matter is the multi-sensor case like for pseudo-cubemaps and panorama sensors. 👍

erikwijmans:
> have the shared memory write directly to the pinned memory

Unfortunately shared pinned memory is not a thing (as far as I know).  

In the case of shared memory VecEnv, we'd likely not call this function, so all would be fine.

Skylion007:
Sounds good. Let's get this merged.

PR #598, opened 2021-02-22
[Sensor] Fix hab-lab examples to work with Hierarchical SensorSpec changes

vauduong:
## Motivation and Context

See: https://github.com/facebookresearch/habitat-sim/pull/1090
PR#1090 in habitat-sim introduces a Hierarchical SensorSpec to match the new sensor framework for VisualSensor. 

image

`SensorSpec::create()` is no longer used, use correct SensorSpec initialization to create correct sensor, such as `CameraSensorSpec::create()`. 

This PR fixes examples to use CameraSensorSpec() to initialize sensorSpecs

## How Has This Been Tested

Built and tested locally with version of hab-sim in PR#1090, checked that tests passed

## Types of changes

## Checklist


habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -258,7 +258,7 @@ def create_sim_config(
 
         sensor_specifications = []
         for sensor in _sensor_suite.sensors.values():
-            sim_sensor_cfg = habitat_sim.SensorSpec()
+            sim_sensor_cfg = habitat_sim.CameraSensorSpec()
```
vauduong, line 261: Hi @Skylion007 ! I noticed that in hab-lab sensors we don't have SensorSubTypes, and was wondering what the reasoning for that was - would it be a good idea to add this enum?

For context, I check that SensorSubTypes are correct before initializing a CameraSensorSpec() here: 
https://github.com/facebookresearch/habitat-sim/pull/1090/files#diff-1c82f32ca8e8fe4738a140dd923ac12f5d482300e28446788e41eafa248a62a1R66-R103

vauduong: I'm also not sure how BumpSensor is initialized, since this isn't something I was able to find in hab-sim or hab-lab. I know we can't initialize it with CameraSensorSpec as we do the other sensors though. Any insight to this would be appreciated! @Skylion007 @bigbike @mathfac @erikwijmans @abhiskk

Skylion007:
I started doing that here: https://github.com/facebookresearch/habitat-lab/pull/533

habitat/sims/habitat_simulator/habitat_simulator.py
Skylion007: We don't actually have implementations of BumpSensor, I don't believe.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -257,41 +263,69 @@ def create_sim_config(
...
-            # We know that the Sensor has to be one of these Sensors
-            sensor = cast(HabitatSimVizSensors, sensor)
-            sim_sensor_cfg.sensor_type = sensor.sim_sensor_type
-            sim_sensor_cfg.gpu2gpu_transfer = (
-                self.habitat_config.HABITAT_SIM_V0.GPU_GPU
-            )
-            sensor_specifications.append(sim_sensor_cfg)
+
+            # Check if type VisualSensorSpec, we know that Sensor is one of HabitatSimRGBSensor, HabitatSimDepthSensor, HabitatSimSemanticSensor
+            if (
```
Skylion007, line 278: Not all sensors are VisualSensor in Habitat.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -160,9 +164,11 @@ def get_observation(
 @registry.register_sensor
 class HabitatSimSemanticSensor(SemanticSensor):
     sim_sensor_type: habitat_sim.SensorType
+    sim_sensor_subtype: habitat_sim.SensorSubType
```
Skylion007, line 167: I suppose, we could tie the SensorSubtype to the Sensor Class like this, but I was thinking about having it be a SensorConfig string that determined it (similar to how we do for UUIDs now). Any reason why not do it that way? It would give the user more flexability down the road (unless you expect SensorSubtypes to have entirely different config types).

Doing it like this though causes a quadratic explosion in the number of Sensor modalities -> sensor_subtypes though.

In other words, is there a good reason this cannot/should not be read from the config?

This is all fine for now since we only support PINHOLE, but will be annoying later when we supports RGBFisheye/DepthFisheye etc...

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -257,41 +263,69 @@ def create_sim_config(
         )
 
         sensor_specifications = []
+        VisualSensorTypeSet = {
```
Skylion007, line 266: We can programatically get the values from a enum by iterating through them with `.value()` 

Also wouldn't this be better defined by some inheritance structure?

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -257,41 +263,69 @@ def create_sim_config(
...
+
+            # Check if type VisualSensorSpec, we know that Sensor is one of HabitatSimRGBSensor, HabitatSimDepthSensor, HabitatSimSemanticSensor
+            if (
+                getattr(sensor, "sim_sensor_type", [])
+                not in VisualSensorTypeSet
+            ):
+                raise ValueError(
+                    f"""{getattr(sensor, "sim_sensor_type", [])} is an illegal sensorType that is not implemented yet"""
+                )
+            else:
```
Skylion007, line 285: No reason for an else statement here, athe ValueError will terminate it.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -257,41 +263,69 @@ def create_sim_config(
...
+                # Check if type CameraSensorSpec
+                if (
+                    getattr(sensor, "sim_sensor_subtype", [])
+                    not in CameraSensorSubTypeSet
+                ):
+                    raise ValueError(
+                        f"""{getattr(sensor, "sim_sensor_subtype", [])} is an illegal sensorSubType for a VisualSensor"""
+                    )
+                else:
```
Skylion007, line 294: Like wise, no reason for else

Skylion007 (changes requested):
This is fine as a cludge, but there are some issues with support multiple sensor types n the future. Also, some issues with the current error handling / indentation

habitat/sims/habitat_simulator/habitat_simulator.py
vauduong: Not all sensors are VisualSensors but VisualSensorSpecs are the only kinds of specs that can be initialized at the moment.

habitat/sims/habitat_simulator/habitat_simulator.py
vauduong: Hi Aaron! I did this because this file goes through some defined configs and initializes the right sensorSpec. 

So far, only CameraSensorSpec exists, but I'm adding the structure for checks so that when Fisheye/Equirectangular sensors are implemented, we can follow the logic to initialize the right sensorSpec.

habitat/sims/habitat_simulator/habitat_simulator.py
vauduong: Are you saying that instead of checking if it's in a set we should get the value it and compare it with the enum numbers? This is just a way to check which type of SensorSpec to initialize based on the values in the config.

habitat/sims/habitat_simulator/habitat_simulator.py
mathfac: Yes, we have only [CollisionMeasure](https://github.com/facebookresearch/habitat-lab/blob/aa24551fd9942b11b7af4a076d9c05e080594d08/habitat/tasks/nav/nav.py#L670), but that would be great to have BumpSensor, but it looks like it can be HabitatLab Task sensor, not Simulator one.

mathfac:
Looking forward for this change.

habitat/sims/habitat_simulator/habitat_simulator.py
mathfac: Does Subtype infer how observation space will be defined? Does different Subtypes of sensors require separate classes wrappers in HabLab?

Skylion007 (approved):

Skylion007:
Want to let @erikwijmans to take a look at this before merging. Otherwise, looks good to me. We may want to add an error to upgrade habitat_sim though.

habitat/sims/habitat_simulator/habitat_simulator.py
vauduong: Hi @mathfac , I think I don't understand enough about how observation space is defined in hablab vs habsim. What is the difference there and what should I be aware of?

Skylion007: Exactly, I don't think they should require separate wrappers, they should be set by the config.

erikwijmans (approved):
This looks fine to me!

PR #600, opened 2021-02-23
MultiON task integration

shivanshpatel35:
## Motivation and Context

Adding [MultiON](https://shivanshpatel35.github.io/multi-ON/) task (NeurIPS 2020). 

## How Has This Been Tested

Tested locally

## Types of changes

## Checklist

## Dataset download 
Download and place dataset in `data/datasets` folder by running the following commands
```
cd data/datasets
wget -O multinav.zip "http://aspis.cmpt.sfu.ca/projects/multion/multinav.zip"
unzip multinav.zip && rm multinav.zip
cd ../
wget -O multiON_objects.zip "http://aspis.cmpt.sfu.ca/projects/multion/multiON_objects.zip"
unzip multiON_objects.zip && rm multiON_objects.zip
```


habitat/tasks/nav/multi_object_nav_task.py
```diff
@@ -0,0 +1,551 @@
...
+        ep_current_goal_success = task.measurements.measures[
+            CurrentGoalSuccess.cls_uuid
+        ].get_metric()
+
+        if ep_current_goal_success:
+            self._metric += 1 / len(episode.goals)
+
+
+@registry.register_measure
+class MultiONSuccess(Measure):
```
erikwijmans, line 339: In general, I don't think I like this design of task specific measures for metrics that are also well defined in other tasks.  IMO MultiON should just use Success and then that class should dispatch to different implementations based on the task.  In the same way that the topdown map and DistanceToGoal measures currently work for objectnav and pointnav.

habitat/tasks/nav/multi_object_nav_task.py
```diff
@@ -0,0 +1,551 @@
...
+        self._metric = ep_success * (
+            self._start_end_episode_distance
+            / max(
+                self._start_end_episode_distance, self._agent_episode_distance
+            )
+        )
+
+
+@registry.register_measure
+class MultiONPPL(Measure):
```
erikwijmans, line 449: @mathfac @Skylion007 what are you thoughts on generalizing this?  This could be defined for other navigation tasks as
```python
progress * (progress * init_dist_to_goal / max(agent_travel, progress * init_dist_to_goal))
```

This is basically SoftSPL but the inverse Path Length term is weight by the length of the shortest path to reach current progress instead of always shortest path to goal and is actually something that we've discussed using instead.

habitat/tasks/nav/multi_object_nav_task.py
```diff
@@ -0,0 +1,551 @@
...
+            and distance_to_current_goal <= self._config.SUCCESS_DISTANCE
+        ):
+            self._metric = 1
+            task.current_goal_index += 1
+        else:
+            self._metric = 0
+
+
+@registry.register_measure
+class Progress(Measure):
```
erikwijmans, line 298: Either give this a MultiON specific name or generalize it to other navigation tasks.

habitat/tasks/nav/multi_object_nav_task.py
```diff
@@ -0,0 +1,551 @@
...
+                self._dataset.category_to_task_category_id[i]
+                for i in category_name
+            ],
+            dtype=np.int64,
+        )
+        return goals[task.current_goal_index : task.current_goal_index + 1]
+
+
+@registry.register_measure
+class DistanceToCurrentObjectGoal(Measure):
```
erikwijmans, line 206: I think this can be rolled into the existing DistanceToGoal measure.  That measure is ill-defined for MultiON, so defining it to be the current goal makes sense.

erikwijmans (changes requested):
Some questions/comments about how to handle task specific logic for metrics.

habitat/tasks/nav/multi_object_nav_task.py
shivanshpatel35: I am a bit hesitant of the idea of merging MultiON specific MultiONSuccess (and DistanceToCurrentObjectGoal) measures to the previously defined Success (and DistanceToGoal) measures because of the following reasons -
1. This is not scalable. If and when other works will be merged to habitat, all task-specific metrics can't be eventually merged into the existing metric definition.
2. There will be typing conflicts as other works may refer to different episode[[1](https://github.com/facebookresearch/habitat-lab/blob/master/habitat/tasks/nav/nav.py#L962)] and task[[2](https://github.com/facebookresearch/habitat-lab/blob/master/habitat/tasks/nav/nav.py#L519)] class definition. In our case, episode is `MultiObjectGoalNavEpisode` and task is `MultiObjectNavigationTask`.
3. Much of the logic in Success and DistancetoCurrentGoal will pertain to MultiON. This brings an unnecessary overhead for someone trying to work with PointNav/ObjectNav tasks in Habitat.  

This said, I am happy to follow whatever you recommend.

erikwijmans: Concerns 1 and 2 are largely implementation questions that can both be largely resolved by constructing different implementations of a given measure based on the task.  A stack of `if`'s isn't scalable or nice, but we can use other patterns to get around that without framing out that work to the user.

I do see 3, another layer of indirection will make the implementation more confusing.  So it's a question of whether a more confusing implementation is worth a cleaner API.  @mathfac @Skylion007 do either of you have thoughts here?

habitat/core/simulator.py
```diff
@@ -393,3 +393,42 @@ def __enter__(self) -> "Simulator":
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
+
+    def get_existing_object_ids(self) -> List[int]:
```
Skylion007, line 397: Hmm, do we plan to have these be generic simulator options? @mathfac ?

PR #607, opened 2021-03-03
[CI] Add hdf5 headers

erikwijmans:
## Motivation and Context

Fix the CI

## How Has This Been Tested

CI


Skylion007:
The issue is that there aren't any wheels for this distro so it's compiling from source. The issue is that we don't have any hdf5 headers though so it won't compile. @erikwijmans `apt-get install libhdf5-dev` would be a better fix

erikwijmans:
Ah, that makes sense.  Doing that instead!

Skylion007 (approved):

PR #615, opened 2021-03-21
[Baselines] Preemption update

erikwijmans:
## Motivation and Context

Update resuming to work with both types of slurm preemption requeueing and allow it to work without slurm.   Also change things to save the resume state just in the checkpoint dir instead of in a special dir.

## How Has This Been Tested

Locally

## Types of changes


habitat_baselines/config/default.py
```diff
@@ -69,6 +69,19 @@
 _C.RL.SUCCESS_REWARD = 2.5
 _C.RL.SLACK_REWARD = -0.01
 # -----------------------------------------------------------------------------
+# preemption CONFIG
+# -----------------------------------------------------------------------------
+_C.RL.preemption = CN()
+# Append the slurm job ID to the resume state filename if running a slurm job
+# This is useful when you want to have things from a different job but same
+# same checkpoint dir not resume.
+_C.RL.preemption.append_slurm_job_id = False
```
erikwijmans, line 78: @mathfac @Skylion007 what are your thoughts on the default values here?  Currently this is set to do resuming always, but I can also make the default values the old functionality.

habitat_baselines/rl/ddppo/ddp_utils.py
```diff
@@ -125,57 +145,52 @@ def add_signal_handlers() -> None:
...
-        filename = INTERRUPTED_STATE_FILE
-        if not osp.exists(osp.dirname(INTERRUPTED_STATE_FILE)):
-            raise RuntimeError(
-                "Please create a .interrupted_states directory in your home directory for job preemption"
-                "(This is intentionally not created automatically as it can get quite large)"
-            )
+    if isinstance(filename_or_config, Config):
+        filename = resume_state_filename(filename_or_config)
+    else:
+        filename = cast(str, filename_or_config)
```
Skylion007, line 165: Instead of casting here, can't you just do ```filename = str(filename_or_config)``` or do assert isinstance(filename_or_config, str)

habitat_baselines/rl/ddppo/ddp_utils.py
```diff
@@ -125,57 +145,52 @@ def add_signal_handlers() -> None:
...
     """
-    if SLURM_JOBID is None and filename is None:
-        return None
-
-    if filename is None:
-        filename = INTERRUPTED_STATE_FILE
+    if isinstance(filename_or_config, Config):
+        filename = resume_state_filename(filename_or_config)
+    else:
+        filename = cast(str, filename_or_config)
```
Skylion007, line 180: Same as above

Skylion007 (approved):
Some minor typing nits

habitat_baselines/rl/ddppo/ddp_utils.py
erikwijmans: The cast isn't needed here, mypy already knows that filename_or_config is an str, I just added the cast because it makes it a little clearer, but I'll remove it.  I don't think actually forcing this to be a string is the right call because this is python!

Skylion007: @erikwijmans You can do that with a comment or a filename : str = ... . 

cast overrides the typing inference which could lead to it eating typing errors later one.

PR #618, opened 2021-03-29
[BUGFIX] Sets cubemap projection height/width for Fisheye Projection

Skylion007:
Properly sets the cubemap projection depending the resolution of the fisheye camera.

The Fisheye camera did not work for depth images of any resolution other than 256, this should solve that issue. Would appreciate feedback from @matsuren that I didn't break anything else by accident when doing.

## Motivation and Context
Realized that the fisheye projection could only be used on depth images if the output fisheye size was 256x256. 

## How Has This Been Tested
* Locally

## Types of changes

## Checklist

### DIFF
--- a/habitat_baselines/common/obs_transformers.py
+++ b/habitat_baselines/common/obs_transformers.py
@@ -1033,7 +1033,7 @@ class Cube2Fisheye(ProjectionConverter):
         """
 
         # Cubemap input
-        input_projections = get_cubemap_projections()
+        input_projections = get_cubemap_projections(fish_h, fish_w)
 
         # Fisheye output
         output_projection = FisheyeProjection(


Skylion007:
Actually, a bunch of these transforms require that the initial cubemap be 256x256, which prevents high resolution fisheye and other sensors.
We should perhaps parameterize it or cache it based on the resolution of the initial cubemaps, thoughts?

https://github.com/facebookresearch/habitat-lab/blob/d3fb7044b1c6860d4de9ed8baeb9b27c22cb57a0/habitat_baselines/common/obs_transformers.py#L843

Skylion007:
Actually, looking at the code, various other params are also hard coded here, such as equirect2cubemap. We need to parameterize these transforms so that they work at more resolutions than just 256.

bigbike:
Yes, correct. Thanks @Skylion007. Such a fix is a must have to make sure the incoming comparisons between the existing solution and the new habitat-sim shader solution is apples to apples.

mathfac (approved):

matsuren:
Hi, thank you for this modification.

I think you need the input size as well as the output size for the flexibility. (Let's say you want 1024x1024 fisheye image from 512x512 cubemap images).

I didn't add the input size as parameters in my previous pull request because I didn't want to take a risk of breaking some compatibility (I still don't understand the full picture of habitat-lab). It would be great if you could add the input size as the parameters, or get initial cubemap image size as you mentioned.

Also in `Equirect2Cube`, the following line looks like hard coded, 
https://github.com/facebookresearch/habitat-lab/blob/d3fb7044b1c6860d4de9ed8baeb9b27c22cb57a0/habitat_baselines/common/obs_transformers.py#L1133
but `projection` for `EquirectProjection` doesn't require the image size information. So any size is fine (I should've added as comments).

Thanks,

Skylion007:
Going to merge this on for now, because it at least allows square images, which is better than what is currently there. Will submit a follow on PR to fix this properly.

PR #619, opened 2021-04-01
Stream-line of tasks with single Env

NickLucche:
## Motivation and Context

In domains like Continual/Lifelong Learning (or any mult-task context)  it would be convenient to be able to serve multiple tasks without the need to change Env, thus providing an easier interface. This change addresses this use-case building upon the current env-single task abstraction and extending it to env-multiple tasks.  

## How Has This Been Tested
Added pytest untittest for the proposed functionalities in `test_cl_env.py` and `test_cl_config` files. Added an interactive example `habitat_cl_example.py`.
## Types of changes
New feature (non-breaking change which adds functionality).
Added support for task streaming through `tasks` property to CLEnv as well as task `scheduling` policy, essentially going from vanilla configuration (still compatible):
``` 
[ ... ]
TASK:  
  TYPE: Nav-v0
  SUCCESS_DISTANCE: 0.2

  SENSORS: ['POINTGOAL_WITH_GPS_COMPASS_SENSOR']
  POINTGOAL_WITH_GPS_COMPASS_SENSOR:
    GOAL_FORMAT: "POLAR"
    DIMENSIONALITY: 2
  GOAL_SENSOR_UUID: pointgoal_with_gps_compass
[ ... ]
```
to:
```
TASKS:
  - TYPE: Nav-v0
    SUCCESS_DISTANCE: 0.169
    TASK_LABEL: MyTask
    SENSORS: ['POINTGOAL_WITH_GPS_COMPASS_SENSOR']
    SUCCESS:
      SUCCESS_DISTANCE: 0.2
    DATASET:  # you can specify your own dataset per task
      TYPE: PointNav-v1
      SPLIT: test
      DATA_PATH: data/datasets/pointnav/habitat-test-scenes/v1/{split}/{split}.json.gz
  - TYPE: Nav-v0
    SUCCESS_DISTANCE: 1.5
    # "global' dataset is used if unspecified
CHANGE_TASK_BEHAVIOUR:
  TYPE: FIXED # or RANDOM 
  CHANGE_TASK_PROB: 0.1 # for RANDOM, how likely it is to change task at defined horizon
  AFTER_N_EPISODES: 10   # After how many episodes we want to change task
  AFTER_N_CUM_STEPS: 10 # After how many cumulative steps we want to change task
  LOOP: ORDER # RANDOM for sampling a different task with uniform probability
```

## Checklist

### DIFF
--- a/habitat/tasks/multi_task/env.py
+++ b/habitat/tasks/multi_task/env.py
@@ -27,14 +27,32 @@ class TI_TASK_SAMPLING(enum.IntEnum):
 
 class MultiTaskEnv(Env):
     r"""Environment base class extension designed to handle a stream-line of tasks.
+    Multiple tasks can be defined using all configuration properties available for
+    the single task use.
+    The use-cases range from supporting multiple tasks each with a different dataset
+    and scenes up to a single-scene dataset shared by many tasks, each executed
+    without interruption of the episode in a 'continual' stream of tasks.
+    Scene switch will in fact be inhibited when the current active task is changed mid-episode
+    if the tasks share the same scene by replacing the initial position of the agent.
+
+    One can identify the current task being executed by inspecting the `task_idx` key
+    of each observation returned by the MultiTaskEnv.
+    Action space, observation space and episode iterator will also change as tasks are
+    swapped and you can always retrieve them as properties of this class instance using
+    `actions_space`, `observations_space` and `episode_iterator` respectively.
+
+    The environment abstracts away the serving of multiple tasks under
+    the scheduling preferences expressed in the `MULTI_TASK.TASK_ITERATOR` configuration,
+    allowing to cycle through tasks sequentially or sample a different one each time.
+    Timesteps at which active task changes can either be "fixed" or vary by having
+    a `CHANGE_TASK_PROB` which specifies the probability of triggering a change of task
+    after a mininum number of episodes/steps is reached. This can be useful to train or
+    evaluate your agent with no bias on task minimum duration.
 
     The major :ref:`task` component is hereby intended as current task with which
     the agent is currently interacting.
     All tasks handled by the MultiTaskEnv are available through the :ref:`tasks` object
     property.
-
-    The environment abstracts away the serving of multiple tasks under
-    the scheduling preferences expressed in the configuration.
     """
     _tasks: List[EmbodiedTask]
     _episode_iterators: Optional[List[EpisodeIterator]]
@@ -249,6 +267,12 @@ class MultiTaskEnv(Env):
         self._curr_task_label = self._task._config.get(
             "TASK_LABEL", self._curr_task_idx
         )
+
+        # point to new dataset and episodes
+        self._dataset = self._task._dataset
+        self.number_of_episodes = len(self._dataset.episodes)
+        self._episodes = self._dataset.episodes
+
         logger.info(
             "Current task changed from {} (id: {}) to {} (id: {}).".format(
                 self._tasks[prev_task_idx].__class__.__name__,
@@ -309,10 +333,35 @@ class MultiTaskEnv(Env):
         return obs
 
 
-class CRLEnv(RLEnv):
+class MultiTaskRLEnv(RLEnv):
+    r"""Reinforcement Learning (RL) environment class which subclasses ``gym.Env``
+    and adds support for a stream-line of tasks, specifically useful in
+    domains such as Continual Reinforcement Learning or evaluating Hierarchical or
+    Multi-task RL agents.
+
+    This is a wrapper over :ref:`MultiTaskEnv` for RL users. To create custom RL
+    environments users should subclass `MultiTaskRLEnv` and define the following
+    methods: :ref:`get_reward_range()`, :ref:`get_reward()`,
+    :ref:`get_done()`, :ref:`get_info()`, accessing the current task being executed
+    through the underlying `MultiTaskRLEnv` and the `task_idx` returned with every
+    observation.
+
+    As this is a subclass of ``gym.Env``, it implements `reset()` and
+    `step()`.
+    """
+
     def __init__(self, config: Config, dataset: Optional[Dataset]) -> None:
         self._env = MultiTaskEnv(config, dataset)
-        self.observation_space = self._env.observation_space
-        self.action_space = self._env.action_space
-        self.number_of_episodes = self._env.number_of_episodes
         self.reward_range = self.get_reward_range()
+
+    @property
+    def observation_space(self):
+        return self._env.observation_space
+
+    @property
+    def action_space(self):
+        return self._env.action_space
+
+    @property
+    def number_of_episodes(self):
+        return self._env.number_of_episodes


habitat/config/__init__.py
```diff
@@ -120,6 +120,6 @@ def main():
 
 ```"""
 
-from habitat.config.default import Config, get_config
+from habitat.config.default import Config, get_config, get_crl_config
 
 __all__ = ["Config", "get_config"]
```
Skylion007, line 125: get_crl_config is not re-exported with __all__

habitat/core/env.py
```diff
@@ -423,3 +424,238 @@ def __enter__(self):
...
+            :param dataset: reference to dataset used for first task instance level
+            information. Can be defined as :py:`None` in which case
+            dataset will be built using ``make_dataset`` and ``config``.
+        """
+        # we let superclass instantiate first task as to not break/change habitat code
+        super().__init__(config, dataset=dataset)
+        # instatiate other tasks
+        self._tasks = [self._task]
+        self._curr_task_idx = 0
+        if len(config.TASKS) > 1:
```
Skylion007, line 461: Unnecessary if check, as long as there is one TASK in the config.TASKS it should work. If there is only 1, the loop will not execute because config.TASKS[1:] will be empty.

habitat/core/env.py
```diff
@@ -423,3 +424,238 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
+
+
+class CLEnv(Env):
```
Skylion007, line 429: Little iffy on the naming of CLEnv. It's not very discreptive. SequentialTaskEnv or something more descriptive would be helpful.

Skylion007 (changes requested):
You need to run pre-commit on your submission to fix the formatting issues.

habitat/core/env.py
NickLucche: Thanks a lot, I'm going to fix that.

habitat/core/env.py
NickLucche: Yes sorry about that, I narrowed down on context too much.

examples/habitat_cl_example.py
```diff
@@ -0,0 +1,90 @@
...
+from habitat.tasks.nav.nav import NavigationEpisode
+
+FORWARD_KEY = "w"
+LEFT_KEY = "a"
+RIGHT_KEY = "d"
+FINISH = "f"
+
+
+def rgb2bgr(image):
+    return image[:, :, [2, 1, 0]]
```
Skylion007, line 17:
```suggestion
    return image[..., ::-1]
```
or use the cv2.cvtColor function.

examples/habitat_cl_example.py
```diff
@@ -0,0 +1,90 @@
...
+        "-i",
+        "--interactive",
+        help="Run demo interactively",
+        action="store_true",
+    )
+    args = args.parse_args()
+    ### One Env, many tasks ###
+    # cfg = get_config('pointnav.yaml')
+    cfg = get_multi_task_config("configs/test/habitat_cl_example.yaml")
+    env = MultiTaskEnv(config=cfg)
```
Skylion007, line 37: Use a With Env context to ensure proper closure.

examples/habitat_cl_example.py
NickLucche: I didn't want to change too much from https://aihabitat.org/docs/habitat-lab/quickstart.html for clarity. I'll fix it asap.

Skylion007:
```
self = <multiprocessing.connection.Connection object at 0x7f1232449198>
size = 4, read = <built-in function read>

    def _recv(self, size, read=_read):
        buf = io.BytesIO()
        handle = self._handle
        remaining = size
        while remaining > 0:
            chunk = read(handle, remaining)
            n = len(chunk)
            if n == 0:
                if remaining == size:
>                   raise EOFError
E                   EOFError
```
Some of the sub processes are crashing when PyTest is run. Does the test suite pass locally with all the relevant datasets downloaded?

Skylion007 (changes requested):
Fix Failing Tests

Skylion007:
Specifically:
```
       1 failed
         - test/test_habitat_env.py:224 test_rl_vectorized_envs[True]
```

test/test_cl_env.py
```diff
@@ -0,0 +1,256 @@
...
+def test_random_task_loop():
+    change_after = 3
+    cfg = get_config(TEST_CFG_PATH)
+    cfg.defrost()
+    cfg.CHANGE_TASK_BEHAVIOUR.TYPE = "FIXED"
+    cfg.CHANGE_TASK_BEHAVIOUR.AFTER_N_EPISODES = change_after
+    cfg.CHANGE_TASK_BEHAVIOUR.AFTER_N_CUM_STEPS = None
+    cfg.CHANGE_TASK_BEHAVIOUR.LOOP = "RANDOM"
+    cfg.freeze()
+    env = MultiTaskEnv(config=cfg)
```
Skylion007, line 244: Put these in with contexts too

habitat/core/env.py
```diff
@@ -423,3 +424,233 @@ def __enter__(self):
...
+            k.lower(): v
+            for k, v in self._config.ENVIRONMENT.ITERATOR_OPTIONS.items()
+        }
+        iter_option_dict["seed"] = self._config.SEED
+        self._episode_iterator: EpisodeIterator = (
+            self._task._dataset.get_episode_iterator(**iter_option_dict)
+        )
+
+        self.action_space = self._task.action_space
+        print("task changed")
```
Skylion007, line 594: No print statements in the core habitat. Please use logger

NickLucche:
> ```
> self = <multiprocessing.connection.Connection object at 0x7f1232449198>
> size = 4, read = <built-in function read>
> 
>     def _recv(self, size, read=_read):
>         buf = io.BytesIO()
>         handle = self._handle
>         remaining = size
>         while remaining > 0:
>             chunk = read(handle, remaining)
>             n = len(chunk)
>             if n == 0:
>                 if remaining == size:
> >                   raise EOFError
> E                   EOFError
> ```
> 
> Some of the sub processes are crashing when PyTest is run. Does the test suite pass locally with all the relevant datasets downloaded?

I wasn't able to reproduce the bug with the newest commit and I'm not quite sure how to affect subprocesses or the test you mentioned as I've essentially only added new lines and tests.  
Is there any chance that can be related to some other issue/commit?

configs/test/habitat_cl_example.yaml
```diff
@@ -0,0 +1,71 @@
...
+  TYPE: FIXED # or RANDOM
+  CHANGE_TASK_PROB: 0.1 # for RANDOM, how likely it is to change task at defined horizon
+  # AFTER_N_EPISODES: 1   # After how many episodes we want to change task
+  AFTER_N_CUM_STEPS: 10 # After how many cumulative steps we want to change task
+  LOOP: ORDER # RANDOM for sampling a different task with uniform probability
+
+DATASET:
+  TYPE: PointNav-v1
+  SPLIT: val
+  # DATA_PATH: data/datasets/pointnav/gibson/v1/{split}/{split}.json.gz
```
Skylion007, line 70: Any reason to leave this comment here?

examples/habitat_cl_example.py
```diff
@@ -0,0 +1,90 @@
+from argparse import ArgumentParser
```
Skylion007, line 1: May want to add this script to tests/test_examples.py so it's run during code refactors.

habitat/core/env.py
```diff
@@ -423,3 +424,232 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
+
+
+class MultiTaskEnv(Env):
```
Skylion007, line 429: May want to add some logging when tasks are changed.

habitat/config/default.py
```diff
@@ -367,3 +376,31 @@ def get_config(
...
     config.freeze()
     return config
+
+
+def get_multi_task_config(
+    config_paths: Optional[Union[List[str], str]] = None,
+    opts: Optional[list] = None,
+) -> CN:
+    config = get_config(config_paths, opts)
+    if len(config.TASKS):
```
Skylion007, line 386: Can a config.TASKS config be read by the normal config loader? Should it error? @erikwijmans thoughts?

habitat/core/env.py
```diff
@@ -423,3 +424,232 @@ def __enter__(self):
...
+            :param dataset: reference to dataset used for first task instance level
+            information. Can be defined as :py:`None` in which case
+            dataset will be built using ``make_dataset`` and ``config``.
+        """
+        # we let superclass instantiate first task as to not break/change habitat code
+        super().__init__(config, dataset=dataset)
+        # instatiate other tasks
+        self._tasks = [self._task]
+        self._curr_task_idx = 0
+        for task in config.TASKS[1:]:
```
Skylion007, line 461: Wait, so is self.TASKS[0] ignored then and just replace by self._task? If so, why bother including it all? Seems error prone.

habitat/config/default.py
```diff
@@ -367,3 +376,31 @@ def get_config(
 
     config.freeze()
     return config
+
+
+def get_multi_task_config(
```
Skylion007, line 381: I am wondering if this really warrants it's own function or just some flag to enable it? Or maybe we should just deprecate the single task format, thought @erikwijmans @mathfac ?

habitat/core/env.py
Skylion007: Oh I see, it mutated the TASK_CONFIG in the config loader. Still prone to error if a config.TASKS also has a config.TASK defined and people mutate it (as it would if you are printing out the config to rerun experiments). We probably want some error or at least a warning if the config value.TASK is being overwritten. It shouldn't be a hard error though if it's mutated like this. 

A lot of code directly accesses the underlying task config though so I definitely see the potential issue with this. Your way might be the best solution for now, will consult with other contributors.

configs/test/habitat_cl_example.yaml
NickLucche: just wanted to show some available options in config file, I can get rid of that

examples/habitat_cl_example.py
NickLucche: great point thanks

habitat/core/env.py
NickLucche: Exactly, in order to subclass `Env` and make use of the default task init logic without 'wasting' the super() call I had to overwrite config.TASK with the first config.TASKS entry, so that `self._task` remains as the current task variable.
Not sure whether we can have TASK and TASKS co-exist in a meaningful way in the config file, so I'll add a warning message as you suggested when overwriting the existing config.TASK.

habitat/config/default.py
NickLucche: Yes this could easily extend the current single-task format, tho I didn't want to introduce such a change without some feedback on the idea first.

erikwijmans: Yeah, I think this logic should just be in the normal config loader.  No need to even have a flag since config.TASKS can be the empty list and things will fall back to the old behavior.

habitat/config/default.py
erikwijmans: Yep, it can be.  I think merging this function will solve this concern.

habitat/core/__init__.py
```diff
@@ -3,3 +3,14 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+import enum
+
+
+class CHANGE_TASK_BEHAVIOUR(enum.IntEnum):
```
erikwijmans, line 9: Please move these to `env.py` and then import them from there here.  We haven't been defining things in `__init__.py` files so this would match that.

habitat/core/env.py
erikwijmans: I think it would be better suited to have the constructor here do the logic of moving overwriting `config.TASK` with `config.TASKS[0]`.  If folks end up programmatically changing the config, I think this will be less confusing.

habitat/core/env.py
```diff
@@ -423,3 +424,232 @@ def __enter__(self):
...
+        self, action: Union[int, str, Dict[str, Any]], **kwargs
+    ) -> Observations:
+        obs = super().step(action, **kwargs)
+        # check task change behaviour here
+        self._cumulative_steps_counter += 1
+        if self._check_change_task():
+            # task needs to be changed
+            self._change_task()
+
+        obs["task_label"] = self._task_label
```
erikwijmans, line 635: The observation space needs to be updated to expect this `task_label` observation.

NickLucche: Good point thanks, I'll fix that.

NickLucche: I guess I'll pass the current task index instead of `task_label` here with a `Discrete` space, I don't think motivations for a task label are as strong as to define a custom gym space with support to arbitrary strings here, due to warnings in https://github.com/openai/gym/blob/master/gym/spaces/space.py.

habitat/core/env.py
```diff
@@ -423,3 +425,276 @@ def __enter__(self):
...
+            ).item()
+
+        self._task = self._tasks[self._curr_task_idx]
+        # update episode iterator with task dataset
+        iter_option_dict = {
+            k.lower(): v
+            for k, v in self._config.ENVIRONMENT.ITERATOR_OPTIONS.items()
+        }
+        iter_option_dict["seed"] = self._config.SEED
+        self._episode_iterator: EpisodeIterator = (
```
mathfac, line 618: When the task A was switched and after the loop switched to Task A again, does that mean that iteration will start from the initial state as EpisodeIterator is recreated each time? That means that there can be part of episode dataset that will never be reached. Maybe that make more sense to keep iterators alive for each task?

mathfac (changes requested):
@NickLucche thank you for the valuable contribution. Continuous learning and multi tasking are very appealing topics and functionality we would like to see in Habitat.
Your PR looks like great first step to bring it to Habitat Lab and I am pretty surprised with what little delta you are able to add it.

Several questions:
1. Do you have working baselines or the benchmark you already tested it with?
2. As I understand you consider only tasks with same observation and action space right?

Thoughts:
1. Alternative way how we can think about Multi tasking is that mixed Dataset can have Episodes from different Tasks (class), where each episode has a reference to the initialized Task. Then all shuffling and cycling can be done on episodes level using EpisodeIterator capabilities.
2. Tasks switching logic looks very similar but orthogonal to scene switching logic we have. Scene switching logic is very important for training speed and performance. I have concern that MultiTask will have the same needs of Task Sampling logic to be randomized across workers for large scale training.
3. The implementation provided and it's maturity in terms of existing baselines, proposed benchmark from existing tasks is not there to be moved into core package and occupy several level 0 namespaces. I would create the main code into `habitat/tasks/multi_task` package. Then after baselines and benchmarks will be well tested and not only navigation tasks will be supported that will make more sense to move it to core API.

NickLucche:
Hi @mathfac, thanks a lot for your thorough review. I understand there needs to a series of changes in order to add this contribution to core and I'm totally on board with it, in fact with this PR I mainly wanted to pitch in the idea of supporting multiple tasks (which I'd love to leverage for my thesis) and I'm happy to see you're open for supporting this feature.

> Do you have working baselines or the benchmark you already tested it with?
If you're referring to a rl-baseline then no I still didn't train an agent using the multi-task setting but I'm planning to do that asap, I first wanted to put together a few tasks from the already implemented ones and design some meaningful experiment.

> As I understand you consider only tasks with same observation and action space right?
I'd like to generalize as much as possible but I'm not sure atm what the best solution for supporting multiple obs/action spaces is, as I was currently more focused on tasks with same obs space. Please do let me know what you think the best approach here is.

>Alternative way how we can think about Multi tasking is that mixed Dataset can have Episodes from different Tasks (class), where each episode has a reference to the initialized Task. Then all shuffling and cycling can be done on episodes level using EpisodeIterator capabilities.
I thought about this too when I started implementing the feature but I wasn't too sure about how explicit/visible the tasks would be at this level of abstraction; so I opted for inserting tasks at the highest level of abstraction leaving everything else as it was meant to be when designed. I didn't mind about having multiple EpisodeIterators handling different datasets with possibly different scheduling policies, but I don't have a strong opinion on that.

Regarding performance of the multitask env I thought some overhead was acceptable wrt the single-task environment, but I'll have to setup and run some benchmarks to put in some qualitative figures and maybe reason on top of that. I might have some trouble testing on large-scale distributed settings tho. 

I'll make sure to implement the changes you requested as soon as possible.

habitat/core/env.py
```diff
@@ -423,3 +425,276 @@ def __enter__(self):
...
+            self._task._dataset.get_episode_iterator(**iter_option_dict)
+        )
+
+        self.action_space = self._task.action_space
```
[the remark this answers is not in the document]
NickLucche: I'll add a test for both tasks with different action spaces and obs spaces to make sure they are changed as intended.

habitat/core/env.py
```diff
@@ -423,3 +425,276 @@ def __enter__(self):
...
+    FIXED = 0
+    RANDOM = 1
+
+
+class CHANGE_TASK_CYCLE_BEHAVIOUR(enum.IntEnum):
+    ORDER = 0
+    RANDOM = 1
+
+
+class MultiTaskEnv(Env):
```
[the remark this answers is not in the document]
NickLucche: Sure thing

habitat/core/env.py
NickLucche: Yeah probably keeping iterators alive would provide a more consistent experience too. I'll implement that.

habitat/config/default.py
```diff
@@ -364,6 +373,18 @@ def get_config(
 
     if opts:
         config.merge_from_list(opts)
+    # multi-task handling
```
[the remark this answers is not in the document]
NickLucche: I'm not sure I understood that correctly, ideally one wouldn't use both TASK and TASKS, I mostly added the "TASK copy behavior" to have each task in TASKS start off the same way as when you write the config for a single file. 
I didn't mean for it to be used consistently together. 

Should we enforce that division and only copy over the defaults for TASK (regardless of user config for TASK) or would you suggest to keep the default value somewhere just to see where they came from..?

habitat/core/env.py
```diff
@@ -423,3 +425,276 @@ def __enter__(self):
...
+        self._change_task_behavior = config.CHANGE_TASK_BEHAVIOUR.TYPE
+        self._task_cycling_behavior = config.CHANGE_TASK_BEHAVIOUR.LOOP
+        # custom task label can be specified
+        self._task_label = self._task._config.get(
+            "TASK_LABEL", self._curr_task_idx
+        )
+        # add task_idx to observation space
+        self.observation_space = spaces.Dict(
+            {
+                **self._sim.sensor_suite.observation_spaces.spaces,
```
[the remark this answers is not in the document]
NickLucche: I've added tests for tasks with different obs/action spaces in the last commit.

habitat/config/default.py
```diff
@@ -333,6 +333,15 @@ def __init__(self, *args, **kwargs):
 _C.DATASET.DATA_PATH = (
     "data/datasets/pointnav/habitat-test-scenes/v1/{split}/{split}.json.gz"
 )
+# -----------------------------------------------------------------------------
+# Continual Learning -- TASKS
```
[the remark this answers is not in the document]
NickLucche: Yes this sounds good, I guess `SEQ_TASKS.TASKS` describes it better.

habitat/config/default.py
```diff
@@ -333,6 +333,15 @@ def __init__(self, *args, **kwargs):
...
+# -----------------------------------------------------------------------------
+_C.TASKS = []
+# Continual Learning -- Task change policy (how and when to cycle through tasks)
+_C.CHANGE_TASK_BEHAVIOUR = CN()
```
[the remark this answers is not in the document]
NickLucche: One random is meant for successive task sampling, the other random is meant for task changing behavior (allows to change task at non-fixed timesteps).

habitat/core/env.py
```diff
@@ -423,3 +425,276 @@ def __enter__(self):
...
+
+    @property
+    def tasks(self) -> List[EmbodiedTask]:
+        return self._tasks
+
+    @property
+    def current_task_label(self) -> str:
+        return str(self._task_label)
+
+    def _check_change_task(self, is_reset=False):
```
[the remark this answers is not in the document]
NickLucche: I made the code a bit clearer by grouping repetitive checks under a common function

NickLucche (commented):
Main changes are tests with differents tasks and action/observation spaces.
I've also simplified some logic in changing tasks, changed a lot of names to improve clarity and added some more properties to MultiTaskEnv.

Skylion007:
Ping @erikwijmans

PR #624, opened 2021-04-13
Addition of velocity_control actions and support for Gaussian action distribution

naokiyokoyama:
## Motivation and Context

This change allows the agent to move forward and turn simultaneously using habitat-sim's VelocityControl physics object for calculating end positions given a time step, linear velocity, and angular velocity. Support has also been added such that the policy can now output continuous actions, using a Gaussian action distribution. 

## How Has This Been Tested

Train and evaluation can be done the same as before; the line

    action_distribution_type: "gaussian"

just needs to be added under RL.POLICY in the config yaml. The changes made to ppo_trainer.py will then automatically adjust the policy to output a Gaussian distribution, and if training, will make changes to the RolloutStorage to save float value previous actions (rather than long for discrete) and to save two values (lin vel, ang vel) rather than one (discrete action index). I trained for about 100M frames and evaluated the checkpoints successfully for both gaussian and categorical (original) distributions, with the only difference between the runs being the addition or deletion of the aforementioned line to the config yaml. 

## Types of changes

There was an issue where sometimes the quaternions from get_agent_state were quantized too much such that its norm was off by about 1e-4 from a value of 1.0, which caused errors since habitat-sim only allows deviations of 1e-5.

## Checklist

### DIFF
--- a/habitat_baselines/utils/common.py
+++ b/habitat_baselines/utils/common.py
@@ -93,7 +93,7 @@ class ObservationBatchingCache:
     r"""Helper for batching observations that maintains a cpu-side tensor
     that is the right size and is pinned to cuda memory
     """
-    _pool: Dict[Any, torch.Tensor] = attr.Factory(dict)
+    _pool: Dict[Any, Union[torch.Tensor, np.ndarray]] = attr.Factory(dict)
 
     def get(
         self,
@@ -101,7 +101,7 @@ class ObservationBatchingCache:
         sensor_name: str,
         sensor: torch.Tensor,
         device: Optional[torch.device] = None,
-    ) -> torch.Tensor:
+    ) -> Union[torch.Tensor, np.ndarray]:
         r"""Returns a tensor of the right size to batch num_obs observations together
 
         If sensor is a cpu-side tensor and device is a cuda device the batched tensor will
@@ -127,7 +127,9 @@ class ObservationBatchingCache:
             and device.type == "cuda"
             and cache.device.type == "cpu"
         ):
-            cache = cache.pin_memory()
+            # Pytorch indexing is slow,
+            # so convert to numpy
+            cache = cache.pin_memory().numpy()
 
         self._pool[key] = cache
         return cache
@@ -171,21 +173,42 @@ def batch_obs(
 
     for sensor_name in sensor_names:
         for i, obs in enumerate(observations):
-            sensor = torch.as_tensor(obs[sensor_name])
+            sensor = obs[sensor_name]
             if cache is None:
-                batch[sensor_name].append(sensor)
+                batch[sensor_name].append(torch.as_tensor(sensor))
             else:
                 if sensor_name not in batch_t:
                     batch_t[sensor_name] = cache.get(
-                        len(observations), sensor_name, sensor, device
+                        len(observations),
+                        sensor_name,
+                        torch.as_tensor(sensor),
+                        device,
                     )
 
-                batch_t[sensor_name][i].copy_(sensor)
+                # Use isinstance(sensor, np.ndarray) here instead of
+                # np.asarray as this is quickier for the more common
+                # path of sensor being an np.ndarray
+                # np.asarray is ~3x slower than checking
+                if isinstance(sensor, np.ndarray):
+                    batch_t[sensor_name][i] = sensor
+                elif torch.is_tensor(sensor):
+                    batch_t[sensor_name][i].copy_(sensor, non_blocking=True)
+                # If the sensor wasn't a tensor, then it's some CPU side data
+                # so use a numpy array
+                else:
+                    batch_t[sensor_name][i] = np.asarray(sensor)
 
         # With the batching cache, we use pinned mem
         # so we can start the move to the GPU async
         # and continue stacking other things with it
         if cache is not None:
+            # If we were using a numpy array to do indexing and copying,
+            # convert back to torch tensor
+            # We know that batch_t[sensor_name] is either an np.ndarray
+            # or a torch.Tensor, so this is faster than torch.as_tensor
+            if isinstance(batch_t[sensor_name], np.ndarray):
+                batch_t[sensor_name] = torch.from_numpy(batch_t[sensor_name])
+
             batch_t[sensor_name] = batch_t[sensor_name].to(
                 device, non_blocking=True
             )


naokiyokoyama:
I will fix these failing tests and push again to the branch

habitat_baselines/utils/common.py
```diff
@@ -74,6 +74,50 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+        self.std_max = std_max
+
+        nn.init.orthogonal_(self.mu.weight, gain=0.01)
+        nn.init.constant_(self.mu.bias, 0)
+        nn.init.orthogonal_(self.std.weight, gain=0.01)
+        nn.init.constant_(self.std.bias, 0)
+
+    def forward(self, x: Tensor) -> CustomNormal:
+        mu = torch.tanh(self.mu(x))
+        std = torch.clamp(self.std(x), min=self.std_min, max=self.std_max)
```
erikwijmans, line 116: Following up with the convo from slack, I do think we should have some configuration here since there isn't really an agreed upon way to predict STD.

habitat_baselines/utils/common.py
```diff
@@ -74,6 +74,50 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+        self.std_min = std_min
+        self.std_max = std_max
+
+        nn.init.orthogonal_(self.mu.weight, gain=0.01)
+        nn.init.constant_(self.mu.bias, 0)
+        nn.init.orthogonal_(self.std.weight, gain=0.01)
+        nn.init.constant_(self.std.bias, 0)
+
+    def forward(self, x: Tensor) -> CustomNormal:
+        mu = torch.tanh(self.mu(x))
```
erikwijmans, line 115: tanh seems specific to velocity control and won't be applicable everywhere.  Maybe make a more generic base class and then specific version for velocity control?

erikwijmans (changes requested):

habitat_baselines/utils/common.py
naokiyokoyama: What would an alternative to tanh be? Just unbounded actions?

erikwijmans: Yeah, just unbounded

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -544,7 +546,30 @@ def set_agent_state(
         agent = self.get_agent(agent_id)
         new_state = self.get_agent_state(agent_id)
         new_state.position = position
-        new_state.rotation = rotation
+
+        # TODO: Sometimes 'rotation' is off by ~1e-4 in terms of if the
```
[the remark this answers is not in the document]
naokiyokoyama: Yeah I think the best thing to do would be to relax the sim side validator. Should I just submit a PR on sim and remove this change?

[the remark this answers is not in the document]
naokiyokoyama: Removed the fix and submitted a PR to loosen the tolerance to habsim instead, which has been merged (thanks Aaron).

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -433,11 +445,14 @@ def forward(
             goal_output = self.goal_visual_encoder({"rgb": goal_image})
             x.append(self.goal_visual_fc(goal_output))
 
-        prev_actions = prev_actions.squeeze(-1)
-        start_token = torch.zeros_like(prev_actions)
-        prev_actions = self.prev_action_embedding(
-            torch.where(masks.view(-1), prev_actions + 1, start_token)
-        )
+        if self.discrete_actions:
```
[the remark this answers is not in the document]
naokiyokoyama: Missed that. Should be fixed now.

habitat_baselines/utils/common.py
```diff
@@ -74,6 +74,50 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+    def sample(
+        self, sample_shape: Size = torch.Size()  # noqa: B008
+    ) -> Tensor:
+        return super().rsample(sample_shape)
+
+    def log_probs(self, actions):
+        ret = (
+            super()
+            .log_prob(actions)
+            .view(actions.size(0), -1)
```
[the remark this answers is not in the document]
naokiyokoyama: Removed, was not needed.

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -91,6 +93,11 @@ def __init__(self, config=None):
         self._is_distributed = get_distrib_size()[2] > 1
         self._obs_batching_cache = ObservationBatchingCache()
 
+        if config.RL.POLICY.action_distribution_type == "gaussian":
```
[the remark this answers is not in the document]
naokiyokoyama: Changed so that POSSIBLE_ACTIONS needs to be supplied in the task config instead.

habitat_baselines/utils/common.py
naokiyokoyama: Added support for:

- Tanh activation
- no activation
- softplus activation
- std (min max provided by config)
- log_std (min max provided by config)

habitat_baselines/utils/common.py
naokiyokoyama: Added support for no activation

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -240,6 +247,20 @@ def _init_train(self):
 
         self._init_envs()
 
+        if self.using_velocity_ctrl:
+            self.policy_action_space = ActionSpace(
```
erikwijmans, line 251: Is the task action not able to return this properly?  If not we need to fix that interface instead of patching it in the trainer

habitat_baselines/utils/common.py
```diff
@@ -74,6 +75,56 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+        self.std = nn.Linear(num_inputs, num_outputs)
+
+        nn.init.orthogonal_(self.mu.weight, gain=0.01)
+        nn.init.constant_(self.mu.bias, 0)
+        nn.init.orthogonal_(self.std.weight, gain=0.01)
+        nn.init.constant_(self.std.bias, 0)
+
+    def forward(self, x: Tensor) -> CustomNormal:
+        mu = self.mu(x)
+        if self.config.action_activation == "tanh":
```
erikwijmans, line 116: Can you encapsulate this logic in `__init__` -- that'll work with TorchScript which can be worth doing for some models.

i.e.
```
self.mu_act = ....
```

habitat_baselines/utils/common.py
```diff
@@ -74,6 +75,56 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+            mu = torch.tanh(mu)
+        elif self.config.action_activation == "softplus":
+            mu = torch.nn.functional.softplus(mu)
```
erikwijmans, line 119: Softplus on mu seems odd.  I think this is suppose to be on std?

habitat_baselines/utils/common.py
```diff
@@ -74,6 +75,56 @@ def forward(self, x: Tensor) -> CustomFixedCategorical:
...
+
+        std = torch.clamp(self.std(x), min=self.min_std, max=self.max_std)
+        if self.config.use_log_std:
+            std = std.exp(std)
```
erikwijmans, line 123:
```suggestion
            std = torch.exp(std)
```

erikwijmans (changes requested):

habitat_baselines/rl/ppo/ppo_trainer.py
naokiyokoyama: If POSSIBLE_ACTIONS is just ['VELOCITY_CONTROL'], then self.envs.action_space has only one action, but the policy technically has two, linear and angular velocity, so without this fix, the policy would only have one action. I don't know how to fix this other than doing something like ['LINEAR_VELOCITY','ANGULAR_VELOCITY'] and then coding each of the two separately, but I don't know how it would be possible to implement these move_fns independently from one another, as you can't just translate and then rotate (or vice versa).

habitat_baselines/common/rollout_storage.py
```diff
@@ -23,7 +23,9 @@ def __init__(
         action_space,
         recurrent_hidden_state_size,
         num_recurrent_layers=1,
+        action_shape: int = -1,
```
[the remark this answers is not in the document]
naokiyokoyama: It's not the same as dim_actions. dim_actions would be "4" for the original discrete policy, but action shape would be "1" as the RS just needs to store one action index (0, 1, 2, or 3). For gaussian, dim_actions would be 2, and action_shape would be two, as RS would store both the linear and angular velocity. Is there another name I could use?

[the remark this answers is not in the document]
erikwijmans: Ah I see, shape makes sense but maybe make it support a tuple?

habitat_baselines/rl/ppo/ppo_trainer.py
erikwijmans: Hmm, I guess the VELOCITY_CONTROL really needs to be able to return a dictionary of spaces?  @mathfac Is that supported? I remember you working on something like this for the EQA answer action

habitat/tasks/nav/nav.py
```diff
@@ -37,6 +38,8 @@
     quaternion_rotate_vector,
 )
 from habitat.utils.visualizations import fog_of_war, maps
+from habitat_sim.bindings import RigidState
```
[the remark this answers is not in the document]
Skylion007: +1 on this

habitat/tasks/nav/nav.py
```diff
@@ -9,6 +9,7 @@
 from typing import Any, List, Optional, Tuple
 
 import attr
+import magnum as mn
```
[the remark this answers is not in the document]
naokiyokoyama: Softened

habitat/tasks/nav/nav.py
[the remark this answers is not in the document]
naokiyokoyama: Softened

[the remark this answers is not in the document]
naokiyokoyama: Also, the shortest path follower dependency was softened, as it also requires habsim. I needed to soften this when using hablab without sim for igibson challenge stuff

habitat_baselines/utils/common.py
naokiyokoyama: Yep, you're right, changed it to activate the std instead

habitat_baselines/utils/common.py
naokiyokoyama: Done

habitat_baselines/config/default.py
```diff
@@ -86,6 +86,15 @@
 # -----------------------------------------------------------------------------
 _C.RL.POLICY = CN()
 _C.RL.POLICY.name = "PointNavResNetPolicy"
+_C.RL.POLICY.action_distribution_type = "categorical"  # or 'gaussian'
+# For gaussian action distribution:
+_C.RL.POLICY.use_log_std = False
```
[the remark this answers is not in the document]
naokiyokoyama: Added to subfield

habitat/tasks/nav/nav.py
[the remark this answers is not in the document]
naokiyokoyama: Softened

habitat_baselines/common/rollout_storage.py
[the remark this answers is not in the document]
naokiyokoyama: action_shape is now fed as a tuple

habitat_baselines/rl/ppo/ppo_trainer.py
naokiyokoyama: @mathfac bumping this, do you know if returning a spaces dict from one action is supported?

habitat_baselines/utils/common.py
```diff
@@ -474,3 +525,20 @@ def create_tar_archive(archive_path: str, dataset_path: str) -> None:
...
+    action: torch.Tensor,
+) -> Union[int, str, Dict[str, Any]]:
+    lin_vel, ang_vel = torch.clip(action, min=-1, max=1)
+    step_action = {
+        "action": {
+            "action": "VELOCITY_CONTROL",
+            "action_args": {
+                "linear_velocity": lin_vel.item(),
+                "angular_velocity": ang_vel.item(),
+                "allow_sliding": True,
```
erikwijmans, line 542: Should this be None?  That way it follows the configured value by default?

habitat_baselines/rl/ppo/ppo_trainer.py
erikwijmans: I checked, it is supported.  You can see its use in the Teleport action just above this.  Habitat baselines will need a bit of working on to get things working on its side however.

erikwijmans: That working on habitat-baselines is likely outside the scope of this PR (just took a look and doing this in a general will take a fair amount of thought).  For now, let's just return the correct action space in the action and then this can be
```
self.policy_action_space = self.envs.action_spaces[0]['VELOCITY_CONTROL']
```

habitat_baselines/utils/common.py
```diff
@@ -161,65 +211,24 @@ def batch_obs(
...
-        key=lambda name: 1
-        if isinstance(obs[name], numbers.Number)
-        else np.prod(obs[name].shape),
-        reverse=True,
-    )
-
-    for sensor_name in sensor_names:
-        for i, obs in enumerate(observations):
-            sensor = obs[sensor_name]
+    for i, obs in enumerate(observations):
```
erikwijmans, line 214: Looks like these changes are from a bad merge

habitat_baselines/common/rollout_storage.py
```diff
@@ -23,7 +24,9 @@ def __init__(
         action_space,
         recurrent_hidden_state_size,
         num_recurrent_layers=1,
+        action_shape: Optional[Tuple[int]] = None,
```
Skylion007, line 27: This isn't right. Optional[Tuple[Int, Int]] maybe? Or do you eally intend it o be a singleton tuple?

habitat_baselines/common/rollout_storage.py
```diff
@@ -54,18 +57,23 @@ def __init__(
...
             numsteps + 1, num_envs, 1
         )
-        if action_space.__class__.__name__ == "ActionSpace":
-            action_shape = 1
-        else:
-            action_shape = action_space.shape[0]
+
+        if action_shape is None:
+            if action_space.__class__.__name__ == "ActionSpace":
+                action_shape = (1,)
```
Skylion007, line 63: Why a singleton Tuple out of curiosity?

habitat_baselines/rl/ppo/ppo_trainer.py
naokiyokoyama: @erikwijmans Could you clarify what you mean by "return the correct action space in the action"?

PR #625, opened 2021-04-14
[Baselines] Overlapped CPU->GPU and sensor batching

erikwijmans:
## Motivation and Context

We can overlap the CPU->GPU transfer and batching when there are multiple sensors, so we should!

## How Has This Been Tested

Made sure that `non_blocking=True` is working correctly.

## Types of changes

### DIFF
--- a/habitat_baselines/utils/common.py
+++ b/habitat_baselines/utils/common.py
@@ -159,9 +159,10 @@ def batch_obs(
         batch: DefaultDict[str, List] = defaultdict(list)
 
     # Order sensors by size, stack and move the largest first
-    sensor_names = list(observations[0].keys())
-    sensor_names.sort(
-        key=lambda name: np.prod(observations[0][name].shape), reverse=True
+    sensor_names = sorted(
+        observations[0].keys(),
+        key=lambda name: np.prod(observations[0][name].shape),
+        reverse=True,
     )
 
     for sensor_name in sensor_names:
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

habitat_baselines/utils/common.py
```diff
@@ -158,10 +159,13 @@ def batch_obs(
...
 
+    obs = observations[0]
     # Order sensors by size, stack and move the largest first
     sensor_names = sorted(
-        observations[0].keys(),
-        key=lambda name: np.prod(observations[0][name].shape),
+        obs.keys(),
+        key=lambda name: 1
+        if isinstance(obs[name], numbers.Number)
+        else np.prod(obs[name].shape),
```
Skylion007, line 168: Technically, we could have return values that aren't a shape though (like a string). Shouldn't we do an hasattr check?

habitat_baselines/utils/common.py
```diff
@@ -158,10 +159,13 @@ def batch_obs(
...
         batch: DefaultDict[str, List] = defaultdict(list)
 
+    obs = observations[0]
     # Order sensors by size, stack and move the largest first
     sensor_names = sorted(
-        observations[0].keys(),
-        key=lambda name: np.prod(observations[0][name].shape),
+        obs.keys(),
+        key=lambda name: 1
+        if isinstance(obs[name], numbers.Number)
```
Skylion007, line 167: This check should just be name is not hasattr(obs[name], 'shape'), right?

erikwijmans: I thought about this but wasn't convinced that 1 as the number of elements was a sane default for anything that doesn't have a shape

PR #653, opened 2021-06-03
[Baselines] Use Numpy for CPU Batching

erikwijmans:
## Motivation and Context

Pytorch indexing is slow because it needs to track things for autograd even in no_grad mode, thereby making numpy faster for indexing and batching of obs.

## How Has This Been Tested

With a little test script.  This is about 2x faster sensors with normal Depth and GPS+Compass and becomes more beneficial as more little sensors are added.

## Types of changes

### DIFF
--- a/habitat_baselines/utils/common.py
+++ b/habitat_baselines/utils/common.py
@@ -204,10 +204,10 @@ def batch_obs(
         if cache is not None:
             # If we were using a numpy array to do indexing and copying,
             # convert back to torch tensor
-            # We know that batch_t[sensor] is either an np.ndarray
+            # We know that batch_t[sensor_name] is either an np.ndarray
             # or a torch.Tensor, so this is faster than torch.as_tensor
-            if isinstance(batch_t[sensor], np.ndarray):
-                batch_t[sensor] = torch.from_numpy(batch_t[sensor])
+            if isinstance(batch_t[sensor_name], np.ndarray):
+                batch_t[sensor_name] = torch.from_numpy(batch_t[sensor_name])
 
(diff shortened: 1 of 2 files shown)


Skylion007 (approved):

habitat_baselines/utils/common.py
```diff
@@ -171,21 +173,42 @@ def batch_obs(
...
+                        torch.as_tensor(sensor),
+                        device,
                     )
 
-                batch_t[sensor_name][i].copy_(sensor)
+                # Use isinstance(sensor, np.ndarray) here instead of
+                # np.asarray as this is quickier for the more common
+                # path of sensor being an np.ndarray
+                # np.asarray is ~3x slower than checking
+                if isinstance(sensor, np.ndarray):
```
Skylion007, line 192: @erikwijmans I think I found a better solution here: np.asanyarray https://stackoverflow.com/a/56805220/2444240

erikwijmans: That's fixing a different problem, `np.asanyarray` slower than `np.asarray` by a very small but still measurable amount in my testing.

Skylion007: Still I think we should replace the asarray at the bottom with asanyarray.

erikwijmans: Why?  asanyarray just lets subclasses of `np.ndarray` pass through, but we know sensor cannot be a subclass of `np.ndarray` if that if `isinstance(sensor, np.ndarray)` returned false.

Skylion007:
Oh yeah, good point. NVM.

PR #660, opened 2021-06-28
The Pick Task: a Base Case of Rearrangement

mathfac:
## Motivation and Context

**Task Definition**
Agent (a Fetch robot) is spawned close to a receptacle (a table) that holds multiple objects (e.g. cracker box, bowl). The task for the robot is to pick up a target object with center-of-mass coordinates s_0 (provided in robot’s coordinate system) as efficiently as possible without excessive collisions. We aim for systematic generalization to new clutter layout on the receptacle, to new objects, and to new receptacles.

Fetch is a wheeled base with a 7 degree of freedom arm manipulator and a parallel-jaw gripper, equipped with two RGBD cameras (128x128 resolution and 90 degree field of view) mounted on its head and arm. The agent can sense its proprioceptive-state in the form of arm joint angles (7-dim), end-effector position (3-dim), and base-egomotion (6-dim).

**Cameras attached to the robot**
Action space: gross motor control
The agent policy performs end-effector control at 30Hz. Specifically, it outputs the desired change in end-effector position (delta x, delta y, delta z), which is transformed into joint-motor positions by robot. In the current Pick task, the base is fixed, but in Home Assistant Benchmarks, the agent controls the base by emitting linear and angular velocities ( 𝑣,𝑤 ).

**Abstracted grasping**
We deliberately focus on gross motor control (the arm) and abstract away the challenges in fine motor control (the gripper). Specifically, the agent has a discrete grasp action that snaps the object closest to the robot's end-effector into its parallel-jaw gripper. Importantly, this grasp action is only available once the end-effector is within 15 cm of the object; at other times, it has no consequence. If an object is currently held in the gripper, calling the grasp action releases it and the object is physically simulated as falling.
To be clear, Home Assistant Benchmark fully supports rigid-body physics-based gripping; the abstract grasping is entirely a task-level simplification that may easily be removed in the future.

The PR is co-authored with @ASzot.

## How Has This Been Tested

All Rearrangment experience can be launched with few commands (data download will be automatic):
```
conda install habitat-sim withbullet headless -c conda-forge -c aihabitat
git clone --branch stable https://github.com/facebookresearch/habitat-lab.git; cd habitat-lab; pip install -e .
python examples/example.py
```

`pytest test/test_hab_rearrange_task.py::test_rearrange_habitat_env`

We adding more tests.

## Types of changes

## Checklist


habitat/tasks/rearrange/envs/hab_simulator.py
```diff
@@ -0,0 +1,846 @@
...
+
+            self._prev_sim_obs = self.get_sensor_observations()
+            # self._sim_obs_queue.put(self._sensor_suite.get_observations(self._prev_sim_obs))
+            # obs = self._sim_obs_queue.get()
+            obs = self._sensor_suite.get_observations(self._prev_sim_obs)
+
+        else:
+            self._prev_sim_obs = (
+                self.get_sensor_observations_async_start()
+            )
```
erikwijmans, line 724:
```suggestion
```

habitat/tasks/rearrange/envs/hab_simulator.py
```diff
@@ -0,0 +1,846 @@
...
+
+            if self.habitat_config.get("STEP_PHYSICS", True):
+                for i in range(self.ac_freq_ratio):
+                    self.internal_step(-1)
```
erikwijmans, line 728:
```suggestion
                    if i == 0:
                        self.start_async_render_and_step_physics(-1)
                    else:
                        self.internal_step(-1)
```

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+    if np.linalg.norm(y) != 0:
+        y_norm = y / np.linalg.norm(y)
+    else:
+        y_norm = y
+    return np.arccos(np.clip(np.dot(x_norm, y_norm), -1, 1))
+
+
+def make_render_only(obj_idx, sim):
+    if hasattr(MotionType, "RENDER_ONLY"):
```
erikwijmans, line 42: This is probably for backwards compatability?

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+            matches = get_collision_matches(
+                "id %i" % snapped_obj_id, colls, "link"
+            )
+            for match in matches:
+                if coll_name(match, "fetch"):
+                    continue
+                obj_scene_colls += 1
+
+    total_colls = robo_obj_colls + robo_scene_colls + obj_scene_colls
+    return total_colls > 0, CollDetails(
```
erikwijmans, line 237: I'd prefer an optional instead of `bool, thing`

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+def get_nav_mesh_settings_from_height(height):
+    navmesh_settings = NavMeshSettings()
+    navmesh_settings.set_defaults()
+    navmesh_settings.agent_radius = 0.4
+    navmesh_settings.agent_height = height
+    navmesh_settings.agent_max_climb = 0.05
+    return navmesh_settings
+
+
+def convert_legacy_cfg(obj_list):
```
erikwijmans, line 257: Are these cfgs in question here ones that we control?  If so we should just ship the converted ones.

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+    return rot
+
+
+def allowed_region_to_bb(allowed_region):
+    if len(allowed_region) == 0:
+        return allowed_region
+    return mn.Range2D(allowed_region[0], allowed_region[1])
+
+
+def recover_nav_island_point(v, ref_v, sim):
```
erikwijmans, line 314: Let's remove this function.  It isn't used anywhere and it is kinda a landmine.

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+    # Go through the closest points until one has the same island radius.
+    for i in ordered_idxs:
+        if nav_vs_r[i] == ref_r:
+            return nav_vs[i]
+    print("Could not find point off of island")
+    return v
+
+
+def get_largest_island_point(sim, height_thresh=None, z_min=None):
```
erikwijmans, line 339: Same comment about this function.

habitat/tasks/rearrange/rearrange_pick_task.py
```diff
@@ -0,0 +1,392 @@
...
+            str(mtime)
+            + dataset.config.SPLIT
+            + str(self._config.COUNT_OBJ_COLLISIONS)
+        )
+        cache_name += str(self._config.BASE_NOISE)
+        cache_name = cache_name.replace(".", "_")
+
+        fname = data_path.split("/")[-1].split(".")[0]
+
+        self.cache = CacheHelper(
```
erikwijmans, line 105: How does this cache work when you are doing multiprocess training?  Won't all the processes just end up over writing one another?

habitat/tasks/rearrange/envs/utils.py
```diff
@@ -0,0 +1,427 @@
...
+                time.sleep(1.0 + np.random.uniform(0.0, 1.0))
+                return self.load(load_depth + 1)
+            return self.def_val
+        else:
+            return self.def_val
+
+    def save(self, val):
+        with open(self.cache_id, "wb") as f:
+            if self.verbose:
+                print("Saving cache @", self.cache_id)
```
erikwijmans, line 413: Use logger

habitat/tasks/rearrange/rearrange_pick_task.py
erikwijmans: Also, should we just ship whatever this cache caches instead of having user generate this themselves?

habitat/tasks/rearrange/rearrange_pick_task.py
```diff
@@ -0,0 +1,392 @@
...
+            "start_pos", cache_name, {}, verbose=True, rel_dir=fname
```
erikwijmans, line 106:
```suggestion
            "start_pos", cache_name, {}, verbose=False, rel_dir=fname
```
I'd prefer it to be not verbose unless uses really need to know that something was saved

habitat/tasks/rearrange/rearrange_pick_task.py
```diff
@@ -0,0 +1,392 @@
...
+            if not did_collide:
+                break
+        if attempt == timeout - 1:
+            if not is_easy_init:
+                # print('Could not satisfy for %s! Trying with easy init' % sim.ep_info['scene_config_path'])
+                start_pos, angle_to_obj, sel_idx = self._gen_start_pos(
+                    sim, True
+                )
+            else:
+                print("collided", did_collide, details)
```
erikwijmans, line 274: Use logger

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+        return "3rd_depth"
+
+
+@registry.register_sensor
+class TargetPointGoalGPSAndCompassSensor(PointGoalSensor):
+    cls_uuid: str = "target_point_goal_gps_and_compass_sensor"
+
+    def __init__(self, task, *args, **kwargs):
+        self._task = task
+        super(TargetPointGoalGPSAndCompassSensor, self).__init__(
```
erikwijmans, line 61:
```suggestion
        super().__init__(
```

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+class DynObjPosStartOrGoal(MultiObjSensor):
+    """
+    Returns the goal position if the robot is holding an object and the object
+    starting position if the robot is not holding the object.
+    """
+
+    cls_uuid: str = "dyn_obj_start_or_goal_sensor"
+
+    def get_observation(self, observations, episode, *args, **kwargs):
+        global_T = self._sim.robot.ee_transform
```
erikwijmans, line 183: Based on the way this is being used, I think this is the `transform_world_ee` as it transforms a point from the `ee` coordinate frame to world coordinates.  Then `T_inv` is `transform_ee_world`.  Assuming that is correct, please rename

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+            dtype=np.float32,
+        )
+
+    def get_observation(self, observations, episode, *args, **kwargs):
+        pos = self._sim.get_target_objs_start()
+        return pos
+
+
+@registry.register_sensor
+class DynObjPosStartOrGoal(MultiObjSensor):
```
erikwijmans, line 174: Drop the `Dyn`.  This similar to the discussion we had around pointnav, and we decided on dropping `dynamic` in favor of specifying that that setting requires the robot to have `GPS+Compass` and in this case you need that plus precise exact proprioception

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+    cls_uuid: str = "dyn_obj_start_or_goal_sensor"
+
+    def get_observation(self, observations, episode, *args, **kwargs):
+        global_T = self._sim.robot.ee_transform
+        T_inv = global_T.inverted()
+
+        if self._sim.snapped_obj_id is not None:
+            _, pos = self._sim.get_targets()
+        else:
+            pos = self._sim.get_target_objs_start()
```
erikwijmans, line 189: Seems like `get_targets` and co should be pushed up to the task level

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+class AbsGoalSensor(MultiObjSensor):
+    cls_uuid: str = "abs_obj_goal_sensor"
+
+    def get_observation(self, observations, episode, *args, **kwargs):
+        _, pos = self._sim.get_targets()
+        return pos
+
+
+@registry.register_sensor
+class DummySensor(Sensor):
```
erikwijmans, line 224: What is this sensor for?

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+        return spaces.Box(
+            shape=(4,),
+            low=np.finfo(np.float32).min,
+            high=np.finfo(np.float32).max,
+            dtype=np.float32,
+        )
+
+    def get_observation(self, observations, episode, *args, **kwargs):
+        trans = self._sim.robot.sim_obj.transformation
+        forward = np.array([1.0, 0, 0])
```
erikwijmans, line 265: The coordinate frame in habitat is that `-Z` is forward, why is this being done differently here?

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+        heading = np.array(trans.transform_vector(forward))
+        forward = forward[[0, 2]]
+        heading = heading[[0, 2]]
+
+        heading_angle = get_angle(forward, heading)
+        c = np.cross(forward, heading) < 0
+        if not c:
+            heading_angle = -1.0 * heading_angle
```
erikwijmans, line 273: Why multiple by -1?  Won't this make this always positive?

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,618 @@
...
+
+        reward += self._get_coll_reward()
+
+        self.prev_picked = cur_picked
+
+        self._metric = reward
+
+
+@registry.register_measure
+class RearrangePickSuccess(Measure):
```
erikwijmans, line 570: I'm not opposed to Reward as a measure (although this may encourage people to report reward in papers which I don't love), but we should consider having Rewards be their own thing

habitat/tasks/utils.py
```diff
@@ -46,3 +46,19 @@ def compute_pixel_coverage(instance_seg, object_id):
     cand_mask = instance_seg == object_id
     score = cand_mask.sum().astype(np.float64) / cand_mask.size
     return score
+
+
+def get_angle(x, y):
```
erikwijmans, line 51: Seems like this function is duplicated a bunch

erikwijmans (commented):
Initial set of comments

habitat/tasks/rearrange/rearrange_sensors.py
mathfac: Yes, Reward shouldn't be part of the task definition and it's not. It just a vehicle to deliver calculation that should have type Reward in some point,

habitat/tasks/rearrange/rearrange_sensors.py
mathfac: @ASzot

erikwijmans: There's also another instance of forward being defined like this in `hab_simulator.py`

habitat/tasks/rearrange/rearrange_sensors.py
erikwijmans: Yeah, I am not saying we should stick the task, but a Measure might not be the right place either since those are normally evaluation metrics, which a reward isn't

habitat/tasks/rearrange/rearrange_pick_task.py
mathfac: We are not releasing the cache for now, as it has system dependent paths and etc. In follow up PRs, cache will be part of the episode dataset.

habitat/tasks/rearrange/rearrange_sensors.py
mathfac: Yes, we plan to make RearrangeSim as tiny as possibly or remove in follow up PRs.

habitat/tasks/rearrange/rearrange_sensors.py
mathfac: @ASzot

habitat/datasets/rearrange/rearrange_dataset.py
```diff
@@ -0,0 +1,55 @@
...
+from habitat.config import Config
+from habitat.core.dataset import Episode
+from habitat.core.registry import registry
+from habitat.core.utils import DatasetFloatJSONEncoder
+from habitat.datasets.pointnav.pointnav_dataset import PointNavDatasetV1
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class RearrangeEpisode(Episode):
+    art_objs: object
```
Skylion007, line 21: we realy dont have any better typing fo these objects?

habitat/tasks/rearrange/actions.py
```diff
@@ -0,0 +1,230 @@
...
+            closest_obj_pos = scene_obj_pos[closest_obj_idx]
+            to_target = np.linalg.norm(ee_pos - closest_obj_pos, ord=2)
+
+            if to_target < self.thresh_dist:
+                self._sim.set_snapped_obj(closest_obj_idx)
+
+        # Get the marker the EE is closest to.
+        marker_name_pos = self._sim.get_marker_positions()
+        if self.snap_markers and len(marker_name_pos) > 0:
+            marker_name, marker_pos = zip(*marker_name_pos.items())
```
Skylion007, line 87: No need   to zip through items. .keys() and .values should have same ordering

habitat/tasks/rearrange/rearrange_pick_task.py
Skylion007: AN easy fix ot expose a worker number.

habitat/tasks/rearrange/rearrange_pick_task.py
```diff
@@ -0,0 +1,240 @@
...
+        return start_pos, angle_to_obj, sel_idx
+
+    def _should_prevent_grip(self, action_args):
+        return (
+            self._is_holding_obj()
+            and action_args.get("grip_ac", None) is not None
+            and action_args["grip_ac"] <= 0
+        )
+
+    def step(self, action, episode):
```
Skylion007, line 170: Type Annotations

habitat/tasks/rearrange/rearrange_pick_task.py
```diff
@@ -0,0 +1,240 @@
...
+
+        if self._should_prevent_grip(action_args):
+            # No releasing the object once it is held.
+            action_args["grip_ac"] = None
+        obs = super().step(action=action, episode=episode)
+        obs = self._trans_obs(obs)
+
+        return obs
+
+    def _trans_obs(self, obs):
```
Skylion007, line 189: This is bad code smell in general. ANd this is a pattern we should better extract.

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -0,0 +1,601 @@
...
+
+
+@registry.register_sensor
+class ArmDepthSensor(HabitatSimDepthSensor):
+    def _get_uuid(self, *args, **kwargs):
+        return "arm_depth"
+
+
+@registry.register_sensor
+class ThirdRgbSensor(HabitatSimRGBSensor):
```
Skylion007, line 51: This is really unnecessary. UUIDs are defined in the configs. These sensor can be defined from raw onfigs and don't warrant a sensor type

Skylion007 (changes requested):
Changes Requested. Big issue are the useless sensor classes.

test/test_rearrange_task.py
```diff
@@ -0,0 +1,163 @@
...
+    ), "Intersection of split datasets is not the empty set"
+
+
+def test_rearrange_habitat_env():
+    config = get_config(CFG_TEST)
+
+    if not RearrangeDatasetV0.check_config_paths_exist(config.DATASET):
+        pytest.skip("Test skipped as dataset files are missing.")
+
+    config.defrost()
```
Skylion007, line 117: Useless remove.

configs/tasks/rearrang_pick_task.yaml
```diff
@@ -0,0 +1,156 @@
+ENVIRONMENT:
```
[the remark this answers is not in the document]
mathfac: Removed one of them.

habitat/tasks/rearrange/actions.py
```diff
@@ -0,0 +1,230 @@
...
+        return {
+            "forces": forces,
+            "vel": vel,
+            "pos": art_pos,
+        }
+
+    def _set_robo_state(self, robot_id, sim: RearrangeSim, set_dat):
+        sim.set_articulated_object_forces(robot_id, set_dat["forces"])
+        sim.set_articulated_object_velocities(robot_id, set_dat["vel"])
+        sim.set_articulated_object_positions(robot_id, set_dat["pos"])
```
aclegg3, line 163: These sim calls are unused? This API was not carried forward past the prototype into Habitat-sim v0.2.0, so this will fail.

habitat/tasks/rearrange/obj_loaders.py
```diff
@@ -0,0 +1,148 @@
...
+        motion_type = MotionType.DYNAMIC
+        if obj_type == -2:
+            fixed_base = False
+        else:
+            fixed_base = True
+        if len(obj_ids) == 0:
+            ao_mgr = sim.get_articulated_object_manager()
+            ao = ao_mgr.add_articulated_object_from_urdf(name, fixed_base)
+        else:
+            ao = obj_ids[i]
```
aclegg3, line 86: This won't work if these are actually ids. ao is expected to be a wrapper object.

habitat/tasks/rearrange/obj_loaders.py
```diff
@@ -0,0 +1,148 @@
...
+        # if auto_sleep:
+        #    ao.can_sleep = True
+        ao.motion_type = motion_type
+        art_obj_ids.append(ao)
+    if len(obj_ids) != 0:
+        return obj_ids
+    return art_obj_ids
+
+
+def init_art_objs(idx_and_state, sim, auto_clamp=False):
```
aclegg3, line 100: This function is not updated to new API and won't work.

habitat/tasks/rearrange/obj_loaders.py
```diff
@@ -0,0 +1,148 @@
...
+        obj_type = obj_dat[1]
+
+        use_trans = mn.Matrix4(trans)
+        sim.set_transformation(use_trans, obj_id)
+        sim.set_linear_velocity(mn.Vector3(0, 0, 0), obj_id)
+        sim.set_angular_velocity(mn.Vector3(0, 0, 0), obj_id)
+        sim.set_object_motion_type(MotionType(obj_type), obj_id)
+        # TODO: Broken in Hab2.0 release
+        # if auto_sleep:
+        #    sim.set_object_sleep(obj_id, True)
```
aclegg3, line 144: Should be using the new API. Sleeping would be: `obj.awake = False`

habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -0,0 +1,776 @@
...
+        "kitchen_counter.urdf",
+        [0.28, 0.5, 0.75],
+        [-0.067, 0.5, -0.7],
+        [0.524, -0.896],
+    ),
+]
+
+
+# temp workflow for loading lights into Habitat scene
+def load_light_setup_for_glb(json_filepath):
```
aclegg3, line 63: Should be encoded in the SceneDataset for ReplicaCAD or in general the LightAttributeManager can do this for you  and register the setup in Simulator.

habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -0,0 +1,776 @@
...
+
+        # Disables arm control. Useful if you are hiding the arm to perform
+        # some scene sensing.
+        self.ctrl_arm = True
+
+        self._light_setup = load_light_setup_for_glb(
+            "data/misc_data/frl_apartment_stage_pvizplan_empty_lights.json"
+        )
+        obj_attr_mgr = self.get_object_template_manager()
+        obj_attr_mgr.load_configs("data/objects")
```
aclegg3, line 145: Probably need to update this path.

habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -0,0 +1,776 @@
...
+        - set_back: (bool): Pushes the robot even further back if specified by
+          offset. Used when spawning the robot in front of an open cabinet.
+        """
+        pos = mn.Vector3(*pos)
+        force_spawn_pos = self.ep_info["force_spawn_pos"]
+        if set_back and force_spawn_pos is not None:
+            rel_art_bb_id, offset = force_spawn_pos
+            _, urdf_name, _, _, robo_pos = ART_BBS[rel_art_bb_id]
+            art_id = self.art_name_to_id[urdf_name]
+            art_T = self._sim.get_articulated_object_root_state(art_id)
```
aclegg3, line 215: API update needed.

habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -0,0 +1,776 @@
...
+        if len(self.snapped_obj_constraint_id) != 0:
+            self.desnap_object()
+
+        self.snapped_obj_id = use_snap_obj_id
+
+        max_impulse = 1000.0
+        if self.snap_to_link_on_grab:
+            # Set collision group to GraspedObject so that it doesn't collide
+            # with the links of the robot.
+            grasped_object_group = 16  # see Habitat-sim CollisionGroupHelper.h
```
aclegg3, line 543: The default groups have changed somewhat, be sure you have these configured correctly.

habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -0,0 +1,776 @@
...
+            if not self.do_grab_using_constraint:
+                return
+
+            # Set the transformation to be in the robot's hand already.
+            ee_T = self.get_end_effector_trans()
+            obj_local_T = mn.Matrix4.translation(EE_GRIPPER_OFFSET)
+            global_T = ee_T @ obj_local_T
+            self.set_transformation(global_T, use_snap_obj_id)
+
+            def create_hold_constraint(pivot_in_link, pivot_in_obj):
```
aclegg3, line 556: Constraint API has changed, these will need to be updated.

habitat/tasks/rearrange/actions.py
```diff
@@ -0,0 +1,219 @@
...
+from habitat.tasks.rearrange.envs.utils import rearrang_collision
+
+
+@registry.register_task_action
+class ArmAction(SimulatorTaskAction):
+    """An arm control and grip control into one action space."""
+
+    def __init__(self, *args, config, sim, **kwargs):
+        super().__init__(*args, config=config, sim=sim, **kwargs)
+        arm_controller_cls = eval(self._config.ARM_CONTROLLER)
```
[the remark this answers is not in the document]
mathfac: Will be addressed in follow up PRs.

aclegg3 (approved):
This needs a lot of polish. This review is high level, mostly targeting sim API inconsistencies and code which won't run with current sim master. Approving for release, but requires follow-up.

habitat/tasks/rearrange/rearrange_pick_task.py
mathfac: refactored and removed.

mathfac:
Rearrangment tests are now launching on CI using data download: 
Screen Shot 2021-06-30 at 2 45 07 AM

mathfac:
All Rearrangment experience can be launched with few commands (data download will be automatic):
```
conda install habitat-sim withbullet headless -c conda-forge -c aihabitat
git clone --branch stable https://github.com/facebookresearch/habitat-lab.git; cd habitat-lab; pip install -e .
python examples/example.py
```

habitat/tasks/rearrange/rearrange_sensors.py
ASzot: Yes, I think I had changed the coordinate system in a config file in my experiments, however, this is not correct in this branch. I am working on fixing this.

habitat/tasks/rearrange/rearrange_sensors.py
ASzot: Yes, this sensor should actually be deprecated. It's no longer used. I will work on this as I merge back in the navigation actions.

habitat/tasks/rearrange/rearrange_sim.py
ASzot: Yes, working on this now for another PR.

PR #667, opened 2021-06-30
Fix rearrang typos

erikwijmans:
## Motivation and Context

Rearrang should be rearrange

## How Has This Been Tested

Visually


mathfac (approved):
Thank you!

PR #668, opened 2021-07-01
Fixed automatic data download outside of git repo

mathfac:
## Motivation and Context
Data downloader works without explicit data-path only inside of git repo. To support usage with PyPi package setting data-path.

## How Has This Been Tested
CI

## Types of changes

## Checklist

### DIFF
--- a/habitat/datasets/rearrange/rearrange_dataset.py
+++ b/habitat/datasets/rearrange/rearrange_dataset.py
@@ -49,7 +49,9 @@ class RearrangeDatasetV0(PointNavDatasetV1):
             print(
                 "Rearrange task assets are not downloaded locally, downloading and extracting now..."
             )
-            data_downloader.main(["--uids", "rearrange_task_assets"])
+            data_downloader.main(
+                ["--uids", "rearrange_task_assets", "--data-path", "data/"]
+            )
             print("Downloaded and extracted the data.")
         check_and_gen_physics_config()
 


erikwijmans (commented):
This seems fine for now, but if we want to install things download as part of a package, `<current_dir>/data` probably isn't the right thing as things will get re-downloaded if your current directory changes.

mathfac:
@erikwijmans, maybe have environment variable with data folder path, something like `HABITAT_DATA_PATH`. If that isn't set download to current dir and set the variable.

erikwijmans:
Setting the environment variable won't work as you can't set environment variables for your parent process AFAIK. 

In general I've seen things like `$HOME/.habitat/data` be used, I think MMF does that and I'm pretty sure torchvision does that. I don't love that as it's kinda hard to figure out that's it's doing that and a pain to direct it elsewhere (even if it's configurable it's easy to forget to change it in some config), but I get why it's done.

Skylion007:
> Setting the environment variable won't work as you can't set environment variables for your parent process AFAIK.
> 
> In general I've seen things like `$HOME/.habitat/data` be used, I think MMF does that and I'm pretty sure torchvision does that. I don't love that as it's kinda hard to figure out that's it's doing that and a pain to direct it elsewhere (even if it's configurable it's easy to forget to change it in some config), but I get why it's done.

Agree with @erikwijmans . This is the standard with an ENV variable usually be used to optionally override the location (if your home folder has limited space for instance). TorchVision also does this for example.

Skylion007:
Is this intended to fix the broken CI we have for Habitat-Sim. Or would that require an additional change?

mathfac:
When habitat-lab is pip installed there is no git repo. We need specify the data folder otherwise example with RearrangeTask is throwing the error.

mathfac:
@erikwijmans , @Skylion007 I like the idea with `$HOME/.habitat/data`, but that will require deeper refactoring as we everywhere in configs assume that "<current_dir>/data" is the directory with all the data.  This change may not be needed after https://github.com/facebookresearch/habitat-sim/pull/1369 lands.

erikwijmans:
It's deeper than a 1 line fix, but it isn't that deep and I do think it's worth doing.  If the goal here is to make habitat-lab a package that people can install and plays nicely with other RL training libraries, I think we should have the default be a stable path.  Otherwise users will end up redownloading data or will need to make a trail of symlinks.

PR #680, opened 2021-07-09
Fix CI

erikwijmans:
## Motivation and Context

Don't prompt for replacement on the auto download.
## How Has This Been Tested

Fixes the CI


eundersander (approved):
LGTM

PR #725, opened 2021-09-16
Fix failing gym 2.0 test

erikwijmans:
## Motivation and Context

Gym 2.0 makes `contains` much stricter so now types need to be castable and it doesn't like casting from float64 to float32.

## How Has This Been Tested

Via the test.


Skylion007 (approved):

erikwijmans:
Landing this as the CI test that failed is expected to failure currently.

PR #734, opened 2021-09-24
Migrate ci resource type in advance of deprecation

aclegg3:
## Motivation and Context

CircleCI is deprecating legacy resource types (e.g. gpu.small) by 2022. We should migrate to the new resource types.

## How Has This Been Tested

CI jobs running with new resource type.


aclegg3:
covered by #807

PR #764, opened 2021-11-24
Episode Creator Script for Habitat 2.0

ASzot:
## Motivation and Context

New episode creator and ReplicaCAD integration code for Habitat 2.0. Also new Gym API and upgrades to the interactive play script. These are the core functionalities I need to test things are working. 

New components listed in detail below. 

Download Rearrangment assets:
`python -m habitat_sim.utils.datasets_download --uids rearrange_task_assets`

### Episode Generation
Running `python habitat/datasets/rearrange/rearrange_generator.py --run` 
will generate a single episode with default config and save it locally as `rearrange_ep_dataset.json.gz`.
Use:
- `--config` to pass a custom generator config file. **(See example config: habitat/datasets/rearrange/configs/test_config.yaml)**
- `--num-episodes` to specify the number of episodes to generate.
- `--out` to specify the output location for the serialized dataset.
- `--debug` to also generate and save debugging images/videos.
- `--db-output` to specify the output location for debugging images/videos (default is `rearrange_ep_gen_output/`).

**NOTE:** `python habitat/datasets/rearrange/rearrange_generator.py --list` will print concise list of scene, object, and receptacle handles for reference when authoring a custom generator config file.

### Gym API
Create Habitat environments from Gym by first `import habitat_baselines.utils.gym_definitions` and then `env = gym.make("EnvironmentID")` the environment IDs are listed below:

* `HabitatGymPick-v0`
* `HabitatGymPickState-v0`
* `HabitatGymReachState-v0`
* `HabitatGymOpenCabState-v0`
* `HabitatGymCloseCabState-v0`

You can also create a "render" version of the environment which has a high-resolution 3rd person camera to show what the robot with `env.render()` (but this is slower!). Do this using prefix `HabitatGymRender`, for example `HabitatGymRenderPick-v0` instead of `HabitatGymPick-v0`. 

Pass overrides to the task with the `override_options` named argument in `gym.make`. For example `gym.make("HabitatGymPick-v0", override_options=["TASK_CONFIG.TASK.ACTIONS.ARM_ACTION.GRIP_CONTROLLER", "SuctionGraspAction"])`.

### Interactive Play Script
Example to run
```
python examples/interactive_play.py --cfg configs/tasks/rearrange/composite.yaml --play-task --never-end --add-ik
```
Some useful arguments:

- `--never-end` if specified the episode will never end due to a time out. 
- `--add-ik` will change the arm action space to IK control. Note you need
    PyBullet installed for this to work. 
- `--play-task` will add the visualization camera to the task so you can
    easily view what's going on in the play window. 
- Any config argument can be changed by specifying additional arguments. 

## How Has This Been Tested

Added new tests:
* RL Environments tested in `test/test_rearrange_task.py::test_rearrange_baseline_envs`
* Tasks tested in `test/test_rearrange_task.py::test_rearrange_tasks`
* Interactive play script: `test/test_examples.py::test_rearrange_example_modules`
* Gym wrappers: `test/test_gym_wrapper.py`

## Types of changes

## Checklist


habitat/datasets/rearrange/generate_episode_inits.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+
+from tqdm import tqdm
```
Skylion007, line 9:
```suggestion
from tqdm import trange
```

habitat/datasets/rearrange/generate_episode_inits.py
```diff
@@ -0,0 +1,32 @@
...
+
+import habitat
+
+
+def generate_inits(cfg_path, opts):
+    config = habitat.get_config(cfg_path, opts)
+    with habitat.Env(config=config) as env:
+        for _ in tqdm(range(env.number_of_episodes)):
```
Skylion007, line 17:
```suggestion
        for _ in trange(env.number_of_episodes):
```

habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -0,0 +1,1015 @@
...
+        type=int,
+        default=1,
+        help="The number of episodes to generate.",
+    )
+
+    args, _ = parser.parse_known_args()
+
+    # merge the configuration from file with the default
+    cfg = get_config_defaults()
+    print(f"\n\nOriginal Config:\n{cfg}")
```
Skylion007, line 909: The should all use the Habitat Core Logger

ASzot:
Thanks @Skylion007! All your comments should be addressed now.

habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -20,87 +20,155 @@ class RearrangeGraspManager:
...
         self._snap_constraints = []
         self._leave_info = None
         self._config = config
 
-    def reset(self):
-        self._leave_info = None
+    def reconfigure(self):
+        for constraint_id in self._snap_constraints:
+            self._sim.remove_rigid_constraint(constraint_id)
+        self._snap_constraints = []
```
Skylion007, line 31:
```suggestion
        self._snap_constraints.clear()
```

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -219,36 +200,28 @@ def get_observation(self, observations, episode, *args, **kwargs):
...
         )
 
     def get_observation(self, observations, episode, *args, **kwargs):
-        pos = self._sim.get_track_markers_pos()[0]
-
-        trans = self._sim.get_robot_transform()
-        pos = trans.inverted().transform_point(pos)
-        return np.array(pos).astype(np.float32)
+        joints_pos = self._sim.robot.arm_velocity
+        return np.array(joints_pos).astype(np.float32)
```
Skylion007, line 224:
```suggestion
        return np.array(joints_pos,dtype=np.float32)
```

habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -311,7 +284,7 @@ def get_observation(self, observations, episode, task, *args, **kwargs):
 
         relative_desired_resting = task.desired_resting - local_ee_pos
 
-        return np.array(relative_desired_resting)
+        return np.array(relative_desired_resting).astype(np.float32)
```
Skylion007, line 287:
```suggestion
        return np.array(relative_desired_resting, dtype=np.float32)
```

habitat_baselines/utils/common.py
```diff
@@ -180,9 +180,12 @@ def get(
             and device.type == "cuda"
             and cache.device.type == "cpu"
         ):
+            cache = cache.pin_memory()
```
Skylion007, line 183: Ping @erikwijmans about this part.

ASzot: Oops I have no idea how that got there, thanks for catching it, this is now removed.

ASzot:
Thanks again @Skylion007, addressed your comments.

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,838 @@
...
+import os.path as osp
+from typing import Any, Dict, List, Optional, Union
+
+import magnum as mn
+import numpy as np
+
+import habitat_sim
+
+
+class DebugVisualizer:
```
eundersander, line 16: This looks like cool functionality that could be useful outside of just rearrange dataset generation. How about make this class its own file that doesn't live under `habitat/datasets`?

And how about class documentation?

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,838 @@
...
+
+    def __init__(
+        self,
+        sim: habitat_sim.Simulator,
+        output_path: str = "visual_debug_output/",
+        default_sensor_uuid: str = "rgb",
+    ) -> None:
+        """
+        Initialize the debugger provided a Simulator and the uuid of the debug sensor.
+        NOTE: Expects the debug sensor attached to and coincident with agent 0's frame.
```
eundersander, line 30: How about assert this? Or enforce this assumption directly by setting the agent and sensor scene nodes directly?

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,838 @@
...
+            print(
+                f"DebugVisualizer: Aborting 'make_debug_video': Failed to create the output directory: {output_path}"
+            )
+            return
+        print(
+            f"DebugVisualizer: 'make_debug_video' output directory did not exist and was created: {output_path}"
+        )
+
+
+class Receptacle:
```
eundersander, line 275: A lot of receptacle logic here... how about put this in its own file?

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,838 @@
...
+    # articulated objects #TODO: merge with above
+    for obj_handle in ao_mgr.get_object_handles():
+        obj = ao_mgr.get_object_by_handle(obj_handle)
+        user_attr = obj.user_attributes
+
+        for sub_config_key in user_attr.get_subconfig_keys():
+            if sub_config_key.startswith("receptacle_"):
+                sub_config = user_attr.get_subconfig(sub_config_key)
+                # this is a receptacle, parse it
+                assert sub_config.has_value("position")
```
eundersander, line 507: This logic to parse Receptacle fields from a subconfig is mostly duplicated three times in this function. Can we consolidate?

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,838 @@
...
+        "base_rel_height": base_rel_height,
+        "surface_snap_point": surface_snap_point,
+        "raycast_results": raycast_results,
+    }
+
+
+def snap_down(
+    sim: habitat_sim.Simulator,
+    obj: habitat_sim.physics.ManagedRigidObject,
+    support_obj_ids: Optional[List[int]] = None,
```
eundersander, line 747: We should explain what these "object ids" actually are... if I understand correctly, it's an id returned by our collision queries, and it can refer to either a rigid object, and articulated object base, or an articulated object link.

habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -270,6 +270,11 @@ def __init__(self, config: Config) -> None:
         self.sim_config = self.create_sim_config(self._sensor_suite)
         self._current_scene = self.sim_config.sim_cfg.scene_id
         super().__init__(self.sim_config)
+        # load additional object paths specified by the dataset
+        # TODO: Should this be moved elsewhere?
+        obj_attr_mgr = self.get_object_template_manager()
+        for path in self.habitat_config.ADDITIONAL_OBJECT_PATHS:
```
eundersander, line 276: The need for `additional_object_paths` feels a bit hacky. Why is this needed in practice? Why can't the ReplicaCAD scene dataset include all the objects we need?

Are we going to need `additional_urdf_paths`, `additional_fairmotion_paths`, etc.  in the future?

eundersander (commented):
I've reviewed the `.py` files up to `habitat/tasks/rearrange/__init__.py` and left comments. I'll review the rest if/when I get a chance.

habitat/sims/habitat_simulator/habitat_simulator.py
aclegg3: A single SceneDataset is self-contained. In this case we are loading scenes from ReplicaCAD, but we are adding clutter objects from YCB. Since these are separate SceneDatasets and we don't want to contrive some explicit merging, we need to load multiple data sources into the same `MetaDataMediator` instance. Adding this field to the generator config is the most lightweight option. 

Your extrapolation to adding multiple fields for merging different types of data into the active SceneDataset programmatically makes sense and though it may not be pretty, it would probably still be the most lightweight option if those additional types were needed. Another option would be to simply load multiple SceneDatasets into the same MetaDataMediator. In this case we would need to swap the active SceneDataset depending on the source of each asset we attempt to load (which would be cumbersome), or merge multiple SceneDatasets into a single programmatic data-structure. Perhaps @jturner65 can weigh in, but I think the latter could be challenging due to potential issues with naming collisions.

jturner65: One possible solution would be to have an extra scene dataset configuration defining the paths to all these "extra" asset paths.  The MM is designed to support loading the metadata for multiple datasets and having them all available at one time.  The user could then just switch datasets on demand to acquire the configurations desired to instantiate the various objects they require.

habitat/core/dataset.py
```diff
@@ -55,6 +55,14 @@ class Episode:
 
     episode_id: str = attr.ib(default=None, validator=not_none_validator)
     scene_id: str = attr.ib(default=None, validator=not_none_validator)
+    # path to the SceneDataset config file
+    scene_dataset_path: str = attr.ib(
```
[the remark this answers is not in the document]
jturner65: I would say the issue here is more that the episodes should query both dataset and scene id, and not make the assumption that the dataset does not change (probably an oversight).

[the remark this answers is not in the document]
jturner65: To add to my answer above - if all the desired scene datasets are loaded at ahead of time, switching between datasets during episodes would be no more expensive than switching between scenes.

habitat/datasets/rearrange/rearrange_dataset.py
```diff
@@ -21,16 +22,16 @@
...
-    static_objs: List[List[Any]]
-    targets: List[List[Any]]
-    fixed_base: bool
-    art_states: List[Any]
-    nav_mesh_path: str
-    scene_config_path: str
-    allowed_region: List[Any] = []
-    markers: List[Dict[str, Any]] = []
-    force_spawn_pos: List = None
+    ao_states: Dict[
```
vincentpierre, line 25: For readability, I think the comments should be separate from the fields.
```
class RearrangeEpisode(Episode):
   """
   ao_states: <comment about what it is>
   """
   ao_states: Dict[str, Dict[int,float]]
```

It might be useful to create types for readability. Like:
```
InstanceHandle = str
Link = int
State = float
```
so the typing is more explicit.

habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -0,0 +1,1018 @@
...
+    RearrangeDatasetV0,
+    RearrangeEpisode,
+)
+
+
+class RearrangeEpisodeGenerator:
+    def __enter__(self) -> "RearrangeEpisodeGenerator":
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
vincentpierre, line 30: I think this method is missing typing

habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -0,0 +1,1018 @@
...
+import habitat.datasets.rearrange.sim_utilities as sutils
+import habitat_sim
+from habitat.core.logging import logger
+from habitat.datasets.rearrange.rearrange_dataset import (
+    RearrangeDatasetV0,
+    RearrangeEpisode,
+)
+
+
+class RearrangeEpisodeGenerator:
```
vincentpierre, line 26: Should RearrangeEpisodeGenerator be inheriting a generic EpisodeGenerator? 
Can you add a comment on what an EpisodeGenerator does and how to use it?

habitat/tasks/rearrange/actions.py
```diff
@@ -238,6 +214,8 @@ def step(self, set_pos, should_step=True, *args, **kwargs):
 
 @registry.register_task_action
 class BaseVelAction(SimulatorTaskAction):
+    """ """
```
vincentpierre, line 217: This comment is empty, should this line be removed or a comment added?

habitat/core/dataset.py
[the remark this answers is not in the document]
eundersander: Ok, we if plan to eventually support these fields changing per episode, then we can leave them here. But we should at least add an assert to catch users trying to use this (nonexistent) functionality. Just assert that all episodes have the same values for these fields.

habitat/core/env.py
```diff
@@ -91,7 +91,13 @@ def __init__(
             self._setup_episode_iterator()
             self.current_episode = next(self.episode_iterator)
             self._config.defrost()
-            self._config.SIMULATOR.SCENE = self.current_episode.scene_id
+            self._config.SIMULATOR.SCENE_DATASET = self._dataset.episodes[
```
mathfac, line 94: Why we decided to use episode 0 instead of Iterator next episode? This line will cause error when iterator isn't using 0 as first one.

habitat/sims/habitat_simulator/habitat_simulator.py
aclegg3: I agree that from a user perspective the cleanest option at this time is to manually create your own SceneDataset merging all the assets you want to have available.

I also agree that the most likely version of this paradigm to support the current application would be an extra SceneDataset config file which externally references assets and metadata from other datasources as @jturner65 suggests. This would allow us to avoid re-packaging and re-distributing composite datasets which could have conflicting license/use-age terms. For example, we could bundle `data_source` dependencies with the top level metadata file in a `data_group` within the `datasets_download.py` utility.

habitat/core/env.py
```diff
@@ -91,7 +91,13 @@ def __init__(
...
+                0
+            ].scene_dataset_path
+            self._config.SIMULATOR.SCENE = self._dataset.episodes[0].scene_id
```
mathfac, line 97: same here, I would use "current_episode" instead of episodes[0].

mathfac (dismissed):
@ASzot this PR looks amazing and in much mature state I have seen before. Left first portion of the review.

habitat/datasets/rearrange/generate_episode_inits.py
```diff
@@ -0,0 +1,32 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+
+from tqdm import trange
+
+import habitat
+
+
+def generate_inits(cfg_path, opts):
```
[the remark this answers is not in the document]
aclegg3: I don't think this file is used for anything, may be leftover and candidate for removal? @ASzot

habitat/datasets/rearrange/samplers.py
```diff
@@ -0,0 +1,618 @@
...
+            if new_object is not None:
+                targets_found += 1
+                found_match = False
+                for object_instance in self.object_instance_set:
+                    if (
+                        object_instance.creation_attributes.handle
+                        == new_object.creation_attributes.handle
+                        and object_instance.handle not in new_target_objects
+                    ):
+                        new_target_objects[object_instance.handle] = new_object
```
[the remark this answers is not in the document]
aclegg3: Yes, this supports multiple instances of the same object. The trick is that multiple copies of the handle exist in the `self.object_set`. When one is used, it is removed but the others are still available. When an instance is matched, it is added to `new_target_objects`, so it won't be matched again.

habitat/datasets/rearrange/samplers.py
```diff
@@ -0,0 +1,618 @@
...
+        For all matching AOs in the scene, sample and apply the joint state for this sampler.
+        Return a list of tuples (instance_handle, link_name, state)
+        """
+        ao_states: Dict[
+            habitat_sim.physics.ManagedArticulatedObject, Dict[int, float]
+        ] = {}
+        # TODO: handle sampled invalid states (e.g. fridge open into wall in some scenes)
+        aom = sim.get_articulated_object_manager()
+        # get all AOs in the scene with the configured handle as a substring
+        matching_ao_instances = aom.get_objects_by_handle_substring(
```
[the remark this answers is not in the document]
aclegg3: Good point. This loop is intended to catch multiple instances (e.g. `fridge_0000` and `fridge_0001`) but it would also currently catch `fridge2_0000` and if that object also had a link with the same name it would be modified. We should additionally cull urdf file names which don't exactly match the input.

habitat/datasets/rearrange/sim_utilities.py
aclegg3: Yeah, I meant to flesh this out a bit more. Not sure if it shouldn't live in the habitat-sim with some of the other utilities.

configs/tasks/rearrange/pick.yaml
```diff
@@ -3,17 +3,24 @@ ENVIRONMENT:
...
     TYPE: RearrangePickTask-v0
     MAX_COLLISIONS: -1.0
     COUNT_OBJ_COLLISIONS: True
     COUNT_ROBOT_OBJ_COLLS: False
+    DESIRED_RESTING_POSITION: [0.5, 0.0, 1.0]
 
     # In radians
     BASE_ANGLE_NOISE: 0.15
     BASE_NOISE: 0.05
+    CONSTRAINT_VIOLATE_SHOULD_END: True
```
vincentpierre, line 18:
```suggestion
    CONSTRAINT_VIOLATION_ENDS_EPISODE: True
```

configs/tasks/rearrange/pick.yaml
```diff
@@ -44,52 +51,52 @@ TASK:
...
     RELATIVE_RESTING_POS_SENSOR:
         TYPE: "RelativeRestingPositionSensor"
-    LOCAL_SENSOR:
-        TYPE: "LocalizationSensor"
-    SENSORS: ["TARGET_START_SENSOR", "GOAL_SENSOR", "JOINT_SENSOR", "IS_HOLDING_SENSOR", "END_EFFECTOR_SENSOR", "LOCAL_SENSOR", "RELATIVE_RESTING_POS_SENSOR"]
+    SENSORS: ["TARGET_START_SENSOR", "JOINT_SENSOR", "IS_HOLDING_SENSOR", "END_EFFECTOR_SENSOR", "RELATIVE_RESTING_POS_SENSOR"]
     ROBOT_FORCE:
         TYPE: "RobotForce"
         MIN_FORCE: 20.0
+    FORCE_TERMINATE:
```
vincentpierre, line 58: I think this one can be misinterpreted: Is this a "force terminate" in the sense of "forcing a process to terminate" or is it in the sense that if the force (applied on the Agent) is too high the agent terminates? f it is the latter, I think this should have the same pattern as "CONSTRAINT_VIOLATE_SHOULD_END". Something like "EXCESSIVE_FORCE_SHOULD_END".

configs/tasks/rearrange/pick_state.yaml
```diff
@@ -48,11 +53,12 @@ TASK:
         TYPE: "EEPositionSensor"
     IS_HOLDING_SENSOR:
         TYPE: "IsHoldingSensor"
-    LOCAL_SENSOR:
-        TYPE: "LocalizationSensor"
     RELATIVE_RESTING_POS_SENSOR:
         TYPE: "RelativeRestingPositionSensor"
-    SENSORS: ["TARGET_START_SENSOR", "GOAL_SENSOR", "JOINT_SENSOR", "IS_HOLDING_SENSOR", "END_EFFECTOR_SENSOR", "LOCAL_SENSOR", "RELATIVE_RESTING_POS_SENSOR"]
+    JOINT_VEL_SENSOR:
```
vincentpierre, line 58:
```suggestion
    JOINT_VELOCITY_SENSOR:
```

configs/tasks/rearrange/pick_state.yaml
```diff
@@ -64,111 +70,93 @@ TASK:
...
-        DIST_REWARD: 20.0
-        SUCC_REWARD: 10.0
-        PICK_REWARD: 20.0
+        DIST_REWARD: 10.0
+        PICK_REWARD: 5.0
         CONSTRAINT_VIOLATE_PEN: 10.0
         DROP_PEN: 5.0
         WRONG_PICK_PEN: 5.0
-        COLL_PEN: 1.0
+        COLL_PEN: 0.0
```
vincentpierre, line 83: COLL_PEN is an abbreviation that is not immediately obvious to me. What are the conventions for naming these configurations? Is there a glossary somewhere?

examples/example.py
```diff
@@ -11,9 +11,7 @@ def example():
     # Note: Use with for the example testing, doesn't need to be like this on the README
```
vincentpierre, line 11: Suggestion : `example` is a bit too generic a method name. All this does is run one episode. I would rename the method to `run_one_episode` and make the config an argument (which defaults to `configs/tasks/rearrange/pick.yaml` if none is given).

vincentpierre: (probably out of scope for this PR now that read more of it)

examples/interactive_play.py
```diff
@@ -211,11 +239,11 @@ def play_env(env, args, config):
     if args.load_actions is not None:
         with open(args.load_actions, "rb") as f:
             use_arm_actions = np.load(f)
+            print("Loaded arm actions")
```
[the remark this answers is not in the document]
vincentpierre: Should we be using a logger instead of print in the examples?

examples/interactive_play.py
```diff
@@ -246,12 +274,16 @@ def play_env(env, args, config):
         if use_arm_actions is not None and i >= len(use_arm_actions):
             break
 
-        # obs, reward, done, info = step_result
         obs = step_result
-        reward = 0.0
         info = env.get_metrics()
+        reward_key = [k for k in info if "reward" in k]
```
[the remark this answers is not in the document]
vincentpierre: "reward" is a bit of a magic string here. I think it should be made a constant string and should also be made case insensitive (`k.lower()`)

examples/interactive_play.py
```diff
@@ -246,12 +274,16 @@ def play_env(env, args, config):
...
+        if len(reward_key) > 0:
+            reward = info[reward_key[0]]
```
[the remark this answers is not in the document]
vincentpierre: Should the reward be the first key's value or the sum of all the rewards ?

examples/interactive_play.py
```diff
@@ -279,14 +311,19 @@ def play_env(env, args, config):
...
     if args.save_actions:
-        assert len(all_arm_actions) > 200
-        all_arm_actions = np.array(all_arm_actions)[:200]
-        save_dir = "orp/start_data/"
+        if len(all_arm_actions) < args.save_actions_count:
+            raise ValueError(
+                f"Only did {len(all_arm_actions)} actions but {args.save_actions_count} are required"
+            )
+        all_arm_actions = np.array(all_arm_actions)[: args.save_actions_count]
+        save_dir = "data/interactive_play_replays"
```
vincentpierre, line 319: This should be a const or an environment variable even, I am sure this path will be reused in other scripts at some point.

examples/interactive_play.py
```diff
@@ -279,14 +311,19 @@ def play_env(env, args, config):
...
         os.makedirs(save_dir, exist_ok=True)
-        save_path = osp.join(save_dir, "bench_ac.txt")
+        save_path = osp.join(save_dir, "play_actions.txt")
```
vincentpierre, line 321: This should be a const string as well. Even better, it could be an optional argument so the user can manage their own recorded actions files.

habitat/datasets/rearrange/configs/pick/counter.yaml
```diff
@@ -0,0 +1,178 @@
+---
+dataset_path: "data/replica_cad/replicaCAD.scene_dataset_config.json"
```
vincentpierre, line 2: Not really relevant to this PR but I get the sense that configurations and datasets/scenes are often referred as their paths. I think this is going to be maybe hard to maintain on the long-term. This forces users to hard code paths at a lot of places. I think having some static registry of keys to paths might be a better approach. This way if you want to move an asset, you would just need to change it at one place (and you could also populate this registry every time an asset is downloaded).

habitat/datasets/rearrange/receptacle.py
```diff
@@ -0,0 +1,292 @@
...
+
+from typing import Dict, List, Optional
+
+import magnum as mn
+import numpy as np
+
+import habitat_sim
+
+
+class Receptacle:
```
vincentpierre, line 15: Is the Receptacle class only going to be useful in the rearrange scenario? I think this could be moved outside of datasets maybe?

habitat/datasets/rearrange/sim_utilities.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    # if output directory doesn't exist, create it
+    if not osp.exists(output_path):
+        import os
+
+        try:
+            os.makedirs(output_path)
+        except OSError:
+            print(
+                f"DebugVisualizer: Aborting 'make_debug_video': Failed to create the output directory: {output_path}"
```
vincentpierre, line 267: what is `make_debug_video` ? I do not think we should assume that check_make_dir is always called by make_debug_video. Same with DebugVisualizer

habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -17,90 +17,166 @@
 
 
 class RearrangeGraspManager:
```
vincentpierre, line 19: Looks like there are some missing typing on this script.

habitat_baselines/motion_planning/mp_sim.py
```diff
@@ -177,18 +181,23 @@ def add_sphere(self, radius, color=None):
         return sphere_id
 
     def remove_object(self, obj_id):
-        self._sim.remove_object(obj_id)
+
```
vincentpierre, line 184:
```suggestion
```

habitat/core/env.py
aclegg3: Good catch, thanks!

habitat/datasets/rearrange/rearrange_generator.py
aclegg3: I'll add some better comments, but perhaps defining the API for an abstract EpisodeGenerator may be out of scope here. I generally agree we should follow-up with design and refactor. For example, `pointnav_generator.py` defines a classless API for generating episodes which should then be refactored to use a generator also.

habitat/core/dataset.py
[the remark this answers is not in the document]
aclegg3: @mathfac I changed the name. 👍

habitat/datasets/rearrange/sim_utilities.py
aclegg3: Opted to move this into `habitat/sims/habitat_simulator` to group it with other functionality that leverages and requires a habitat-sim installation.

habitat_baselines/utils/gym_definitions.py
```diff
@@ -0,0 +1,109 @@
...
+GYM_AUTO_NAME_KEY = "GYM_AUTO_NAME"
+HABLAB_INSTALL_PATH = "HABLAB_BASE_CFG_PATH"
+
+if HABLAB_INSTALL_PATH in os.environ:
+    base_dir = os.environ[HABLAB_INSTALL_PATH]
+else:
+    base_dir = osp.dirname(osp.dirname(osp.dirname(osp.abspath(__file__))))
+
+
+def get_config_no_base_task_load(cfg_file_path):
```
vincentpierre, line 30: Are these two methods public facing (do we expect users to know about them and use them)? If yes, I think the comments should explain them a little more, other wise, I think these should be have a `_` prefix.

habitat_baselines/utils/gym_adapter.py
```diff
@@ -89,10 +89,15 @@ def __init__(self, env, save_orig_obs: bool = False):
         self._gym_achieved_goal_keys = env._rl_config.get(
             "GYM_ACHIEVED_GOAL_KEYS", []
         )
-        self._fix_info_dict = env._rl_config.get("GYM_FIX_INFO_DICT", False)
+        self._fix_info_dict = env._rl_config.get("GYM_FIX_INFO_DICT", True)
         self._gym_action_keys = env._rl_config.get("GYM_ACTION_KEYS", None)
         self._gym_obs_keys = env._rl_config.get("GYM_OBS_KEYS", None)
 
+        if self._gym_obs_keys is None:
```
vincentpierre, line 96: It looks like the source of truth is the config and if there is no `GYM_OBS_KEYS` in the config, then the fallback is the observation space of the environment. Is that the expected behavior? 
If the observation_space is always available, why is there a `GYM_OBS_KEYS` in the config? (It looks redundant).
I think we should always use `env.observation_space.spaces` and throw an error if it is different from the config provided.

habitat/datasets/rearrange/receptacle.py
aclegg3: Agreed, this could be useful elsewhere, but no concrete ideas on that yet. Let's defer this to a later refactor.

habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -5,118 +5,232 @@
...
-        self._leave_info = None
+    def reconfigure(self) -> None:
+        """Removes any existing constraints managed by this structure.
+        Called from _sim.reconfigure().
+        """
+        for constraint_id in self._snap_constraints:
+            self._sim.remove_rigid_constraint(constraint_id)
+        self._snap_constraints.clear()
 
+    def reset(self) -> None:
```
aclegg3, line 53: Do reconfigure and reset need to be separate? Both called from sim.reconfigure only presently.

habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -5,118 +5,232 @@
...
 
     @property
-    def snap_idx(self):
+    def snap_idx(self) -> int:
+        """The index of the grasped RigidObject."""
         return self._snapped_obj_id
 
     @property
-    def snap_rigid_obj(self):
+    def snapped_marker_id(self) -> str:
```
aclegg3, line 148: All marker "ids" are actually strings, so maybe should be refactored to "key"

habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -125,23 +239,21 @@ def snap_to_obj(self, snap_obj_id: int, force: bool = True):
...
+                mn.Vector3(0.0, 0, 0),
+                mn.Vector3(-0.1, 0, 0),
+                self._snapped_obj_id,
             ),
-            create_hold_constraint(
-                mn.Vector3(0.1, 0.0, 0.1), mn.Vector3(0.0, 0.0, 0.1)
+            self.create_hold_constraint(
+                mn.Vector3(0.1, 0.0, 0.1),
+                mn.Vector3(0.0, 0.0, 0.1),
+                self._snapped_obj_id,
```
aclegg3, line 256: The fixed constraint type is working nicely these days for 6dof constraints, so maybe we can git rid of this hack and just use that.

aclegg3 (commented):
Some thoughts after documenting/typing the code.

configs/tasks/rearrange/pick_state.yaml
ASzot: Fixed

configs/tasks/rearrange/pick_state.yaml
ASzot: There is no glossary. I think it's important we get documentation for the configs but I am not sure where to put it.

ASzot: I updated the name.

habitat_baselines/utils/gym_definitions.py
ASzot: They are indeed private methods, fixed.

habitat_baselines/utils/gym_adapter.py
ASzot: Yes, this is how it works and I think this should be the desired behavior. The GYM_OBS_KEY is used for filtering out the observations that are wanted in the Gym environment.

habitat/tasks/rearrange/rearrange_grasp_manager.py
ASzot: What do you mean by this?

habitat/tasks/rearrange/rearrange_grasp_manager.py
ASzot: Yes, let's address this in future PR.

aclegg3 (approved):
LGTM

PR #818, opened 2022-02-28
Added updates for HM3D ObjectNav training

srama2512:
## Motivation and Context

This PR adds necessary configs for HM3D ObjectNav training. It includes new features for training with a `SemanticCategory` sensor. It also updates the generate_video function to include scene_id, goal_name in the save path.

## How Has This Been Tested

Tested locally.

## Types of changes

## Checklist

- [ x] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] I have completed my CLA (see **CONTRIBUTING**)


habitat/tasks/nav/nav.py
```diff
@@ -514,6 +521,118 @@ def get_observation(
...
+        semantic[semantic >= self.instance_id_to_task_id.shape[0]] = 0
+        # Map from instance id to task id
+        semantic_category = np.take(self.instance_id_to_task_id, semantic)
+        if self.config.CONVERT_TO_RGB:
+            semantic_category = self.convert_semantic_to_rgb(semantic_category)
+
+        return semantic_category
+
+    def convert_semantic_to_rgb(self, x):
+        max_valid_id = max(list(self.category_to_task_category_id.values()))
```
Skylion007, line 621:
```suggestion
        max_valid_id = max(self.category_to_task_category_id.values())
```

habitat/tasks/nav/nav.py
```diff
@@ -514,6 +521,118 @@ def get_observation(
...
+                high=np.iinfo(np.int32).max,
+                shape=(self.config.HEIGHT, self.config.WIDTH),
+                dtype=np.int32,
+            )
+        return observation_space
+
+    def get_observation(
+        self, *args: Any, observations, episode, **kwargs: Any
+    ):
+        episode_uniq_id = f"{episode.scene_id} {episode.episode_id}"
```
Skylion007, line 593: The semantic sensor oservation here could be a Tensor right, which would create an unnecessary CPU <-> GPU. I think you can easily modify this to operator on both NpArrays and Torch.Tensor

Skylion007:
@srama2512 This needs testing in the test suite like from so: https://github.com/facebookresearch/habitat-lab/blob/3f2c7cb13f6f8c91643302541c358f5c712b7be8/test/test_sensors.py#L443

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -119,15 +119,22 @@ def __init__(
         else:
             self._n_input_depth = 0
 
+        if "semantic_category" in observation_space.spaces:
+            self._n_input_semantic_category = observation_space.spaces["semantic_category"].shape[2]
+            spatial_size = observation_space.spaces["semantic_category"].shape[0] // 2
+            assert self._n_input_semantic_category == 3, "ResNetEncoder only supports RGB values from SemanticCategory sensor!"
```
mathfac, line 125: Why `semantic_category` why it has RGB values, not int id of the category?

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -119,15 +119,22 @@ def __init__(
...
+        else:
+            self._n_input_semantic_category = 0
+
         if normalize_visual_inputs:
             self.running_mean_and_var: nn.Module = RunningMeanAndVar(
-                self._n_input_depth + self._n_input_rgb
+                self._n_input_depth + self._n_input_rgb + self._n_input_semantic_category
```
mathfac, line 131: normalizing semantic_category inputs doesn't look right.

habitat/tasks/nav/nav.py
```diff
@@ -514,6 +521,118 @@ def get_observation(
...
+        assert config.DATASET in ["gibson", "mp3d", "hm3d"]
+        if config.DATASET == "gibson":
+            cat_mapping = GIBSON_CATEGORY_TO_TASK_CATEGORY_ID
+        elif config.DATASET == "mp3d":
+            cat_mapping = MP3D_CATEGORY_TO_TASK_CATEGORY_ID
+        else:
+            cat_mapping = HM3D_CATEGORY_TO_TASK_CATEGORY_ID
+        self.category_to_task_category_id = cat_mapping
+        if config.RAW_NAME_TO_CATEGORY_MAPPING != "":
+            with open(config.RAW_NAME_TO_CATEGORY_MAPPING, "r") as fp:
```
mathfac, line 557: If we leave this code it should use some csv reader or etc.

habitat/tasks/nav/nav.py
```diff
@@ -514,6 +521,118 @@ def get_observation(
         )
 
 
+@registry.register_sensor(name="SemanticCategorySensor")
+class SemanticCategorySensor(Sensor):
```
mathfac, line 525: Do we use this sensor for training or only for visualization purposes?

habitat/tasks/nav/nav.py
```diff
@@ -514,6 +521,118 @@ def get_observation(
...
+                    obj_name = self.raw_to_cat_mapping[obj_name]
+                    obj_task_id = self.category_to_task_category_id[obj_name]
+                    self.instance_id_to_task_id[obj_inst_id] = obj_task_id
+        # Set invalid instance IDs to unknown object 0
+        semantic = np.copy(observations["semantic"])
+        semantic[semantic >= self.instance_id_to_task_id.shape[0]] = 0
+        # Map from instance id to task id
+        semantic_category = np.take(self.instance_id_to_task_id, semantic)
+        if self.config.CONVERT_TO_RGB:
+            semantic_category = self.convert_semantic_to_rgb(semantic_category)
```
mathfac, line 616: RGB conversion should live in `def observations_to_image(observation: Dict, info: Dict) -> np.ndarray:`.
Something like this:
```
    if "semantic_category" in observation:
        flat_sem = observation[
            obj_semantic_name
        ]  # to move to same scale #.permute(2, 0, 1).unsqueeze(0).data.max(1)[1].cpu().numpy()[0]
        flat_sem[flat_sem == 41] = 40
        if not isinstance(flat_sem, np.ndarray):
            flat_sem = flat_sem.cpu().numpy()
        semantic_segmentation = (
            color_label(flat_sem).squeeze().transpose(1, 2, 0).astype(np.uint8)
        )
        egocentric_view.append(semantic_segmentation)

        if "objectgoal" in observation and "episode_info" in info:
            from habitat.tasks.nav.object_nav_task import task_cat2mpcat40

            # permute tensor to dimension [CHANNEL x HEIGHT X WIDTH]
            idx = task_cat2mpcat40[observation["objectgoal"][0]]
            goal_segmentation = (
                color_label(flat_sem == idx)
                .squeeze()
                .transpose(1, 2, 0)
                .astype(np.uint8)
            )
            egocentric_view.append(goal_segmentation)
```

habitat/tasks/nav/semantic_constants.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+GIBSON_CATEGORY_TO_TASK_CATEGORY_ID = {
```
mathfac, line 8: This file should be somewhere: `habitat/datasets/object_nav/...`.

habitat_baselines/rl/ddppo/policy/resnet_policy.py
srama2512: Right now, the semantic category sensor outputs an RGB image by default. For each goal category, there is a unique color. For all other categories and backgrounds, the pixels are set to zeros.

habitat_baselines/rl/ddppo/policy/resnet_policy.py
srama2512: The semantic_category input is forced to be RGB (see earlier comment).

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -66,6 +66,7 @@ class PPOTrainer(BaseRLTrainer):
     supported_tasks = ["Nav-v0"]
 
     SHORT_ROLLOUT_THRESHOLD: float = 0.25
+    SENSORS_BLACKLIST = ["semantic"]
```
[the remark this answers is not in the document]
srama2512: Having the `semantic` value throws errors in various parts of the code (observation transforms, observations rollouts, etc). You could try removing it from the blacklist and checking. This seemed the more elegant solution, but I'm happy to implement a better solution based on feedback.

habitat/tasks/nav/nav.py
srama2512: This sensor is used for training models with GT semantic goal inputs.

habitat/tasks/nav/nav.py
srama2512: This sensor is not for visualization purposes. It is intended to be a model input for training with oracle semantics. So the logic should live within the sensor itself, right?

habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -1058,16 +1076,21 @@ def _eval_checkpoint(
                             current_episodes[i].episode_id,
                         )
                     ] = episode_stats
+                    goal_name = None
```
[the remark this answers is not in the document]
srama2512: The idea was to add the goal name to the video save path (so that we can visualize episodes based on the goal). How do you suggest we go about this?

habitat/tasks/nav/semantic_constants.py
```diff
@@ -0,0 +1,65 @@
...
+    'cup': 15,
+    'umbrella': 16,
+    'bowl': 17,
+    'gym_equipment': 18,
+    'bench': 19,
+    'clothes': 20
+}
+
+
+MP3D_CATEGORY_TO_TASK_CATEGORY_ID = {
```
erikwijmans, line 33: This is different than https://github.com/niessner/Matterport/blob/master/metadata/mpcat40.tsv which will likely lead to confusion.

habitat/tasks/nav/semantic_constants.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+GIBSON_CATEGORY_TO_TASK_CATEGORY_ID = {
+    'chair': 0,
```
erikwijmans, line 9: Based on "Set invalid instance IDs to unknown object 0" you use `0` as unknown elsewhere.

habitat/tasks/nav/nav.py
mathfac: @srama2512, yes. But that GT data isn't available during evaluation and from this PR isn't clear how it will be replaced.

mathfac (commented):
That's still not clear why we need Semantic_category represent as RGB for training. Non-intuitive and can lead to more issues when categories are cross dataset.
As it uses GT Semantic information it's not clear how this PR is used to train final baseline.

mathfac:
@srama2512, @devendrachaplot how we want evolve this PR? Do we more up to date code focused on ObjectNav

PR #822, opened 2022-03-04
HM3D Semantics update

srama2512:
## Motivation and Context

Added config files for HM3D ObjectNav, and updated README with ObjectNav dataset paths.

## How Has This Been Tested

Tested locally.

## Types of changes

## Checklist


configs/tasks/objectnav_hm3d.yaml
```diff
@@ -0,0 +1,49 @@
...
+    HEIGHT: 0.88
+    RADIUS: 0.18
+  HABITAT_SIM_V0:
+    GPU_DEVICE_ID: 0
+    ALLOW_SLIDING: False
+  SEMANTIC_SENSOR:
+    WIDTH: 640
+    HEIGHT: 480
+    HFOV: 79
+    POSITION: [0, 0.88, 0]
```
Skylion007, line 19: Nit: You can use yaml anchors to avoid repeating this over and over again. I know we don't do that for a lot of Habitat currently, but we should.

erikwijmans (approved):

habitat_baselines/config/objectnav/ddppo_objectnav_hm3d.yaml
```diff
@@ -0,0 +1,74 @@
+BASE_TASK_CONFIG_PATH: "configs/tasks/objectnav_hm3d.yaml"
+CMD_TRAILING_OPTS: ["TASK_CONFIG.ENVIRONMENT.ITERATOR_OPTIONS.MAX_SCENE_REPEAT_STEPS", "50000"]
```
Skylion007, line 2: Is there no better way to express this value than this field?  This is for overriding values from CLI isn't it? Could lead to unexpected behavior left in.

Skylion007 (changes requested):

Skylion007 (approved):
Honestly there are nits and it looks good otherwise.

habitat_baselines/config/objectnav/ddppo_objectnav_hm3d.yaml
mathfac: I am not sure there is one. It was reused form the objectNav config.

mathfac (approved):

configs/tasks/objectnav_hm3d.yaml
mathfac: Possibly we should do it everywhere then.

PR #850, opened 2022-04-12
Add eval support and debug sensor

eundersander:
Some notes for these changes:

off-by-one
- we should look closely at `_eval_checkpoint` to confirm I didn't introduce an off-by-one bug in terms of actions, steps, observations.

`BatchedEnv.is_eval`
- this is a hacky public member var for now but should instead be hooked up to a yaml config
- this member dictates that BatchedEnv should run through all episodes in the dataset only once; envs are disabled as necessary towards the end of the evaluation

recommended config for eval
- `DEBUG_RGB_SENSOR`
    - when evaluating, be sure to include `"DEBUG_RGB_SENSOR"` in `config.SENSORS`
    - be sure not to include this when training (it will hurt runtime perf)
- debug sensor size
    - see `batched_env.py` `debug_width` and `debug_height`
    - the size here greatly impacts eval speed
- `VIDEO_OPTION: ["disk"]`
    - use this for eval
    - Beware I hit a runtime error when I enabled the `"tensorboard"` option (maybe my version of tensorboard is outdated)
- `NUM_ENVIRONMENTS: 32`
    - Maybe try larger, but beware I was seeing a crash on my system seemingly due to GPU OOM.
- `EVAL_CKPT_PATH_DIR`
    - set to the same as `CHECKPOINT_FOLDER`

`gala_kinematic/python/sim_test.py` (not in `habitat-lab`)
- this is a script that generates a few random rollouts with videos (random actions; no inference)
- this is a handy place to tune cameras

### DIFF
--- a/habitat/core/batched_env.py
+++ b/habitat/core/batched_env.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import copy
 from collections import OrderedDict
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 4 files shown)


ASzot:
LGTM. I can also add in more of the rendering functionality from Habitat 2.0, like rendering all of the info metrics to the current frame so we can have better insight into metrics used to compute rewards.

PR #898, opened 2022-07-07
Creating Skeleton for Static Manipulator

laikhtewari:
## Motivation and Context

The static manipulator class represents a type of RobotInterface with a fixed base and controllable arm. These static robots are easily controllable in simulation and reality with a single polymetis controller. The static manipulator differs from the mobile manipulator in two ways. First, no mobility information will be present in the class, providing a semantically succinct interface. Second, additional features may be added to provide an elegant bridge between polymetis and habitat. 

## How Has This Been Tested

A instance of the StaticManipulator class was created using a StaticManipulatorParams object containing dummy data. 

## Types of changes

## Checklist


habitat/robots/static_manipulator.py
```diff
@@ -0,0 +1,103 @@
...
+    :property arm_mtr_max_impulse: The maximum impulse of the arm motor.
+    """
+
+    arm_joints: List[int]
+    gripper_joints: List[int]
+
+    arm_init_params: Optional[np.ndarray]
+    gripper_init_params: Optional[np.ndarray]
+
+    ee_offset: mn.Vector3
```
Skylion007, line 39: Does this need to be a vector 3? Can it not implicitly convert from the np.ndarray?

habitat/robots/static_manipulator.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from collections import defaultdict
+from typing import Dict, List, Optional, Set, Tuple
+
+import attr
+import magnum as mn
```
Skylion007, line 9: It'd be nice for sim2real deployment if we didn't need Magnum.

aclegg3:
Do you think you could re-use any of the gripper and arm control functionality in the mobile manipulator class? Would it make sense to break the manipulator out of that class and share the functionality between a mobile and static manipulator platforms?

aclegg3:
It will be helpful to define a concrete instantiation of the class with unit tests before we merge in the abstract implementation. This will help us iron out details and identify gaps before committing to a design so we avoid code thrashing. Pick a robot platform and test it out. 🙂 

You can choose to do this in a WIP (draft) PR and then break out the pieces to commit separately if you'd prefer to keep PRs smaller.

PR #905, opened 2022-07-14
Variable Experience Rollout [Part 0]

erikwijmans:
## Motivation and Context

These are changes I made that either import QOL and/or are needed for Variable Experience Rollout (the new training system I've been working on).

In an effort to break up merging that code, I am sending in these changes first.

Summary of changes:

1. Improved logging from PPO. I've added things like grad norm, fraction clipped, predicted value, etc. These metrics are really useful for debugging.
2. Preemptable evaluation
3. Build the RNN packed sequence info during batching. This stops a pipeline bubble.
4. Always use the batching cache and make that a singleton. No reason to not use it.
5. Some additional helper functions in TensorDict. Most of those changes are a refactor to better support typing.
6. Evaluate episodes multiple times.
7. A `gather_objects` function that performs `torch.distributed.gather` but on arbitrary python objects.
8. A pretty big change to how the RNN seq info is built that I encourage you to ignore and just trust the unit test :)

## Types of changes

### DIFF
--- a/habitat_baselines/common/base_trainer.py
+++ b/habitat_baselines/common/base_trainer.py
@@ -18,7 +18,13 @@ from habitat_baselines.common.tensorboard_utils import (
     TensorboardWriter,
     get_writer,
 )
-from habitat_baselines.rl.ddppo.ddp_utils import SAVE_STATE, is_slurm_batch_job
+from habitat_baselines.rl.ddppo.ddp_utils import (
+    SAVE_STATE,
+    add_signal_handlers,
+    is_slurm_batch_job,
+    load_resume_state,
+    save_resume_state,
+)
 from habitat_baselines.utils.common import (
     get_checkpoint_id,
     poll_checkpoint_folder,
@@ -81,6 +87,16 @@ class BaseTrainer:
         Returns:
             None
         """
+        if is_slurm_batch_job():
+            add_signal_handlers()
+
+        resume_state = load_resume_state(self.config, filename_key="eval")
+        if resume_state is not None:
+            self.config = resume_state["config"]
+            prev_ckpt_ind = resume_state["prev_ckpt_ind"]
+        else:
+            prev_ckpt_ind = -1
+
         self.device = (
             torch.device("cuda", self.config.TORCH_GPU_ID)
             if torch.cuda.is_available()
@@ -97,7 +113,14 @@ class BaseTrainer:
                 len(self.config.VIDEO_DIR) > 0
             ), "Must specify a directory for storing videos on disk"
 
-        with get_writer(self.config, flush_secs=self.flush_secs) as writer:
+        if prev_ckpt_ind == -1:
+            purge_step = 0
+        else:
+            purge_step = None
+
+        with get_writer(
+            self.config, flush_secs=self.flush_secs, purge_step=purge_step
+        ) as writer:
             if os.path.isfile(self.config.EVAL_CKPT_PATH_DIR):
                 # evaluate singe checkpoint
                 proposed_index = get_checkpoint_id(
@@ -114,7 +137,6 @@ class BaseTrainer:
                 )
             else:
                 # evaluate multiple checkpoints in order
-                prev_ckpt_ind = -1
                 while True:
                     current_ckpt = None
                     while current_ckpt is None:
@@ -129,6 +151,18 @@ class BaseTrainer:
                         writer=writer,
                         checkpoint_index=prev_ckpt_ind,
                     )
+
+                    save_resume_state(
+                        {
+                            "config": self.config,
+                            "prev_ckpt_ind": prev_ckpt_ind,
+                        },
+                        self.config,
+                        filename_key="eval",
+                    )
+
+                    if (prev_ckpt_ind + 1) == self.config.NUM_CHECKPOINTS:
+                        break
 
     def _eval_checkpoint(
         self,
@@ -314,6 +348,7 @@ class BaseRLTrainer(BaseTrainer):
                 batch[k] = v[state_index]
 
             rgb_frames = [rgb_frames[i] for i in state_index]
+            # actor_critic.do_pause(state_index)
 
         return (
             envs,
--- a/habitat_baselines/common/base_trainer.py
+++ b/habitat_baselines/common/base_trainer.py
@@ -79,6 +79,10 @@ class BaseTrainer:
 
         return config
 
+    def _add_preemption_signal_handlers(self):
+        if is_slurm_batch_job():
+            add_signal_handlers()
+
     def eval(self) -> None:
         r"""Main method of trainer evaluation. Calls _eval_checkpoint() that
         is specified in Trainer class that inherits from BaseRLTrainer
@@ -87,11 +91,15 @@ class BaseTrainer:
         Returns:
             None
         """
-        if is_slurm_batch_job():
-            add_signal_handlers()
+
+        self._add_preemption_signal_handlers()
 
         resume_state = load_resume_state(self.config, filename_key="eval")
         if resume_state is not None:
+            # If we have a resume state saved, that means
+            # we are resuming an evaluation session that got
+            # preempted. We grab the config and the prev_ckpt_ind
+            # so that we pick-up from the checkpoint we left off with
             self.config = resume_state["config"]
             prev_ckpt_ind = resume_state["prev_ckpt_ind"]
         else:
@@ -113,14 +121,7 @@ class BaseTrainer:
                 len(self.config.VIDEO_DIR) > 0
             ), "Must specify a directory for storing videos on disk"
 
-        if prev_ckpt_ind == -1:
-            purge_step = 0
-        else:
-            purge_step = None
-
-        with get_writer(
-            self.config, flush_secs=self.flush_secs, purge_step=purge_step
-        ) as writer:
+        with get_writer(self.config, flush_secs=self.flush_secs) as writer:
             if os.path.isfile(self.config.EVAL_CKPT_PATH_DIR):
                 # evaluate singe checkpoint
                 proposed_index = get_checkpoint_id(
@@ -152,6 +153,9 @@ class BaseTrainer:
                         checkpoint_index=prev_ckpt_ind,
                     )
 
+                    # We save a resume state during evaluation so that
+                    # we can resume evaluating incase the job gets
+                    # preempted.
                     save_resume_state(
                         {
                             "config": self.config,
--- a/habitat_baselines/common/construct_vector_env.py
+++ b/habitat_baselines/common/construct_vector_env.py
@@ -43,7 +43,7 @@ def construct_envs(
     if len(scenes) < num_environments:
         msg = f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
         if is_eval:
-            raise RuntimeError(msg + "This is invalid for evaluation.")
+            raise ValueError(msg + "This is invalid for evaluation.")
         else:
             logger.warn(
                 msg
--- a/habitat_baselines/common/construct_vector_env.py
+++ b/habitat_baselines/common/construct_vector_env.py
@@ -9,6 +9,7 @@ from habitat.utils.gym_definitions import make_gym_from_config
 def construct_envs(
     config: Config,
     workers_ignore_signals: bool = False,
+    is_eval: bool = False,
 ) -> VectorEnv:
     r"""Create VectorEnv object with specified config and env class type.
     To allow better performance, dataset are split into small ones for
@@ -40,10 +41,14 @@ def construct_envs(
 
     scene_splits: List[List[str]] = [[] for _ in range(num_environments)]
     if len(scenes) < num_environments:
-        logger.warn(
-            f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
-            "Each environment will use all the scenes instead of using a subset."
-        )
+        msg = f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
+        if is_eval:
+            raise RuntimeError(msg + "This is invalid for evaluation.")
+        else:
+            logger.warn(
+                msg
+                + "Each environment will use all the scenes instead of using a subset."
+            )
         for scene in scenes:
             for split in scene_splits:
                 split.append(scene)
--- a/habitat_baselines/common/construct_vector_env.py
+++ b/habitat_baselines/common/construct_vector_env.py
@@ -9,7 +9,7 @@ from habitat.utils.gym_definitions import make_gym_from_config
 def construct_envs(
     config: Config,
     workers_ignore_signals: bool = False,
-    is_eval: bool = False,
+    enforce_scenes_greater_eq_environments: bool = False,
 ) -> VectorEnv:
     r"""Create VectorEnv object with specified config and env class type.
     To allow better performance, dataset are split into small ones for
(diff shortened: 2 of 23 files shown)


habitat_baselines/common/construct_vector_env.py
```diff
@@ -40,10 +41,14 @@ def construct_envs(
 
     scene_splits: List[List[str]] = [[] for _ in range(num_environments)]
     if len(scenes) < num_environments:
-        logger.warn(
-            f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
-            "Each environment will use all the scenes instead of using a subset."
-        )
+        msg = f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
+        if is_eval:
+            raise RuntimeError(msg + "This is invalid for evaluation.")
```
Skylion007, line 46:
```suggestion
            raise ValueError(msg + "This is invalid for evaluation.")
```
Probably should pick a slightly more specific Error.

habitat_baselines/common/rollout_storage.py
```diff
@@ -208,13 +219,37 @@ def recurrent_generator(
...
+                device=self.device,
+                build_fn_result=build_pack_info_from_dones(
+                    dones_cpu[curr_slice],
+                ),
+            )
+
+            yield batch
+
+    def __getstate__(self):
+        return self.__dict__.copy()
```
Skylion007, line 251: Yikes, why a copy? Even if it's shallow?

Skylion007 (approved):
Looks mostly good, just had question about why you are doing a shallow copy of the __dict__ in that one spot, sure you have a good reason (might want to add a comment why). Otherwise, looks good.

habitat_baselines/common/rollout_storage.py
erikwijmans: The rollout storage for VER inherits from this and it has to delete stuff from the state dict, but it should probably be the one to make the copy.

Skylion007:
@erikwijmans There are protocols available from typing_extensions, I would add them back.

habitat_baselines/common/tensor_dict.py
```diff
@@ -15,7 +15,6 @@
     Iterable,
     List,
     Optional,
-    Protocol,
```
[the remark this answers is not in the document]
erikwijmans: I think torch installs that. Should we just use whatever version they install?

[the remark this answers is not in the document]
Skylion007: Seems reasonable to me. It's actually a dependencies of quite a few libraries.

dhruvbatra:
> A pretty big change to how the RNN seq info is built that I encourage you to ignore and just trust the unit test :)

Such a programmer flex :-).

habitat_baselines/common/base_trainer.py
```diff
@@ -18,7 +18,13 @@
...
     get_writer,
 )
-from habitat_baselines.rl.ddppo.ddp_utils import SAVE_STATE, is_slurm_batch_job
+from habitat_baselines.rl.ddppo.ddp_utils import (
+    SAVE_STATE,
+    add_signal_handlers,
+    is_slurm_batch_job,
+    load_resume_state,
+    save_resume_state,
+)
```
vincentpierre, line 27: I am against importing those in both base_trainer and in ppo_trainer. Also against doing 
```
        if is_slurm_batch_job():
            add_signal_handlers()
```
In both eval and training. If both need to call it, let's put in in a helper method and call it from both.

habitat_baselines/common/base_trainer.py
```diff
@@ -81,6 +87,16 @@ def eval(self) -> None:
         Returns:
             None
         """
+        if is_slurm_batch_job():
+            add_signal_handlers()
+
+        resume_state = load_resume_state(self.config, filename_key="eval")
+        if resume_state is not None:
+            self.config = resume_state["config"]
+            prev_ckpt_ind = resume_state["prev_ckpt_ind"]
```
vincentpierre, line 96: Why keeping the previous? Can you add a one liner explaining it?

habitat_baselines/common/base_trainer.py
```diff
@@ -97,7 +113,14 @@ def eval(self) -> None:
                 len(self.config.VIDEO_DIR) > 0
             ), "Must specify a directory for storing videos on disk"
 
-        with get_writer(self.config, flush_secs=self.flush_secs) as writer:
+        if prev_ckpt_ind == -1:
+            purge_step = 0
```
vincentpierre, line 117: Move this line 99

habitat_baselines/common/base_trainer.py
```diff
@@ -97,7 +113,14 @@ def eval(self) -> None:
...
+        else:
+            purge_step = None
+
+        with get_writer(
+            self.config, flush_secs=self.flush_secs, purge_step=purge_step
```
vincentpierre, line 122: The WandB or future writer might have an error if it has this new key it does not know about. Is purge_step necessary?

habitat_baselines/common/base_trainer.py
```diff
@@ -130,6 +152,18 @@ def eval(self) -> None:
                         checkpoint_index=prev_ckpt_ind,
                     )
 
+                    save_resume_state(
```
vincentpierre, line 155: Why are we saving the state in the evaluation? Feels kind of confusing to me. If it is needed, can you add a comment as of why?

habitat_baselines/common/construct_vector_env.py
```diff
@@ -9,6 +9,7 @@
 def construct_envs(
     config: Config,
     workers_ignore_signals: bool = False,
+    is_eval: bool = False,
```
vincentpierre, line 12: I am against this. 
Environments should not know about wether or no they are used for evaluation. Adding this new argument does not change the behavior of the environments and only affects wether a log message or an error happens. If you really want it, rename to `enforce_less_scenes_than_environemnts`.

habitat_baselines/common/construct_vector_env.py
```diff
@@ -40,10 +41,14 @@ def construct_envs(
 
     scene_splits: List[List[str]] = [[] for _ in range(num_environments)]
     if len(scenes) < num_environments:
-        logger.warn(
-            f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
-            "Each environment will use all the scenes instead of using a subset."
-        )
+        msg = f"There are less scenes ({len(scenes)}) than environments ({num_environments}). "
+        if is_eval:
+            raise ValueError(msg + "This is invalid for evaluation.")
```
vincentpierre, line 46: Why do we want to raise an error? Should we just reduce the number of environments accordingly if you really don't want the environments to have all scenes?

habitat_baselines/common/rollout_storage.py
```diff
@@ -84,9 +88,11 @@ def __init__(
 
         assert (self._num_envs % self._nbuffers) == 0
 
-        self.numsteps = numsteps
+        self.num_steps = numsteps
         self.current_rollout_step_idxs = [0 for _ in range(self._nbuffers)]
 
+        self.device = torch.device("cpu")
```
vincentpierre, line 94: Add a comment explaining why the device needs to be the cpu

erikwijmans: It doesn't need to be the CPU, you can move the rollout storage to another device with the `.to` method. The default device for a torch tensor is the CPU, so everything we've allocated is on the CPU. This is just tracking that fact.

habitat_baselines/rl/ddppo/ddp_utils.py
```diff
@@ -287,3 +301,144 @@ def find_free_port() -> int:
         sock.bind(("localhost", 0))
         _, port = sock.getsockname()
         return port
+
+
+def get_free_port_distributed(
```
vincentpierre, line 306: code comment

habitat_baselines/common/construct_vector_env.py
erikwijmans: Sure, I'm fine with the name change.

We do want this change. If people have more environments than scenes that will silently cause evaluation to be incorrect.

habitat_baselines/rl/ddppo/ddp_utils.py
```diff
@@ -287,3 +301,144 @@ def find_free_port() -> int:
...
+    rank: int, output_rank: int, world_size: int
+) -> int:
+    return (
+        rank - output_rank
+        if rank >= output_rank
+        else rank - output_rank + world_size
+    )
+
+
+def gatherv(
```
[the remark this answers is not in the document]
vincentpierre: This method is rather long, it it possible to split it into smaller methods?

habitat_baselines/rl/ddppo/ddp_utils.py
```diff
@@ -287,3 +301,144 @@ def find_free_port() -> int:
...
+
+    if is_mine:
+        assert output is not None
+    else:
+        assert output is None
+
+    return output
+
+
+def gather_objects(
```
vincentpierre, line 422: code comment.

habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -423,6 +423,7 @@ def forward(
         rnn_hidden_states,
         prev_actions,
         masks,
+        rnn_build_seq_info=None,
```
vincentpierre, line 426: Add typing?

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -4,169 +4,201 @@
...
+from habitat_baselines.common.tensor_dict import TensorDict
+
 
 def _invert_permutation(permutation: torch.Tensor) -> torch.Tensor:
-    output = torch.empty_like(permutation)
+    output = torch.empty_like(permutation.view(-1))
     output.scatter_(
         0,
-        permutation,
+        permutation.view(-1),
```
vincentpierre, line 21: permutation.view(-1) called twice, store the result of permutation.view(-1) and then use it twice.

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -4,169 +4,201 @@
...
         torch.arange(0, permutation.numel(), device=permutation.device),
     )
-    return output
+    return output.view_as(permutation)
+
 
+def _np_invert_permutation(permutation: np.ndarray) -> np.ndarray:
```
vincentpierre, line 27: this private method does not seem to be used.

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -4,169 +4,201 @@
...
 
-def _build_pack_info_from_dones(
-    dones: torch.Tensor,
-    T: int,
-) -> Tuple[
-    torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor
-]:
+
+# This is some pretty wild code. I recommend you just trust
+# the unit test on it and leave it be.
```
vincentpierre, line 32:
```suggestion
```

habitat_baselines/rl/models/rnn_state_encoder.py
```diff
@@ -4,169 +4,201 @@
...
 
-    # We will want these on the CPU for torch.unique_consecutive,
-    # so move now.
-    cpu_lengths = lengths.to(device="cpu", non_blocking=True)
+    max_length = int(lengths[0])
 
-    episode_starts = episode_starts.index_select(0, sorted_indices)
-    select_inds = torch.empty((T * N), device=dones.device, dtype=torch.int64)
+    #  for i, eid in enumerate(unique_episode_ids):
+    #  assert sequence_starts[i] == (episode_ids == eid).nonzero()[0].min()
```
vincentpierre, line 84:
```suggestion
```

habitat_baselines/common/tensor_dict.py
```diff
@@ -4,77 +4,161 @@
...
 
-        print(t["a"])
 
-    """
+_MapFuncType = Callable[[T], T]
+
+_ApplyFuncType = Callable[[T], None]
+
+
+class _DictTreeBase(Dict[str, Union["_DictTreeBase[T]", T]]):
```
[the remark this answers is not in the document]
erikwijmans: The ones that aren't used currently will be used in Part 1.

habitat_baselines/common/construct_vector_env.py
erikwijmans: Sure. I can just change the number.

habitat_baselines/rl/models/rnn_state_encoder.py
erikwijmans: It will be.

habitat_baselines/rl/ddppo/algo/ddppo.py
```diff
@@ -6,16 +6,36 @@
...
 
-def distributed_mean_and_var(
+
+def _cpu_to_numpy(inp):
+    return _recursive_apply(
+        inp, lambda t: t.numpy() if t.device.type == "cpu" else t
+    )
+
+
+def _numpy_to_cpu(inp):
```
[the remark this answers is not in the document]
erikwijmans: I see these as backward compatibility hacks because we are using a very old PyTorch on the CI that hasn't deprecated the single-process multi-device of DDP. I think TensorDict and NDArrayDict (or other possible variants) cover most general cases and I've already gone through the pain to give those proper typing.

habitat_baselines/rl/ddppo/algo/ddppo.py
```diff
@@ -6,16 +6,36 @@
...
+    if isinstance(inp, dict):
+        return type(inp)((k, _recursive_apply(v, fn)) for k, v in inp.items())
+    elif isinstance(inp, (tuple, list)):
+        return type(inp)(_recursive_apply(v, fn) for v in inp)
+    else:
+        return fn(inp)
 
-def distributed_mean_and_var(
+
+def _cpu_to_numpy(inp):
```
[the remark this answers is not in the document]
erikwijmans: Yeah, it was. I just converts cpu tensors to numpy so that PyTorch DDP doesn't move them the GPU.

habitat_baselines/common/base_trainer.py
erikwijmans: Actually backed this one out. The one for training is much more important.

vincentpierre (approved):
(Will be on PTO next week so I am approving so my `request changes` is not blocking you.)

habitat_baselines/config/default.py
```diff
@@ -77,6 +77,7 @@
 _C.EVAL.SPLIT = "val"
```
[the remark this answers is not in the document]
ASzot: Yes, this setting is necessary to evaluate the chained skill policy.

ASzot:
@erikwijmans Since this PR and the subsequent PRs change so much of the training code, have you done an end-to-end test that the old code paths still work as expected? For example, does a Pick policy trained with DD-PPO work exactly the same as before?

erikwijmans:
None of this should impact any existing policies as this is almost entirely additive and the things that aren't are simple refactors, but checking is a good idea. I can check with a PointNav model. Do you have a pick model that I could test with?

ASzot:
I mean training a policy from scratch, not evaluating. I agree the changes are additive, but it does touch a lot of training code. Yes, I think it would be great if you could train a PointNav and Pick policy from scratch. Base command to train Pick policy: 
```
python habitat_baselines/run.py --exp-config habitat_baselines/config/rearrange/ddppo_pick.yaml --run-type train
```

erikwijmans:
Seems to be the same within variation due to non-deterministic inference/training.
image
(Blue is this PR, orange is current main).

ASzot (approved):
Since the PR does not affect training, I am happy with the changes.

PR #909, opened 2022-07-20
[WIP] Creating Static Manipulator Interface

laikhtewari:
## Motivation and Context

Habitat provides an easy way to build custom mobile manipulators, but this mobile interface doesn't easily extend to static manipulators. A static manipulator class was created as well as an example Franka Panda class to build support for static manipulators. Moving forward, it may make sense to have the mobile manipulator class inherit the "arm" functionality from a static manipulator and add in additional mobile base / wheel functionality. 

## How Has This Been Tested

A Franka example config (configs/tasks/franka_point.yaml) was created with an RGB sensor as well as an example script that randomly samples the action space (examples/franka_example.py). This test illustrates how the base action space is still geared towards mobile manipulators and should be reworked.

## Types of changes

## Checklist


examples/franka_example.py
```diff
@@ -0,0 +1,50 @@
...
+    with habitat.Env(
+        config=habitat.get_config("configs/tasks/franka_point.yaml")
+    ) as env:
+        print("Environment creation successful")
+        observations = env.reset()  # noqa: F841
+        display_rgb(observations[SENSOR_KEY])
+
+        print("Agent acting inside environment.")
+        count_steps = 0
+        print(env.action_space)
```
1heart, line 39: Where is the action space defined?

If we can redefine it to remove the mobile manipulation actions easily, let's do so.

examples/franka_example.py
```diff
@@ -0,0 +1,50 @@
...
+    img_bgr = image[..., ::-1]
+    cv2.imshow("RGB", img_bgr)
+    _ = cv2.waitKey(0)
+
+
+def example():
+    # Note: Use with for the example testing, doesn't need to be like this on the README
+
+    with habitat.Env(
+        config=habitat.get_config("configs/tasks/franka_point.yaml")
```
1heart, line 31: I don't seem to see this file.

1heart (commented):
Seems like it's on the right track. Let's just file an issue to have mobile manipulator inherit from static manipulator and move forward with this implementation.

habitat/robots/static_manipulator.py
```diff
@@ -0,0 +1,458 @@
...
+        return np.clip(
+            pos,
+            self.params.ee_constraint[:, 0],
+            self.params.ee_constraint[:, 1],
+        )
+
+    @property
+    def arm_motor_forces(self) -> np.ndarray:
+        """Get the current torques on the arm joint motors"""
+        return np.array(self.sim_obj.joint_forces)
```
Skylion007, line 385: As a more meta point, we may want to update the sim python bindings to just return numpy arrays in the future.

examples/franka_example.py
```diff
@@ -0,0 +1,50 @@
...
+    ) as env:
+        print("Environment creation successful")
+        observations = env.reset()  # noqa: F841
+        display_rgb(observations[SENSOR_KEY])
```
Skylion007, line 35: This script can be run from the pytest suite with a little more effort. The main thing missing is that there needs to be a way to disable displaying images so that it doesn't cause pytest to hang on headless machines (see test_examples.py for more details).

examples/franka_example.py
```diff
@@ -0,0 +1,50 @@
...
+    _ = cv2.waitKey(0)
+
+
+def example():
+    # Note: Use with for the example testing, doesn't need to be like this on the README
+
+    with habitat.Env(
+        config=habitat.get_config("configs/tasks/franka_point.yaml")
+    ) as env:
+        print("Environment creation successful")
```
Skylion007, line 33: May want to use habitat_logger?

examples/franka_example.py
```diff
@@ -0,0 +1,50 @@
...
+
+import habitat
+
+SENSOR_KEY = "rgb"
+
+
+def display_grayscale(image):
+    img_bgr = np.repeat(image, 3, 2)
+    cv2.imshow("Depth Sensor", img_bgr)
+    _ = cv2.waitKey(0)
```
0wu, line 18: why store return value?

habitat/robots/static_manipulator.py
```diff
@@ -0,0 +1,488 @@
...
+        self.sim_obj.joint_positions = joint_positions
+
+    def _validate_arm_ctrl_input(self, ctrl: List[float]):
+        """
+        Raises an exception if the control input is NaN or does not match the
+        joint dimensions.
+        """
+        if len(ctrl) != len(self.params.arm_joints):
+            raise ValueError(
+                "Control dimension does not match joint dimension"
```
1heart, line 329:
```suggestion
                f"Control dimension ({len(ctrl)}) does not match joint dimension ({len(self.params.arm_joints)})"
```

or something along those lines.

habitat/robots/franka_robot.py
```diff
@@ -0,0 +1,53 @@
...
+        urdf_path,
+        sim,
+        limit_robo_joints=True,
+    ):
+        super().__init__(
+            self._get_franka_params(),
+            urdf_path,
+            sim,
+            limit_robo_joints,
+        )
```
1heart, line 53:
```suggestion
    def __init__(
        self, *args, **kwargs
    ):
        super().__init__(
            params=self._get_franka_params(), *args, **kwargs)
```

Would this be equivalent? I think it's nice to make clear what's the only thing being overwritten.

examples/franka_example.py
```diff
@@ -0,0 +1,58 @@
...
+                display_rgb(observations[SENSOR_KEY])
+            count_steps += 1
+        logger.info("Episode finished after {} steps.".format(count_steps))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--no-render", action="store_true", default=False)
+    args = parser.parse_args()
+    example(args)
```
1heart, line 58:
```suggestion
def example(render):
    # Note: Use with for the example testing, doesn't need to be like this on the README

    with habitat.Env(
        config=habitat.get_config("configs/tasks/franka_point.yaml")
    ) as env:
        logger.info("Environment creation successful")
        observations = env.reset()  # noqa: F841
        if render:
            display_rgb(observations[SENSOR_KEY])

        logger.info("Agent acting inside environment.")
        count_steps = 0
        logger.info(env.action_space)
        while not env.episode_over:
            action = env.action_space.sample()
            logger.info(f"Executing action: {action}")
            observations = env.step(action)  # noqa: F841
            if not render:
                display_rgb(observations[SENSOR_KEY])
            count_steps += 1
        logger.info("Episode finished after {} steps.".format(count_steps))


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--no-render", action="store_true", default=False)
    args = parser.parse_args()
    render = not args.no_render
    example(render=render)
```

Makes the function slightly more reusable by isolating singleton config file away from the function.

1heart (approved):
lgtm, minor nits

PR #925, opened 2022-08-11
Added Franka blockstacking example

laikhtewari:
## Motivation and Context

Creates a new franka blockstacking example using the static manipulator class. 
[T122948310](https://www.internalfb.com/intern/tasks/?t=122948310)
Screen Shot 2022-08-24 at 3 48 56 PM

## How Has This Been Tested
Running the example (with a display) succeeds in creating an environment with 6 blocks that can be manipulated
`DISPLAY=:100 python examples/franka_blockstacking_example.py`

Screen Shot 2022-08-15 at 4 14 41 PM

## Types of changes

## Checklist


1heart (approved):
Overall lgtm, minor nits:

- Very similarly to the prior PR involving `franka_example.py`, it would be nice to (1) isolate the output of argparse from the function itself, (2) avoid the double negative `not args.no_render`.
- You could also import functions `display_grayscale`, `display_rgb` relatively there, to avoid duplication.
- Add a screenshot of this working.

Skylion007:
Can this run, without rendering as a unit test for test_examples. See https://github.com/facebookresearch/habitat-lab/blob/main/test/test_examples.py for how to hook up the script.

examples/franka_blockstacking_example.py
```diff
@@ -0,0 +1,62 @@
...
+
+import numpy as np
+from display_utils import display_rgb
+
+import habitat
+from habitat.core.logging import logger
+
+SENSOR_KEY = "rgb"
+CONFIG_FILE = "configs/tasks/franka_point.yaml"
+BLOCK_CONFIGS_PATH = "data/test_assets/objects/nested_box"
```
aclegg3, line 17: This glb is mainly intended for testing. You should use the "cubeSolid" primitive instead.

From the interactivity tutorial:
```
# Do this once:
# Get a default cube primitive template
cube_handle = obj_attr_mgr.get_template_handles("cubeSolid")[0]
cube_template_cpy = obj_attr_mgr.get_template_by_handle(cube_handle)
# Modify the template's configured scale
cube_template_cpy.scale = np.array([0.2, 0.2, 0.2])
# Register the modified template under a new name
obj_attr_mgr.register_template(cube_template_cpy, "my_scaled_cube")

#Do this when you want a new instance:
occluder_obj = rigid_obj_mgr.add_object_by_template_handle("my_scaled_cube")
```

aclegg3 (commented):
Looks good overall. 
For this test to run on CI we need to host the panda_arm.urdf somewhere it can be downloaded publicly (assuming the license is amenable).

.circleci/config.yml
```diff
@@ -200,6 +200,17 @@ jobs:
                 wget http://dl.fbaipublicfiles.com/habitat/habitat-test-scenes.zip
                 unzip habitat-test-scenes.zip
                 rm habitat-test-scenes.zip
+                cd -
+              fi
+              if [ ! -f ./habitat-sim/data/robots/franka_panda/panda_arm.urdf ]
```
aclegg3, line 205: Let's refactor this to use the datasets_downloader from habitat-sim instead.

examples/franka_blockstacking_example.py
```diff
@@ -0,0 +1,68 @@
...
+        logger.info("Environment creation successful")
+
+        obj_templates_mgr = env.sim.get_object_template_manager()
+        rigid_obj_mgr = env.sim.get_rigid_object_manager()
+
+        cube_handle = obj_templates_mgr.get_template_handles("cubeSolid")[0]
+        cube_template_cpy = obj_templates_mgr.get_template_by_handle(
+            cube_handle
+        )
+        cube_template_cpy.scale = np.array([0.2, 0.2, 0.2])
```
mathfac, line 34: To keep example clean let's use const `BLOCK_SCALE= 0.2`
```suggestion
        cube_template_cpy.scale = np.array([BLOCK_SCALE * 3])
```

examples/franka_blockstacking_example.py
```diff
@@ -0,0 +1,68 @@
...
+        obj_templates_mgr.register_template(
+            cube_template_cpy, "my_scaled_cube"
+        )
+
+        for _ in range(6):
+            obj = rigid_obj_mgr.add_object_by_template_handle("my_scaled_cube")
+            offset = (np.random.random(3) * 2 - 1) * BLOCK_LOC_RANGE
```
mathfac, line 41: To be more illustrative let's create consts for "2" and "-1" numbers to describe where it comes from.

examples/franka_blockstacking_example.py
```diff
@@ -0,0 +1,68 @@
...
+        cube_handle = obj_templates_mgr.get_template_handles("cubeSolid")[0]
+        cube_template_cpy = obj_templates_mgr.get_template_by_handle(
+            cube_handle
+        )
+        cube_template_cpy.scale = np.array([0.2, 0.2, 0.2])
+        obj_templates_mgr.register_template(
+            cube_template_cpy, "my_scaled_cube"
+        )
+
+        for _ in range(6):
```
mathfac, line 39: `BLOCKS_NUM = 6`
as well

mathfac (approved):
Thank you for adding this example, some comments about code clarity.

examples/franka_blockstacking_example.py
laikhtewari: This line specifies that the blocks should be placed randomly and uniformly around BLOCK_LOC_CENTER, so I think making these additional constants may make actually obfuscate the intention. Do you think the following is clearer or do you still think that I should make the 2 and -1 new constants?

`offset = np.random.random(3) * BLOCK_LOC_RANGE * 2 - BLOCK_LOCK_RANGE`

PR #934, opened 2022-08-26
[WIP] Loading Fremont Space

ASzot:
# Setup Instructions
* Download Fremont dataset, rename the folder to `fremont`, and place under `data/` in the base Habitat Lab directory. 
* Folders like `data/fremont/configs/scenes/` should exist.

Generate the dataset with: `python habitat/datasets/rearrange/run_episode_generator.py  --run --config habitat/datasets/rearrange/configs/fremont.yaml --num-episodes 1 --out data/datasets/fremont/test.json.gz --verbose --limit-scene-set scene_train_split`

Load the dataset into the interactive play script with: `python examples/interactive_play.py  --never-end --cfg configs/tasks/rearrange/rearrange_fremont.yaml`

### DIFF
--- a/configs/tasks/rearrange/rearrange_fremont.yaml
+++ b/configs/tasks/rearrange/rearrange_fremont.yaml
@@ -1,0 +1,95 @@
+# Config for empty task to explore the scene.
+ENVIRONMENT:
+    MAX_EPISODE_STEPS: 0
+DATASET:
+    TYPE: RearrangeDataset-v0
+    SPLIT: train
+    DATA_PATH: data/datasets/fremont/test.json.gz
+    SCENES_DIR: "data/fremont/"
+TASK:
+    TYPE: RearrangeEmptyTask-v0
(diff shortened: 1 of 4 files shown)


dhruvbatra:
This is coming together nicely. Can we also add information (links, how to load, etc) about these scene and episode datasets here: 

https://github.com/facebookresearch/habitat-sim/blob/main/DATASETS.md
https://github.com/facebookresearch/habitat-lab/blob/main/DATASETS.md

aclegg3:
> This is coming together nicely. Can we also add information (links, how to load, etc) about these scene and episode datasets here:
> 
> https://github.com/facebookresearch/habitat-sim/blob/main/DATASETS.md https://github.com/facebookresearch/habitat-lab/blob/main/DATASETS.md

We are not publicly hosting this dataset yet. If we're planning to add them to the public datasets readmes then maybe we should put them on S3 also?

PR #945, opened 2022-09-10
Object rearrangement: pick and place policies

yvsriram:
## Motivation and Context

The current rearrangement task provides the agent with geometric coordinates of the point where the object is to be picked and where it is to be dropped. These changes replace the geometric prior with a categorical prior.

## How Has This Been Tested

Tested on rearrange_easy episodes with additional category fields.

## Types of changes

## Checklist


habitat_baselines/config/default.py
```diff
@@ -65,7 +65,7 @@
 # The name of the project on W&B.
 _C.WB.PROJECT_NAME = ""
 # Logging entity (like your username or team name)
-_C.WB.ENTITY = ""
+_C.WB.ENTITY = "language-rearrangement"
```
ykarmesh, line 68: You will also need to make the following change in this file for wandb to work properly:
```_C.WRITER_TYPE = "wb"```
```_C.WB.PROJECT_NAME = "object_rearrangement"```

Comment (ykarmesh) [reply habitat_baselines/config/default.py:68]: Also, it is recommended that you use a novel `WB.RUN_NAME` for every run that you conduct so its easier to visualize the results

Comment (mathfac) [reply habitat_baselines/config/default.py:68]: Not sure it should be as default value for whole the Habitat Baselines package.

**Review by mathfac — COMMENTED** [review]
Glad to see the PR adding new task and dataset to the Habitat Lab. Several comments.
We need tests for Habitat baselines (training config test) and Habitat lab part of the task (loading task in CI etc.).

File: habitat_baselines/config/default.py (hunk 1 above)
Comment (ykarmesh) [reply habitat_baselines/config/default.py:68]: Hey Oleksandr, we dont plan to merge this PR into main. This PR is to just track the changes we are doing on our side

File: habitat-lab/habitat/config/habitat/task/rearrange/cat_pick.yaml [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+    - distance_to_goal
+    - distance_to_goal_reward
+    - pick_success
+    - pick_reward
+    - did_violate_hold_constraint
+    - num_steps
+
+  - /habitat/task/lab_sensors:
+    - target_start_sensor
+    - joint_sensor
```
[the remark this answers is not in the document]
yvsriram: Currently, geo_place and cat_place do not provide the agent with joint_velocity_sensor. But we can definitely add them

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -344,6 +344,38 @@ class DistToNavGoalSensorConfig(LabSensorConfig):
     type: str = "DistToNavGoalSensor"
 
 
+@dataclass
+class ObjectCategorySensorConfig(LabSensorConfig):
+    type: str = "ObjectCategorySensor"
+    goal_spec_max_val: int = 50
```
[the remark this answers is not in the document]
yvsriram: Removed this. Determining the value from episodes.

habitat-lab/habitat/config/habitat/task/rearrange/cat_place.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - /habitat/task/actions:
+    - arm_action
+    - base_velocity
+  - /habitat/task/measurements:
+    - robot_force
+    - force_terminate
+    - robot_colls
+    - end_effector_to_rest_distance
+    - end_effector_to_goal_distance
+    - object_to_goal_distance
```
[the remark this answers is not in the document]
yvsriram: Dense reward to a random location on the receptacle (currently using the geometric goal). There is an option to switch this to a sparse reward.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -108,6 +109,30 @@ def __init__(self, config: "DictConfig"):
...
+            )
+            self.recep_to_parent_obj[r.name] = r.parent_object_handle
+            self.receptacles[r.name] = r
+        self.rec_handle_to_rec_obj = self.get_rec_handle_to_rec_obj()
+
+    def get_rec_handle_to_rec_obj(self):
+        handle_to_obj = {}
+        rom = self.get_rigid_object_manager()
+        for handle, ro in rom.get_objects_by_handle_substring().items():
+            if "frl" in handle:
```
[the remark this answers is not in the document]
yvsriram: This is not needed anymore. Removed this.

habitat-lab/habitat/config/habitat/task/rearrange/cat_place.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - obj_at_goal
+    - obj_anywhere_on_goal
+    - place_success
+    - place_reward
+    - num_steps
+  - /habitat/task/lab_sensors:
+    - target_start_sensor
+    - goal_sensor
```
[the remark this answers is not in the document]
ykarmesh: Yeah, lets keep the obs_keys consistent in the configs. Otherwise people will keep getting confused.

habitat-lab/habitat/config/habitat/task/rearrange/cat_place.yaml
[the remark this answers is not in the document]
ykarmesh: Using a random location on the receptacle as the goal would cause the agent to learn some spurious correlations. The correct distance to goal in this case would be the distance to the sampling volume. But with the new receptacle annotation codebase, I would expect that code to change. So we can keep using the sparse reward for now I think

habitat-lab/habitat/tasks/rearrange/sub_tasks/place_task.py
```diff
@@ -23,6 +23,13 @@ def _should_prevent_grip(self, action_args):
             and action_args["grip_action"] >= 0
         )
 
+    def get_spawn_recs(self, sim, episode):
+        return [
+            r
+            for r, rec in sim.receptacles.items()
+            if episode.goal_recep_category in rec.parent_object_handle
```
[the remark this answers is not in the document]
yvsriram: Yeah, I am matching to a receptacle category field now. We will need to add this to the receptacle metadata files or maintain a separate mapping.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -981,15 +982,34 @@ def reset_metric(self, episode, *args: Any, **kwargs: Any):
...
+                self.robot_id
+            ).robot.ee_transform.translation
+        else:
+            current_position = self._sim.get_agent_state().position
 
         if self._previous_position is None or not np.allclose(
             self._previous_position, current_position, atol=1e-4
         ):
-            if self._config.distance_to == "POINT":
+            if self._config.distance_to == "EUCLIDEAN_POINT":
```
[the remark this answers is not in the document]
ykarmesh: Ok, I might change this with Sergio's reward

habitat-lab/habitat/config/habitat/task/rearrange/cat_pick.yaml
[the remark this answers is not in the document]
ykarmesh: Lets do an ablation for this once this PR is ready. I can run it on the cluster. Can you remind me if I forget?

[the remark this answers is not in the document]
yvsriram: Sure

habitat-baselines/habitat_baselines/common/tensorboard_utils.py
```diff
@@ -128,10 +128,30 @@ def __init__(
             raise ValueError(
                 "Requested to log with wandb, but wandb is not installed."
             )
+
         if resume_run_id is not None:
             wb_kwargs["id"] = resume_run_id
             wb_kwargs["resume"] = "must"
 
+        else:
+            wandb_id = wandb.util.generate_id()
```
ASzot, line 137: What are all these changes for? Wandb should support resuming on main now already.

yvsriram: Yeah, these are for using the same wandb logs after resuming a previous run (after a reque or for evaluation). It looks like this is not supported on the main. The current main is creating separate wandb runs.

habitat-baselines/habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -326,6 +348,53 @@ def __init__(
             )
             rnn_input_size += 32
 
+        if ObjectEmbeddingSensor.cls_uuid in observation_space.spaces:
```
ASzot, line 351: Pass these via the fuse_keys in the config so you don't have to hand define these sensors as input.

habitat-baselines/habitat_baselines/rl/ddppo/policy/resnet_policy.py
```diff
@@ -283,17 +299,23 @@ def __init__(
             goal_sensor_keys = {
                 IntegratedPointGoalGPSAndCompassSensor.cls_uuid,
                 ObjectGoalSensor.cls_uuid,
+                ObjectCategorySensor.cls_uuid,
```
ASzot, line 302: Same comment as before. We should not hard code these but read their uuids from a config field. This is important when we want to run experiments with different sensor inputs.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -363,6 +367,15 @@ def parse_receptacles_from_user_config(
             )
             receptacle_scale = ao_uniform_scaling * sub_config.get("scale")
 
+            # TODO: parse this from user config
+            category = re.sub(
```
ASzot, line 371: We are anyways saving the `parent_object_handle` in the regular rearrange dataset. Can't we just use this information and not have to change the dataset format for this object rearrange task? I would rather make no changes to the dataset code and have extracting the category be on the task side. Then we don't have to worry about managing different dataset instances.

habitat-baselines/habitat_baselines/rl/ddppo/policy/resnet_policy.py
yvsriram: Currently, the uuids are actually read from a config field (gym.obs_keys) and I control the sensors that are passed. Any new 1d sensors will also get automatically appended to fuse_keys. 

But I will remove the ObjectEmbeddingSensor from here as you suggested.

habitat-baselines/habitat_baselines/rl/ddppo/policy/resnet_policy.py
yvsriram: While ObjectEmbeddingSensor, which is a frozen CLIP embedding, can be passed as fuse_keys, the categorical inputs cannot be. The categorical inputs learn an embedding dictionary which needs to be stored in PointNavResNetPolicy, requiring this additional handling. For this, I just followed the handling of the ObjectGoal sensor that ObjectNav uses. I believe eventually we will be using frozen-embedding-based sensors for objects/receptacles so we will remove the categorical sensors altogether in future. I am keeping them now for experimentation.

For ObjectEmbeddingSensor, I will switch to fuse_keys as you suggested.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
yvsriram: This particular change actually just introduces an optional "category" field for the receptacles in the scene. It does not affect the dataset format. But I did make changes to the dataset format.

**On dataset format**
The [newly added fields](https://github.com/yvsriram/habitat-lab/blob/6429604ba17b2f1d17d004c8e036c27b59bf101f/habitat-lab/habitat/datasets/rearrange/rearrange_dataset.py#L55) are goal specifications (object and receptacle category names) and possible solutions (valid object and receptacles - in the format ObjectNav uses). Currently, I am adding these in a [separate script](https://github.com/yvsriram/lang-rearrange-scripts/blob/master/datasets/add_cat_fields_to_episodes.py) because I thought we would handle this as part of the dataset generation. But this code can definitely be pulled into the task specification. 

**Why a new dataset format may be more appropriate?**
- This would allow the episodes to be modified flexibly. 
    - We can support a range of tasks by only modifying the object goal and possible candidate objects (solutions). For example, in the future, we can directly handle goal specification that includes object attributes like "red colored bottle".
    - We may want to preprocess and store the valid goal viewpoints (for navigation skills) in episodes similar to ObjectNav.
- The new fields can be optional (as they are currently) allowing the new ObjectRearrangeDataset to be also used for loading geometric rearrange datasets. So we needn't worry about handling different dataset instances.
- I feel category extraction for objects/receptacles is better handled independently of the task. We actually may not be extracting the receptacle category from handle names and may use custom categories.


So I think for now, we can continue to generate GeoGoal rearrange episodes and use a separate script to append fields needed for the task. However, we should quickly add these changes to the dataset generation script as well. Let me know if you think otherwise.

habitat-baselines/habitat_baselines/rl/ddppo/policy/resnet_policy.py
ASzot: I see, that makes sense.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
ASzot: Thanks for the explanation. I agree with what you said. It makes sense to have this as a separate script as a "geometric rearrange dataset add on". Nothing to change here then. Your current code is good for this. Just be sure you document these steps of creating the dataset somewhere since this is a new way to create the dataset.

habitat-lab/habitat/config/habitat/task/rearrange/cat_place.yaml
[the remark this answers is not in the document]
yvsriram: For now, I set sparse reward as True by default. 

For the dense reward, the distance would be from the agent's position to which point on the sampling volume?

- I do think the current location I am using: the GeoGoal place locations are entirely random (I will verify by plotting the goal locations in the episodes for a particular receptacle). By rewarding based on random locations, the hope was that the agent will learn to place the object at some reasonable location. But there's a possibility that rewarding based on random locations may confuse the agent on where to place. I am still testing this "place" in the latest setup and will try to understand the behaviour the agent learns.

- Alternatively, we can reward the agent based on the distance to the closest target location where the object can be safely placed. For this, we may need to take into account the size of the object in the agent's hand and also other objects on the receptacle. I can implement this. For now, I will test (with/without scale) first if one of the sparse reward or the current dense reward work.

habitat-baselines/habitat_baselines/common/tensorboard_utils.py
yvsriram: In line with the recent changes in `main`, I made the fixes for loading the run_id from the resume state that gets saved.

habitat-lab/habitat/config/habitat/task/rearrange/cat_place.yaml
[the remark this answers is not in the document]
ykarmesh: The second option could make sense. We can just assume that the max object size is 1 ft in each direction and just find the location which is 1ft away from the receptacle edge (while handling the case that there might be receptacles with less than 1ft of length or width)

PR #948, opened 2022-09-12
Refactor mobile_manipulator and static_manipulator via robot_manipula…

jimmytyyang:
…tor along with the test test_static_robot_wrapper

## Motivation and Context

There are two files that sharing the same code functions for controlling the robots: mobile_manipulator.py and static_manipulator.py. We would like to refactor these two files so that they are organized.

https://github.com/facebookresearch/habitat-lab/pull/909

## How Has This Been Tested

A unit test called test_robot_wrapper.py already exists for the mobile manipulator platform, [Fetch](https://github.com/facebookresearch/habitat-lab/blob/main/habitat/robots/fetch_robot.py), which can validate the refactor. In addition, a new unit test called test_static_robot_wrapper.py is added for the static manipulator platform, [Franka](https://github.com/facebookresearch/habitat-lab/blob/main/habitat/robots/franka_robot.py). Note that the Franka's urdf file (panda_arm.urdf) does not contain grippers. So the parameter of the gripper_joints in [franka_robot.py](https://github.com/facebookresearch/habitat-lab/blob/main/habitat/robots/franka_robot.py) should be an empty list for this unit test.

## Types of changes

## Checklist

- [x ] I have completed my CLA (see **CONTRIBUTING**)


test/test_static_robot_wrapper.py
```diff
@@ -0,0 +1,373 @@
...
+@pytest.mark.skipif(
+    not osp.exists("data/robots/franka_robot"),
+    reason="Test requires Franka robot URDF and assets.",
+)
+@pytest.mark.skipif(
+    not habitat_sim.built_with_bullet,
+    reason="Robot wrapper API requires Bullet physics.",
+)
+
+def test_franka_robot_wrapper():
```
aclegg3, line 274: To reduce duplication of the boiler plate code setting up and making configurations, let's move this function into test_robot_wrapper.py.

habitat/robots/robot_manipulator.py
```diff
@@ -0,0 +1,448 @@
...
+            self.params.ee_constraint[:, 0],
+            self.params.ee_constraint[:, 1],
+        )
+
+    #############################################
+    # WHEEL RELATED
+    #############################################
+
+    # This depends on the robot type.
+
```
aclegg3, line 358: remove these comment tags

habitat/robots/robot_manipulator.py
```diff
@@ -0,0 +1,448 @@
...
+    #############################################
+    # BASE RELATED
```
aclegg3, line 360: What is the rational for base state properties in Manipulator class? Does this provide the global state of the root link of the arm? Should we remove the setters instead of leaving them here?

aclegg3 (commented):
Great to see reduced code duplication. Let's discuss some alternative designs before committing to inheritance as the solution.

aclegg3:
To address the `pre-commit` and `python_lint` CI failures, you can install pre-commit locally to auto format your commits:
```
pip install pre-commit
pre-commit install
# when you commit, this will automatically reformat the files to conform to style requirements:
git commit -a
```

jimmytyyang:
Due to the fact that the current design uses the setter approach, it is hard to pass additional arguments for selecting which arms to control. In addition, currently, we are assuming users need to define their own URDF in a single file, which means that manipulator should handle all the arms at once during initialization. For these two reasons, we do not support selecting the arms, and the user should send all the control commands for all arms once.

aclegg3:
> Due to the fact that the current design uses the setter approach, it is hard to pass additional arguments for selecting which arms to control.

For reference:  we are currently defining the API through properties which don't allow for parameters. We'll follow-up with a separate PR separating the property API from a functional getter/setter API which takes an index parameter. The properties will call the functions with index=0.

habitat/robots/robot_manipulator.py
jimmytyyang: moved base related functions to the robot base module

test/test_static_robot_wrapper.py
jimmytyyang: done!

habitat/robots/robot_manipulator.py
jimmytyyang: done!

habitat/robots/mobile_manipulator.py
```diff
@@ -104,7 +99,7 @@ class MobileManipulatorParams:
     base_link_names: Set[str]
 
 
-class MobileManipulator(RobotInterface):
+class MobileManipulator(RobotManipulator):
```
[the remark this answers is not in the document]
jimmytyyang: Use multiple interferences to design modules

habitat/robots/static_manipulator.py
```diff
@@ -65,421 +65,58 @@ def __init__(
...
-            self.params.ee_constraint[:, 0],
-            self.params.ee_constraint[:, 1],
-        )
-
-    @property
-    def arm_motor_forces(self) -> np.ndarray:
-        """Get the current torques on the arm joint motors"""
-        return np.array(self.sim_obj.joint_forces)
+        Manipulator.reconfigure(self)
+        RobotBase.reconfigure(self)
```
aclegg3, line 109:
```suggestion
        Manipulator.update(self)
        RobotBase.update(self)
```

habitat/robots/static_manipulator.py
```diff
@@ -65,421 +65,58 @@ def __init__(
...
-        for (
-            motor_id,
-            joint_id,
-        ) in self.sim_obj.existing_joint_motor_ids.items():
-            self.joint_motors[joint_id] = (
-                motor_id,
-                self.sim_obj.get_joint_motor_settings(motor_id),
-            )
-
     def _get_translation_from_htm(self, mat: np.ndarray) -> np.ndarray:
```
aclegg3, line 120: This doesn't seem to be used anywere

aclegg3 (commented):
I think the overall design is sound. 

I'm a bit surprised to see that MobileManipulator and StaticManipulator reduce to a difference in parameter structures. 
Maybe this is because RobotBase class is both mobile base and static base. 
Is there any way we can move static base stuff (maybe basic root get/set logic) to RobotInterface so StaticManipulator does not need to inherit from RobotBase?

Then RobotBase can specifically handle mobile base functions like moving on the navmesh and the division makes more sense.

habitat/robots/robotbase.py
```diff
@@ -0,0 +1,176 @@
...
+
+import magnum as mn
+import numpy as np
+
+from habitat.robots.robot_interface import RobotInterface
+from habitat_sim.physics import JointMotorSettings
+from habitat_sim.simulator import Simulator
+
+
+class RobotBase(RobotInterface):
```
aclegg3, line 16: File should be robot_base.py with underscore.

jimmytyyang: Done!

habitat/robots/static_manipulator.py
jimmytyyang: Done!

habitat/robots/robot_manipulator.py
```diff
@@ -0,0 +1,430 @@
...
+
+import magnum as mn
+import numpy as np
+
+from habitat.robots.robot_interface import RobotInterface
+from habitat_sim.physics import JointMotorSettings
+from habitat_sim.simulator import Simulator
+
+
+class RobotManipulator(RobotInterface):
```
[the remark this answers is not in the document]
jimmytyyang: Remove it, done!

habitat/robots/robot_base.py
```diff
@@ -0,0 +1,176 @@
...
+    @property
+    def base_transformation(self):
+        return self.sim_obj.transformation
+
+    def is_base_link(self, link_id: int) -> bool:
+        return (
+            self.sim_obj.get_link_name(link_id) in self.params.base_link_names
+        )
+
+    #############################################
```
mathfac, line 163: While I like your sections devisions here we don't have it as a standard for other source files. To keep code unified let's avoid it here. Thank you.

habitat/robots/static_manipulator.py
```diff
@@ -65,424 +65,51 @@ def __init__(
...
-    def base_rot(self, rotation_y_rad: float):
-        raise NotImplementedError(
-            "Setting the base rotation of a static manipulator is not permitted."
-        )
-
-    @property
-    def base_transformation(self):
-        return self.sim_obj.transformation
-
-    #############################################
```
mathfac, line 422: Ow, it was already used before, anyway let's fix it.

jimmytyyang: Done, thank you!

habitat/robots/robot_base.py
jimmytyyang: Done!

aclegg3 (commented):
I think this is pretty much ready to go.

I still think StaticManipulator should not inherit RobotBase. This separation would better differentiate the role and API associated with a mobile base from the lack thereof. @jimmytyyang do you believe otherwise? If so, what is your opinion here?

jimmytyyang:
Hi @aclegg3, thanks for the comments! I agree with you that the separation can differentiate the role and API. I updated the code, thank you!

test/test_robot_wrapper.py
```diff
@@ -407,3 +417,119 @@ def test_fetch_robot_wrapper(fixed_base):
                 "test_fetch_robot_wrapper__fixed_base=" + str(fixed_base),
                 open_vid=True,
             )
+
+
+@pytest.mark.skipif(
+    not osp.exists("data/robots/hab_franka"),
```
aclegg3, line 423: Also, this should be the path which is created from downloading [this asset](https://github.com/facebookresearch/habitat-sim/blob/83fe46b9695bc9a8d4187f3d739ddbd1b2889619/src_python/habitat_sim/utils/datasets_download.py#L129). I think that's actually `"data/robots/franka_panda`

Once you update that, check that this test isn't skipped on CI. It should be getting the robot assets from the downloader util already.

test/test_robot_wrapper.py
```diff
@@ -407,3 +417,119 @@ def test_fetch_robot_wrapper(fixed_base):
...
+        ground_plane.motion_type = habitat_sim.physics.MotionType.STATIC
+
+        # compute a navmesh on the ground plane
+        navmesh_settings = habitat_sim.NavMeshSettings()
+        navmesh_settings.set_defaults()
+        sim.recompute_navmesh(sim.pathfinder, navmesh_settings, True)
+        sim.navmesh_visualization = True
+
+        # add the robot to the world via the wrapper
+        robot_path = "data/robots/hab_franka/robots/panda_arm.urdf"
```
aclegg3, line 463: Similarly, check that this path is correct. I think it might be "data/robots/franka_panda/panda_arm.urdf"

test/test_robot_wrapper.py
jimmytyyang: Done!

PR #971, opened 2022-10-04
[BUG] pip install in README.md points to the wrong folder

vmoens:
## Motivation and Context

The `pip install` command in the README.md reads
```
cd habitat-lab
pip install -e habitat-lab
```
but I guess that what is meant is
```
cd habitat-lab
pip install -e .
```


vmoens:
I believe this can be closed as the folder is now structured in two separate libs (habitat-lab and habitat-baselines.
However the `git clone --branch stable` does not work (as the stable release does not have that feature yet)
thoughts @vincentpierre?

PR #1014, opened 2022-11-22
[v0.2.3] [Stretch] play stretch demo

jimmytyyang:
## Motivation and Context

 

Provide a demo for running Stretch. Download the stretch asset using data loader in Habitat sim. 

Then run 
```python examples/interactive_play.py --never-end --cfg [path to stretch]/play_stretch.yaml```, 

then you can use a keyboard to control Stretch.

https://user-images.githubusercontent.com/55121504/203428657-bbe63057-7a44-4072-b9af-7029a5b7ff43.mp4

## How Has This Been Tested

Locally

## Types of changes

## Checklist

### DIFF
--- a/examples/interactive_play.py
+++ b/examples/interactive_play.py
@@ -72,7 +72,7 @@ try:
 except ImportError:
     pygame = None
 
-DEFAULT_CFG = "benchmark/rearrange/play.yaml"
+DEFAULT_CFG = "benchmark/rearrange/play_stretch.yaml"
 DEFAULT_RENDER_STEPS_LIMIT = 60
 SAVE_VIDEO_DIR = "./data/vids"
 SAVE_ACTIONS_DIR = "./data/interactive_play_replays"
@@ -108,6 +108,10 @@ def get_input_vel_ctlr(
         ]
         arm_ctrlr = env.task.actions[arm_action_name].arm_ctrlr
         base_action = None
+    elif "stretch" in DEFAULT_CFG:
+        arm_action_space = np.zeros(10)
+        arm_ctrlr = None
+        base_action = [0, 0]
     else:
         arm_action_space = np.zeros(7)
         arm_ctrlr = None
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dhruvbatra** [comment]
The functionality and video both looks great! 

Can we get Noah to add some more texture to that base? It is looking very flat.

**Comment by jimmytyyang** [comment]
Thanks for the comment! I will sync with Noah and add more texture!

File: examples/interactive_play.py [hunk 1]
```diff
@@ -72,7 +72,7 @@
 except ImportError:
     pygame = None
 
-DEFAULT_CFG = "benchmark/rearrange/play.yaml"
+DEFAULT_CFG = "benchmark/rearrange/play_stretch.yaml"
```
aclegg3, line 75: Are we migrating our default to Stretch at this time? Otherwise this should be a comment option or terminal arg. @ASzot

aclegg3 (commented):
Generally looks good to me.

`interactive_play.py` changes seem a bit adhoc, but the entire script is like that, so fine for now. In the future we should try to clean this up for generic robot support.

examples/interactive_play.py
jimmytyyang: Thanks for this! I fixed this.

ASzot:
I am concerned about this mismatch of a kinematic action for Fetch but non-kinematic simulation. I synced with @jimmytyyang about this offline, but we should fix the simulation to also be kinematic, or else there will be weird behavior (like the arm will be able to penetrate surfaces easily). But maybe we can get this PR in now as is and work on that later?

jimmytyyang:
Hi @ASzot Thanks for the information! There is a discussion about if we should use kinematic or dynamics simulation. Fixing the simulation to be kinematic requires other substantial code changes such as how to detect collision and so on. We could get this PR in now to get the ball rolling as there is a lot of interest for people who want to try out Stretch, but I am open to any discussion! Thank you.

aclegg3:
Great point, @ASzot. Agreed, let's get the demo in now and then discuss the broader paradigm moving forward.

PR #1021, opened 2022-11-30
Unpin habitat-lab requirements

Skylion007:
## Motivation and Context

* Pinning the hydra version so strongly is not recommended and could prevent bugfixes and cause compatibility issues. None of the other requirements for habitat-lab are pinned (exception for protobuf and tensorboard for specific reasons). I would recommend unpinning them to prevent issues down the line. I would even be okay if you wanted to pin hydra only, but I don't see a reason to pin the minor version of either library. This will prevent us from getting potential bugfixes and support for newer Python versions.

## How Has This Been Tested
* Locally and with CI

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/requirements.txt
+++ b/habitat-lab/requirements.txt
@@ -10,5 +10,5 @@ imageio-ffmpeg>=0.2.0
 scipy>=1.0.0
 tqdm>=4.0.0
 numba>=0.44.0
-hydra-core==1.2.0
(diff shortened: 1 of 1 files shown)


Skylion007:
I don't have the bandwidth / knowledge of the omegaconf stuff to fix all of it. @rpartsey @vincentpierre mind taking a look when you get the chance?

PR #1050, opened 2022-12-22
Object Rearrangement: Using floorplanner scenes and new objects

yvsriram:
## Types of changes

- Training pick on floorplanner scenes [Done]
- New object splits [Done]
- Script and instructions for generating object rearrangement episodes [Done]

## Checklist


habitat-lab/habitat/datasets/rearrange/README.md
```diff
@@ -0,0 +1,15 @@
+## Generating episodes for Object Rearrangement task
+
+Given the standard GeoGoal rearrange episodes data at <data_dir>/<split>/<geogoal_episodes_name>.yaml, the script `convert_to_object_rearrange_episodes.py` can be used to generate the corresponding episodes for Object Rearrangement task. The command for using this script is as follows:
+
+```
+python habitat-lab/habitat/datasets/rearrange/convert_to_object_rearrange_episodes.py --data_dir <data_dir> --source_episodes_tag <geogoal_episodes_name> --target_episodes_tag <target_episodes_name> --obj_category_mapping_file /path/to/csv/with/object_category_mapping --rec_category_mapping_file /path/to/csv/with/receptacle_category_mapping
```
ykarmesh, line 6: Where is the `convert_to_object_rearrange_episodes` file?

yvsriram: Sorry, I renamed it. [habitat-lab/habitat/datasets/rearrange/modify_episodes_for_object_rearrange.py](https://github.com/facebookresearch/habitat-lab/pull/1050/files/0b08919c3951c0c030a4a94ed7fd342190201279#diff-aed663b40b1037a3932fbca619c2911cd788fcfef320f00835e5533bf8b9371d). I will update this.

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -1033,6 +1033,8 @@ class SimulatorConfig(HabitatBaseConfig):
     # ep_info is added to the config in some rearrange tasks inside
     # merge_sim_episode_with_object_config
     ep_info: Optional[Any] = None
+    # instance ids are recorded in the semantic sensor by adding obj_instance_id_start. This is added to avoid collision with scene semantic categories.
+    obj_instance_id_start: Optional[int] = 50
```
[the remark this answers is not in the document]
yvsriram: This is for oracle experiments and we will most likely remove this later. This is used to provide a mask showing the objects that can be picked. The semantic sensor is set up to record instance ids. These are later matched to the valid instance ids (that come from episodes) for getting the mask. 

In general, I am setting up a more general version of the task we are solving. The category is only used as a prior (for goal specification). The episodes provide the ids of candidate instances to be picked up. This would allow us to use this code directly when we extend to complex language goals including object attributes. We will only need to update the episodes. Also, this keeps the complexity of mapping and checking for "category" away from the task.

If the name "semantic_sensor" makes this confusing, I can see if I can create a new instance id sensor.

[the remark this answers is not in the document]
ykarmesh: Yeah IMO, we shouldn't modify existing sensors if that is going to change the functionality of the sensor.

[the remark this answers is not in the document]
cpaxton: Instance IDs seem useful to have, in general. Would this affect current downstream uses of the semantic sensor though? Could we just add a semantic + instance IDs sensor?

habitat-lab/habitat/datasets/rearrange/configs/rearrange_floorplanner.yaml
```diff
@@ -1,1329 +1,468 @@
...
+  - ASICS_GELLinksmaster_WhiteRasberryGunmetal
+  - Adrenaline_GTS_13_Color_DrkDenimWhtBachlorBttnSlvr_Size_50_yfK40TNjq0V
+  - PureConnect_2_Color_BlckBrllntBluNghtlfeAnthrct_Size_70
+  - TROCHILUS_BOOST
+  - Chelsea_BlkHeelPMP_DwxLtZNxLZZ
+  - ASICS_GELLinksmaster_WhiteCoffeeSand
+  - D_ROSE_ENGLEWOOD_II
+  - Court_Attitude
+  excluded_substrings: []
+- name: test_unseen_spatula
```
ASzot, line 305: To confirm, these names are matching our decided category names correct?

habitat-lab/habitat/datasets/rearrange/modify_episodes_for_object_rearrange.py
```diff
@@ -0,0 +1,427 @@
...
+            object_id = aom.get_object_id_by_handle(
+                receptacle.parent_object_handle
+            )
+        elif rom.get_library_has_handle(receptacle.parent_object_handle):
+            object_id = rom.get_object_id_by_handle(
+                receptacle.parent_object_handle
+            )
+        else:
+            object_id = -1
+        pos = receptacle.get_surface_center(sim)
```
ASzot, line 155: Why do we need the center position of the receptacle?

habitat-lab/habitat/config/default_structured_configs.py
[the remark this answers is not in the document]
yvsriram: I added Instance IDs to the semantic sensor space only for our rearrangement task. So, it should not affect other downstream applications. I think having a separate instance ID sensor is better. We needn't worry about conflicting output spaces.

habitat-lab/habitat/datasets/rearrange/configs/rearrange_floorplanner.yaml
yvsriram: Yes, I did a manual clean-up of category names for a subset of categories before creating this split. But we are yet to work on finalizing the category names for the complete set of objects.

yvsriram: Here's the [sheet](https://docs.google.com/spreadsheets/d/1EnAVC7YMdjA2s3xXKyabjNJdoB-b4Bd7mZTpK5hAMTE/edit?usp=sharing) containing the cleaned category names.

habitat-lab/habitat/datasets/rearrange/modify_episodes_for_object_rearrange.py
ykarmesh: > I think it makes sense to change this to be with respect to a random position on the receptacle for better robustness.
Aren't we already uniformly sampling a start position within 2m of the receptacle center position. From what I understand, you are proposing sampling of a position on the receptacle and then using that position as a center of a circle from which you will uniformly sample a point? I think this is going to bias the distribution of starting points. Do you think it is beneficial to have this bias?

habitat-lab/habitat/config/default_structured_configs.py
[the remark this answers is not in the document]
yvsriram: Keeping semantic + instance IDs as Chris suggested. (The sensor comes with GT semantics defined in the scene. I am adding the instance ids.) I am calling this a panoptic sensor now.

habitat-lab/habitat/datasets/rearrange/modify_episodes_for_object_rearrange.py
yvsriram: Yes, you are right. For pick to succeed, currently we require the nav to end at a point 2m from the center of receptacle. I was thinking that sampling robot position from a point 2m from any point on receptacle would make the Pick more robust - the nav can stop at a point that is close to any point on the receptacle. I think this would be the case when we switch to using viewpoints for Nav's success. But this may make the Pick training harder. For now, I am adding a config param to support and experiment with both.

I don't think this would bias the distribution of starting points because sampling a point on the receptacle is expected to be random.

habitat-lab/habitat/config/default_structured_configs.py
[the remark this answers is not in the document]
ykarmesh: sounds good

habitat-lab/habitat/datasets/rearrange/modify_episodes_for_object_rearrange.py
ykarmesh: > I don't think this would bias the distribution of starting points because sampling a point on the receptacle is expected to be random.

Lets say you have a square table of size 2x2 and you uniformly sample a point on the surface of this table, called p1. Once this point has been sampled, you create a circle of radius 2m and sample a point (p2) from the circle, again uniformly. The maximum distance of p2 from the table center is going to be 4m and p2 will only be sampled if the p1 was sampled at the edge of the table. 

On the other hand, if p2 is very close to the edge of the table, it could have been sampled from a lot of different p1s. So p2 will definitely have a bias.

yvsriram: Oh, got it now. Yes, p2 will be biased. Will keep the current implementation.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -232,6 +247,13 @@ def get_global_transform(self, sim: habitat_sim.Simulator) -> mn.Matrix4:
         # base class implements getting transform from attached objects
         return super().get_global_transform
 
+    def get_local_surface_center(
```
cpaxton, line 250: in the future please document functions like this - one comment is enough, explain how/when we're supposed to use it

habitat-lab/habitat/tasks/rearrange/rearrange_sensors.py
```diff
@@ -193,13 +193,21 @@ def _get_observation_space(self, *args, **kwargs):
         )
 
     def get_observation(self, observations, *args, episode, task, **kwargs):
-        obj_id = self._sim.scene_obj_ids[task.abs_targ_idx]
         if np.random.random() < self._config.blank_out_prob:
             return np.zeros_like(
-                observations["robot_head_semantic"], dtype=np.uint8
+                observations["robot_head_panoptic"], dtype=np.uint8
```
cpaxton, line 198: why are we removing semantic? should this be a config option?

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
yvsriram: Sure, will do that in future. For this function, I will add it [here](https://github.com/facebookresearch/habitat-lab/pull/1067) (the other PR)

PR #1068, opened 2023-01-09
[Bug Fix] Stretch robot

jimmytyyang:
## Motivation and Context

 Fixed parameters of Stretch: (1) The opening angle of gripper is set to be smaller so that the visual looks great. (2) The navmesh radius of Stretch is reduced from 0.3 to 0.25 meter. (3) The name of arm control action of Stretch was wrong. It is the dynamic control, but the function name said it was kinematic control. Fixed the naming issue to avoid confusion.

## How Has This Been Tested

 Ran test_robot_wrapper.py. The fixed base is set to be True for Stretch since Stretch falls down when the fixed base is not true and its gripper cannot be operated properly in that case.

## Types of changes

## Checklist


aclegg3 (approved):
LGTM

PR #1098, opened 2023-01-26
Episode generator fixes

JHurricane96:
## Motivation and Context

Fixes to object rearrange episode generator

## How Has This Been Tested

Manual testing by running episode generator and inspecting objects in the visual debugger.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
@@ -20,6 +20,7 @@ from habitat.datasets.rearrange.samplers.receptacle import (
     OnTopOfReceptacle,
     Receptacle,
     ReceptacleTracker,
+    TriangleMeshReceptacle,
     find_receptacles,
 )
 from habitat.sims.habitat_simulator.debug_visualizer import DebugVisualizer
@@ -261,12 +262,13 @@ class ObjectSampler:
                     new_object.rotation = (
                         habitat_sim.utils.common.random_quaternion()
                     )
-            new_object_height = (
-                new_object.root_scene_node.cumulative_bb.size()[1]
-            )
-            new_object.translation = new_object.translation + mn.Vector3(
-                0, new_object_height, 0
-            )
+            if isinstance(receptacle, TriangleMeshReceptacle):
+                new_object_height = (
+                    new_object.root_scene_node.cumulative_bb.size()[1]
+                )
+                new_object.translation = new_object.translation + mn.Vector3(
+                    0, new_object_height / 2, 0
+                )
 
             if isinstance(receptacle, OnTopOfReceptacle):
                 snap_down = False
@@ -304,6 +306,9 @@ class ObjectSampler:
                         f"Successfully sampled (snapped) object placement in {num_placement_tries} tries."
                     )
                     if not self._is_accessible(sim, new_object):
+                        logger.warning(
+                            f"Failed to navigate to {object_handle} on {receptacle.name} in {num_placement_tries} tries."
+                        )
                         continue
                     return new_object
 
--- a/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
@@ -261,6 +261,12 @@ class ObjectSampler:
                     new_object.rotation = (
                         habitat_sim.utils.common.random_quaternion()
                     )
+            new_object_height = (
+                new_object.root_scene_node.cumulative_bb.size()[1]
+            )
+            new_object.translation = new_object.translation + mn.Vector3(
+                0, new_object_height, 0
+            )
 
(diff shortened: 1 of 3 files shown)


habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -261,6 +261,12 @@ def sample_placement(
                     new_object.rotation = (
                         habitat_sim.utils.common.random_quaternion()
                     )
+            new_object_height = (
+                new_object.root_scene_node.cumulative_bb.size()[1]
+            )
+            new_object.translation = new_object.translation + mn.Vector3(
```
yvsriram, line 267: For  AABB receptacles, we have upper bounds on the max height the object is allowed to take. (example - for a drawer, these would correspond to the bounding box). There is a chance that the object can exceed the bounds with this addition. This could result in the object being placed on a receptacle surface that is above the intended receptacle. Maybe you can add a check: for AABBReceptacle check if the object is within the bounds, or we can make change to sample positions taking the object height into account.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -549,49 +573,13 @@ def import_tri_mesh_ply(ply_file: str) -> Tuple[List[mn.Vector3], List[int]]:
...
-                mesh_data[1].extend(indices[:-1])
-                mesh_data[1].extend(indices[1:])
-            else:
-                indices = [int(x) for x in lines[face_line].split(" ")[1:]]
-                mesh_data[1].extend(indices)
-
+    trimesh_data = trimesh.load(ply_file)
+    mesh_data[0].extend(mn.Vector3(vert) for vert in trimesh_data.vertices)
+    for face in trimesh_data.faces:
+        assert len(face) == 3, f"Faces must be triangles. '{ply_file}' {face}"
```
yvsriram, line 581: In the five scene dataset that Alex shared, there were meshes with a few faces that were quads. I had added code here to handle this issue. If we are sure, we will be moving away from the 5 scene set - this looks good to me.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -549,49 +573,13 @@ def import_tri_mesh_ply(ply_file: str) -> Tuple[List[mn.Vector3], List[int]]:
...
+        mesh_data[1].extend(map(int, face))
     return mesh_data
```
yvsriram, line 583: At the moment, we are converting from `.ply->Trimesh->Tuple[List, List]` and then converting this back to `Trimesh` in `TriangleMeshReceptacle`. This works for now and we can keep it. But it may be cleaner to return `Trimesh` here and maybe also use `Trimesh` for area-weighted sampling. But we can make this change later.

yvsriram:
Minor non-blocking issues. Otherwise, looks good to me.

habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
JHurricane96: I've changed this to do the shift only for mesh receptacles.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
JHurricane96: Yes, I think we should make this change but it may be a bit involved. We can leave it for later.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -549,49 +573,13 @@ def import_tri_mesh_ply(ply_file: str) -> Tuple[List[mn.Vector3], List[int]]:
...
-            if int(lines[face_line][0]) == 4:
-                indices = [int(x) for x in lines[face_line].split(" ")[1:]]
-                mesh_data[1].extend(indices[:-1])
-                mesh_data[1].extend(indices[1:])
-            else:
-                indices = [int(x) for x in lines[face_line].split(" ")[1:]]
-                mesh_data[1].extend(indices)
-
+    trimesh_data = trimesh.load(ply_file)
+    mesh_data[0].extend(mn.Vector3(vert) for vert in trimesh_data.vertices)
```
Skylion007, line 579:
```suggestion
    mesh_data[0].extend(map(mn.Vector3, trimesh_data.vertices))
```

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
yvsriram: SGTM

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
JHurricane96: Thanks, changed it.

yvsriram (approved):

PR #1108, opened 2023-02-01
TriangleMeshReceptacles and RearrangeGenerator Improvements

aclegg3:
## Motivation and Context

**New features include:**
- TriangleMeshReceptacle class - area weighted uniform sampling of 3D positions from an arbitrary triangle mesh surface. Can be loaded from **.PLY** format.
- Sampling efficiency improvement: Added option to remove unstable objects instead of throwing away the episode if minimum number of objects for each sample can still be satisfied.
- Improved reporting of sampling statistics and timings for RearrangeGenerator.
- Added debug images of unstable objects to improve receptacle/scene debugging.
- triangle sampling and "point in triangle" test added to geometry_utils and tested

**Changes include:**
- Receptacle debugging visualization now uses DebugLineRender instead of attaching wire-frames to objects.
- bugfix: max number of sample attempts were not reset after successful pairing.
- bugfix: save file paths for DebugVisualizer were not always propagated correctly.
- new and improved docstrings and typing

## How Has This Been Tested

Added unit tests specifically for parsing and sampling from `AABBReceptacle`s and `TriangleMeshReceptacle`s.
Must be tested against [habitat-sim PR 1994](https://github.com/facebookresearch/habitat-sim/pull/1994) with modified test assets.

## Types of changes

Docs change / refactoring / dependency upgrade
Bug fix (non-breaking change which fixes an issue)
New feature (non-breaking change which adds functionality)

## Checklist


.circleci/config.yml
```diff
@@ -191,6 +191,8 @@ jobs:
               export PATH=$HOME/miniconda/bin:/usr/local/cuda/bin:$PATH
               . activate habitat;
               cd habitat-sim
+              git fetch --all
+              git checkout receptacle-test-assets
```
aclegg3, line 195: Testing only: this will be removed before merge.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -290,9 +485,40 @@ def get_all_scenedataset_receptacles(sim) -> Dict[str, Dict[str, List[str]]]:
...
+        raise NotImplementedError("TODO: multi-mesh receptacle support.")
+
+    mesh_ix = 0
+    mesh = importer.mesh(mesh_ix)
+    assert (
+        mesh.primitive == mn.MeshPrimitive.TRIANGLES
+    ), "Must be a triangle mesh."
+
+    # zero-copy reference to importer datastructures
+    mesh_data: Tuple[List[mn.Vector3], List[int]] = (
```
mosra, line 510: If you'd want to be *exact* with the typing annotation, the actual type is [cr.containers.StridedArrayView1D](https://doc.magnum.graphics/python/corrade/containers/StridedArrayView1D/), not a `List`, but right now I don't have the `[]` "templating" implemented so saying `cr.StridedArrayView1D[mn.Vector3]` would probably cause Python to complain.... and without it it's maybe too nondescriptive (like `np.ndarray` is). I'll investigate what can be done on that front (implementing `__getitem__` on the typeclass? huh).

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -241,13 +274,169 @@ def add_receptacle_visualization(
...
+        dblr = sim.get_debug_line_render()
+        dblr.push_transform(self.get_global_transform(sim))
+        assert len(self.mesh_data[1]) % 3 == 0, "must be triangles"
+        for face in range(int(len(self.mesh_data[1]) / 3)):
+            verts = self.get_face_verts(f_ix=face)
+            for edge in range(3):
+                dblr.draw_transformed_line(
+                    verts[edge], verts[(edge + 1) % 3], color
+                )
+        dblr.pop_transform()
```
mosra, line 427: Just FYI, the dedicated tool for drawing wireframes (without having to manually compose them out of line segments) is the [MeshVisualizer](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1MeshVisualizerGL3D.html) shader. I think it was used somewhere in the C++ viewer already, but no idea how easy is it to make it available here. 

In case the shader would be available, you'd keep the original MeshData (instead of taking just views on positions and indices from it) and draw it directly with the shader.

aclegg3: Cool suggestion, I think this would be a cool way to do some debugging visualization in the future. Construct MeshData for the debug mesh and then register it for the MeshVisualizer shader. 
Is this exposed to python in some way that it could draw on top of an existing framebuffer or would it need to be built into the Habitat rendering pipeline on the C++ side?

mosra: The shader isn't exposed at the moment (but I could do that easily), and yes that would work as long as you either have the framebuffer currently bound or have a way to access & bind it for drawing.

On the other hand, such direct approach would be _very_ incompatible with the batch renderer ... so for easier porting in the future it's probably best to just leave it as-is :sweat_smile: The debug line drawing will have to get ported to the batch renderer in any case, and I could then maybe expand it to handle also wireframe meshes somehow.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -241,13 +274,169 @@ def add_receptacle_visualization(
...
+        :param parent_object_handle: The rigid or articulated object instance handle for the parent object to which the Receptacle is attached. None for globally defined stage Receptacles.
+        :param parent_link: Index of the link to which the Receptacle is attached if the parent is an ArticulatedObject. -1 denotes the base link. None for rigid objects and stage Receptables.
+        :param up: The "up" direction of the Receptacle in local AABB space. Used for optionally culling receptacles in un-supportive states such as inverted surfaces.
+        """
+        super().__init__(name, parent_object_handle, parent_link, up)
+        self.mesh_data = mesh_data
+        self.area_weighted_accumulator = (
+            []
+        )  # normalized float weights for each triangle for sampling
+        assert len(mesh_data[1]) % 3 == 0, "must be triangles"
```
vincentpierre, line 325:
```suggestion
        assert len(mesh_data[1]) % 3 == 0, "Provided mesh is not a Triangle. Triangle Mesh receptacles can only be triangles"
```

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -241,13 +274,169 @@ def add_receptacle_visualization(
...
+        """
+        Get all three vertices of a mesh triangle given it's face index as a list of numpy arrays.
+
+        :param f_ix: The index of the mesh triangle.
+        """
+        verts: List[np.ndarray] = []
+        for ix in range(3):
+            verts.append(
+                np.array(
+                    self.mesh_data[0][self.mesh_data[1][int(f_ix * 3 + ix)]]
```
vincentpierre, line 356: I don't know what is happening here. Can you create an intermediate variable for `int(f_ix * 3 + ix)` and maybe `self.mesh_data[1]`

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -241,13 +274,169 @@ def add_receptacle_visualization(
...
+        """
+
+        def find_ge(a: List[Any], x) -> Any:
+            "Find leftmost item greater than or equal to x"
+            from bisect import bisect_left
+
+            i = bisect_left(a, x)
+            if i != len(a):
+                return i
+            raise ValueError
```
vincentpierre, line 375: Can you add a message for this error ?

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -241,13 +274,169 @@ def add_receptacle_visualization(
...
+
+        :param sim: Simulator must be provided.
+        :param color: Optionally provide wireframe color, otherwise magenta.
+        """
+        # draw all mesh triangles
+        if color is None:
+            color = mn.Color4.magenta()
+        dblr = sim.get_debug_line_render()
+        dblr.push_transform(self.get_global_transform(sim))
+        assert len(self.mesh_data[1]) % 3 == 0, "must be triangles"
```
vincentpierre, line 420: I see this twice already. Make a helper assert method ?

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -262,6 +451,9 @@ def get_all_scenedataset_receptacles(sim) -> Dict[str, Dict[str, List[str]]]:
         stage_template = stm.get_template_by_handle(template_handle)
         for item in stage_template.get_user_config().get_subconfig_keys():
             if item.startswith("receptacle_"):
+                logger.info(
+                    f"template file_directory = {stage_template.file_directory}"
```
vincentpierre, line 455: This and line 467 don't seem very helpful without a more detailed log.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -290,9 +485,39 @@ def get_all_scenedataset_receptacles(sim) -> Dict[str, Dict[str, List[str]]]:
...
+
+    :param ply_file: The input PLY mesh file. NOTE: must contain only triangles.
+    """
+    manager = mn.trade.ImporterManager()
+    importer = manager.load_and_instantiate("AnySceneImporter")
+    importer.open_file(ply_file)
+
+    # NOTE: We don't support mesh merging or multi-mesh parsing currently
+    if importer.mesh_count > 1:
+        raise NotImplementedError("TODO: multi-mesh receptacle support.")
```
vincentpierre, line 500:
```suggestion
    # TODO: We don't support mesh merging or multi-mesh parsing currently
    if importer.mesh_count > 1:
        raise NotImplementedError("Importing multi-mesh receptacles (mesh merging or multi-mesh parsing) is not supported. ")
```

habitat-lab/habitat/utils/geometry_utils.py
```diff
@@ -115,3 +116,53 @@ def agent_state_target2ref(
...
+    coef1 = random.random()
+    coef2 = random.random()
+    if coef1 + coef2 >= 1:
+        # transform "outside" points back inside
+        coef1 = 1 - coef1
+        coef2 = 1 - coef2
+    return v0 + coef1 * (v1 - v0) + coef2 * (v2 - v0)
+
+
+def point_in_triangle_test(
```
vincentpierre, line 138: is test the best name for this method. It sounds too much like a unit_test to me...

vincentpierre: is_point_in_triangle ?

vincentpierre (commented):
Looks great. Left some comments, but I do not have much to say about the overall change.

habitat-lab/habitat/datasets/rearrange/run_episode_generator.py
```diff
@@ -50,6 +50,8 @@ class RearrangeEpisodeGeneratorConfig:
     additional_object_paths: List[str] = field(
         default_factory=lambda: ["data/objects/ycb/"]
     )
+    # optionally correct unstable states by removing extra unstable objects (within minimum samples limitations)
+    correct_unstable_results: bool = False
```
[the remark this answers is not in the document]
aclegg3: Optional feature because:
1.  backwards compatibility
2. user can choose if culling is allowed for their purposes (default True would be ok for this)
3. Doesn't work with target sampling yet

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
mosra: I could expose [MeshTools::concatenate()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a3f38d4e79ecbb0d41f4d58a1f5a59c48) and [SceneTools::flattenMeshHierarchy()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1SceneTools.html#a8b90ec6a17a8f3f3d1150dfdcedab40c) for doing this if you want.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
mosra: I'm thinking that this should ideally be something to handle directly in the Magnum asset pipeline, so you can always be sure that a `MeshPrimitive.TRIANGLES` mesh has the index / vertex count divisible by 3.

Currently that is done only partially (it's always true for PLY, STL or OBJ but for example not for glTF) and it's not enforced by MeshData itself, but I could clean that up. Seems like desirable behavior to me.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
mosra: It's indexing a list of positions. *That* operation is clear to me, what however makes it confusing is that `self.mesh_data[0]` is a position list and `self.mesh_data[1]` is an index list. 

A named tuple instead? `self.mesh_positions` and `self.mesh_indices` instead?

habitat-lab/habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -391,29 +393,15 @@ def generate_scene(self) -> str:
...
-                self.vdb.look_at(
-                    viz_objects[0].root_scene_node.absolute_translation
-                )
-                self.vdb.get_observation()
-            else:
-                logger.warning(
-                    f"visualize_scene_receptacles: no visualization object generated for Receptacle '{receptacle.name}'."
-                )
+            receptacle.debug_draw(self.sim)
+            self.vdb.look_at(receptacle.sample_uniform_global(self.sim, 1.0))
```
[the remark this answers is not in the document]
aclegg3: Receptacle does not have a "position" per-se, rather they are defined w.r.t. to the object's local frame (or the global frame). AABBRecetpacles could use the AABB center, but not TriangleMeshReceptacles.

I suppose I could add a cached `centroid` variable for any Receptacle and use that, but even then it could be misleading (e.g. if a mesh Receptacle had multiple connected components then the centroid may be off the mesh entirely).

habitat-lab/habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -852,5 +866,74 @@ def settle_sim(
                 prefix="settle_", fps=30, obs_cache=settle_db_obs
             )
 
+        # detailed receptacle stability report
+        logger.info("  Detailed sampling stats:")
```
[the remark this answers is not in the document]
aclegg3: Sure, makes sense. I guess the downside is that if the code crashes we never get a log at all to know when that happened unless I add try|catch logic which can obfuscate issues.

habitat-lab/habitat/datasets/rearrange/rearrange_generator.py
```diff
@@ -852,5 +866,74 @@ def settle_sim(
...
+
+        # compute number of unstable objects for each receptacle
+        # receptacle: [num_objects, num_unstable_objects]
```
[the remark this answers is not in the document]
aclegg3: Cool, good suggestion.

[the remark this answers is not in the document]
aclegg3: Actually, this needs to be mutable. I'll use a Dict instead of the List for better comprehension.

habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -415,8 +418,18 @@ def sample(
                     self.receptacle_candidates = None
 
             if new_object is not None:
+                # when an object placement is successful, reset the try counter.
+                logger.info(
+                    f"    found obj|receptacle pairing ({len(new_objects)}/{self.target_objects_number}) in {num_pairing_tries} attempts ({time.time()-pairing_start_time}sec)."
+                )
+                num_pairing_tries = 0
+                pairing_start_time = time.time()
```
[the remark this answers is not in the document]
aclegg3: Actually, this timer is tracking overall time to find a match, not time for each attempt, so it needs to be reset only when a placement is successful. I think the current logic is correct.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -66,11 +70,24 @@ def sample_uniform_local(
...
 
-    @abstractmethod
     def get_global_transform(self, sim: habitat_sim.Simulator) -> mn.Matrix4:
         """
         Isolates boilerplate necessary to extract receptacle global transform of the Receptacle at the current state.
         """
+        if self.parent_object_handle is None:
+            # global identify by default
+            return mn.Matrix4.identity_init()
+        elif not self.is_parent_object_articulated:
```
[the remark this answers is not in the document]
aclegg3: Here we use `parent_object_handle=None` to indicate that a Receptacle is not parented, but instead defined in the global frame, so this logic is correct.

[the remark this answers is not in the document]
aclegg3: I see, you mean we remove the elif and else because they return.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -91,6 +108,18 @@ def add_receptacle_visualization(
         """
         return []
 
+    @abstractmethod
+    def debug_draw(
+        self, sim: habitat_sim.Simulator, color: Optional[mn.Color4] = None
+    ) -> None:
+        """
+        Render the Receptacle with DebugLineRender utility at the current frame.
```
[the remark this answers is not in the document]
aclegg3: This is the abstract definition of the function is used by derived Receptacle types to specify wireframe debug line renders. AABB draws a box and TriangleMesh draws the mesh wireframe. The API is specific and uniform so higher level functions don't need to know which derived type was used. This function can be called like `receptacle.debug_draw(sim, color)` regardless.

 >add a NotImplementedError

Thanks :+1:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: Changed to a named tuple. :+1:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: I think folks are hoping to eventually have multiple separate meshes in a single file which we'll split out here.
Generally happy to see more Magnum mesh tools in python though. :slightly_smiling_face:

aclegg3:
> Looks great. Left some comments, but I do not have much to say about the overall change.

Thanks, I've addressed your comments and resolved those not containing additional discussion.

habitat-lab/habitat/datasets/rearrange/run_episode_generator.py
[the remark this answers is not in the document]
vincentpierre: Can you add a comment explaining just that?

vincentpierre (approved):
LGTM, I would prefer another approval from @mosra if possible.

habitat-lab/habitat/datasets/rearrange/configs/all_receptacles_test.yaml
```diff
@@ -0,0 +1,54 @@
+---
```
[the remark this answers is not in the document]
aclegg3: This is for users to test generation on their scenes with all receptacles defined without needing to define another custom generator config. I can remove it from this PR if that isn't a good enough reason to include it.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: Update: I think we want to store separate receptacle meshes as glTF nodes. So here we'll load the glTF file and iterate over all meshes in the scene, creating a TriangleMeshReceptacle for each one.

mosra (approved):

PR #1117, opened 2023-02-03
Speed Upgrade to Rearrangement Tasks

ASzot:
## Motivation and Context

Small fix with big performance improvements. This sets objects outside a set radius of the agent to `MotionType.STATIC`. In manual tests with the interactive play script, this difference is unnoticable. The PR is not ready to merge until I confirm this has no impact on learning.

Sim speed test:
- Without this PR: 22.09266252303496 seconds
- With this PR: 11.722101256018505 seconds *(almost 2x improvement)*

Training speed test (20 environments, VER trainer, 1GPU, 16CPUs, 10 updates):
- Without this PR: 429 SPS
- With this PR: 633 SPS *(42% faster)*.

## How Has This Been Tested

- Manual interactive play test.

### DIFF
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -1071,6 +1071,7 @@ class SimulatorConfig(HabitatBaseConfig):
     create_renderer: bool = False
     requires_textures: bool = True
     auto_sleep: bool = False
+    sleep_dist: float = 3.0
     step_physics: bool = True
     concur_render: bool = False
     # If markers should be updated at every step:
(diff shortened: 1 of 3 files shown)


aclegg3:
This is actually a nice optimization specifically for home-scale rearrange with one or a small set of agents and local-dynamics. :slightly_smiling_face: 

I'll suggest a couple things to think about:
1. I think you could also apply this to articulated objects for additional gains.
2. do the auto-sleep at a lower frequency than every step (every 10?). This could be measured based on the maximum speed of the robot (if it can't move more than 1 meter in 1 second, then no need to do this more often than once per second, right?) Context: Changing motion type to the same motion type costs nothing, but changing back and forth between STATIC and DYNAMIC will incur a cost as the collision shapes are swapped out and broad-phase is re-computed.
3. You may consider a slightly more accurate distance calculation to avoid possible edge cases (e.g. pad the object COM with approximate object size from the bounding box). This could then apply to objects of various sizes and shapes.
4. We can't discount the possibility of physics artifacts from a heuristic like this. Maybe there is a nice way to check for stability at scale. Something like tracking a debug  setup where the agent moves through a scene and we track the velocity of objects for the first quarter second after setting them back to DYNAMIC.

PR #1129, opened 2023-02-09
add continuous controller

jimmytyyang:
## Motivation and Context

 For habitat sim2real challenge 2023

## How Has This Been Tested

 running interactive_play.py and will add testing script

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -162,6 +162,7 @@ class VelocityControlActionConfig(ActionConfig):
     min_abs_lin_speed: float = 0.025  # meters/sec
     min_abs_ang_speed: float = 1.0  # # deg/sec
     time_step: float = 1.0  # seconds
+    enable_scale_convert: bool = True
 
 
 # -----------------------------------------------------------------------------
@@ -1299,6 +1300,12 @@ cs.store(
     group="habitat/task/actions",
     name="arm_action",
     node=ArmActionConfig,
+)
+cs.store(
+    package="habitat.task.actions.velocity_control",
+    group="habitat/task/actions",
+    name="velocity_control",
+    node=VelocityControlActionConfig,
 )
 cs.store(
     package="habitat.task.actions.base_velocity",
--- a/habitat-lab/habitat/tasks/nav/nav.py
+++ b/habitat-lab/habitat/tasks/nav/nav.py
@@ -40,6 +40,7 @@ from habitat.utils.geometry_utils import (
     quaternion_rotate_vector,
 )
 from habitat.utils.visualizations import fog_of_war, maps
+from habitat.utils.common import continous_controller
 
 try:
     from habitat.sims.habitat_simulator.habitat_simulator import HabitatSim
@@ -1170,6 +1171,7 @@ class VelocityAction(SimulatorTaskAction):
         self.min_abs_ang_speed = config.min_abs_ang_speed
         self.time_step = config.time_step
         self._allow_sliding = self._sim.config.sim_cfg.allow_sliding  # type: ignore
+        self._enable_scale_convert = config.enable_scale_convert
 
     @property
     def action_space(self):
@@ -1195,8 +1197,8 @@ class VelocityAction(SimulatorTaskAction):
         self,
         *args: Any,
         task: EmbodiedTask,
-        linear_velocity: float,
-        angular_velocity: float,
+        linear_velocity: float = None,
+        angular_velocity: float = None,
         time_step: Optional[float] = None,
         allow_sliding: Optional[bool] = None,
         **kwargs: Any,
@@ -1217,17 +1219,63 @@ class VelocityAction(SimulatorTaskAction):
         if time_step is None:
             time_step = self.time_step
 
-        # Convert from [-1, 1] to [0, 1] range
-        linear_velocity = (linear_velocity + 1.0) / 2.0
-        angular_velocity = (angular_velocity + 1.0) / 2.0
+        if "base_vel" in kwargs:  # Velocity control mode
+            linear_velocity = max(0.0, kwargs["base_vel"][0])
+            angular_velocity = kwargs["base_vel"][1]
+        elif "base_dis" in kwargs: # Discrete control mode
+            move = kwargs["base_dis"][0]
+            turn = kwargs["base_dis"][1]
+            if move > 0:
+                linear_velocity = 0.15 * (1/self.time_step)
+            else:
+                linear_velocity = 0
+            if turn > 0:
+                angular_velocity = np.pi * 30.0/180.0 * (1/self.time_step)
+            elif turn < 0:
+                angular_velocity = -np.pi * 30.0/180.0 * (1/self.time_step)
+            else:
+                angular_velocity = 0
+        elif "base_pt" in kwargs: # Waypoint control mode
+            # init the controllers
+            w2v_controller = continous_controller()
+            # Get the transformation of the agent
+            trans = self._sim.agents[0].scene_node.transformation
+            # Define the global position
+            target_x = max(0.0, kwargs["base_pt"][0]*1.0)
+            global_pos = trans.transform_point(np.array([target_x, 0.0, 0.0]))
+            # Set the target rotation angle
+            target_theta = kwargs["base_pt"][1]
+            target_theta = float(self._sim.agents[0].state.rotation.angle()) + target_theta
+            # Set the goal
+            xyt_goal = [
+                global_pos[0],
+                global_pos[2],
+                target_theta,
+            ]
+            w2v_controller.set_goal(xyt_goal)
+            # Get the current location of the agent
+            xyt = [
+                self._sim.agents[0].state.position[0],
+                self._sim.agents[0].state.position[2],
+                float(self._sim.agents[0].state.rotation.angle()),
+            ]
+            # Get the velocity action
+            vel_action = w2v_controller.forward(xyt, trans)
+            linear_velocity = vel_action[0]
+            angular_velocity =  vel_action[1]
 
-        # Scale actions
-        linear_velocity = self.min_lin_vel + linear_velocity * (
-            self.max_lin_vel - self.min_lin_vel
-        )
-        angular_velocity = self.min_ang_vel + angular_velocity * (
-            self.max_ang_vel - self.min_ang_vel
-        )
+        if self._enable_scale_convert:
+            # Convert from [-1, 1] to [0, 1] range
+            linear_velocity = (linear_velocity + 1.0) / 2.0
+            angular_velocity = (angular_velocity + 1.0) / 2.0
+
+            # Scale actions
+            linear_velocity = self.min_lin_vel + linear_velocity * (
+                self.max_lin_vel - self.min_lin_vel
+            )
+            angular_velocity = self.min_ang_vel + angular_velocity * (
+                self.max_ang_vel - self.min_ang_vel
+            )
 
         # Stop is called if both linear/angular speed are below their threshold
         if (
--- a/habitat-lab/habitat/utils/common.py
+++ b/habitat-lab/habitat/utils/common.py
@@ -52,3 +52,159 @@ def cull_string_list_by_substrings(
                     culled_list.append(string)
                     break
     return culled_list
+
+
+
+
+
+def transform_global_to_base(XYT, current_pose, trans=None):
+    """
+    Transforms the point cloud into geocentric frame to account for
+    camera position
+    """
+    goal_pos = trans.inverted().transform_point(
+        np.array([XYT[0], 0.0, XYT[1]])
+    )
+    error_t = XYT[2] - current_pose[2]
+    error_t = (error_t + np.pi) % (2.0*np.pi) - np.pi
+    error_x = goal_pos[0]
+    error_y = goal_pos[2]
+    return [error_x, error_y, error_t]
+
+
+
+
+class continous_controller:
+    """A controller that transforms waypoints to velocity command"""
+    def __init__(self, v_max=1.0, w_max=1.0, acc_lin=2.4, acc_ang=2.4, \
+        max_heading_ang=np.pi / 10, \
+        lin_error_tol=0.001, ang_error_tol=0.001,\
+        track_yaw=True):
+        self.track_yaw = track_yaw
+
+        # Params
+        self.v_max = v_max
+        self.w_max = w_max
+
+        self.acc_lin = acc_lin
+        self.acc_ang = acc_ang
+
+        self.max_heading_ang = max_heading_ang
+
+        self.lin_error_tol = lin_error_tol
+        self.ang_error_tol = ang_error_tol
+
+        # Init
+        self.xyt_goal = np.zeros(3)
+        self.dxyt_goal = np.zeros(3)
+
+    def set_goal(self, goal, vel_goal=None):
+        self.xyt_goal = goal
+        if vel_goal is not None:
+            self.dxyt_goal = vel_goal
+
+    def _compute_error_pose(self, xyt_base, trans=None):
+        """
+        Updates error based on robot localization
+        """
+        xyt_err = transform_global_to_base(self.xyt_goal, xyt_base, trans)
+        if not self.track_yaw:
+            xyt_err[2] = 0.0
+
+        return xyt_err
+
+    @staticmethod
+    def _velocity_feedback_control(x_err, a, v_max):
+        """
+        Computes velocity based on distance from target.
+        Used for both linear and angular motion.
+
+        Current implementation: Trapezoidal velocity profile
+        """
+        t = np.sqrt(2.0 * abs(x_err) / a)
+        v = min(a * t, v_max)
+        return v * np.sign(x_err)
+
+    @staticmethod
+    def _turn_rate_limit(lin_err, heading_diff, w_max, max_heading_ang, tol=0.0):
+        """
+        Compute velocity limit that prevents path from overshooting goal
+
+        heading error decrease rate > linear error decrease rate
+        (w - v * np.sin(phi) / D) / phi > v * np.cos(phi) / D
+        v < (w / phi) / (np.sin(phi) / D / phi + np.cos(phi) / D)
+        v < w * D / (np.sin(phi) + phi * np.cos(phi))
+
+        (D = linear error, phi = angular error)
+        """
+        assert lin_err >= 0.0
+        assert heading_diff >= 0.0
+        if heading_diff > max_heading_ang:
+            return 0.0
+        else:
+            return (
+                w_max
+                * lin_err
+                / (
+                    np.sin(heading_diff)
+                    + heading_diff * np.cos(heading_diff)
+                    + 1e-5
+                )
+            )
+
+    def _feedback_traj_track(self, xyt_err):
+        xyt_err = self._compute_error_pose(xyt)
+        v_raw = (
+            V_MAX_DEFAULT
+            * (K1 * xyt_err[0] + xyt_err[1] * np.tan(xyt_err[2]))
+            / np.cos(xyt_err[2])
+        )
+        w_raw = (
+            V_MAX_DEFAULT
+            * (K2 * xyt_err[1] + K3 * np.tan(xyt_err[2]))
+            / np.cos(xyt_err[2]) ** 2
+        )
+        v_out = min(v_raw, V_MAX_DEFAULT)
+        w_out = min(w_raw, W_MAX_DEFAULT)
+        return np.array([v_out, w_out])
+
+    def _feedback_simple(self, xyt_err):
+        v_cmd = w_cmd = 0
+
+        lin_err_abs = np.linalg.norm(xyt_err[0:2])
+        ang_err = xyt_err[2]
+
+        # Go to goal XY position if not there yet
+        if lin_err_abs > self.lin_error_tol:
+            heading_err = np.arctan2(xyt_err[1], xyt_err[0])
+            heading_err_abs = abs(heading_err)
+            # Compute linear velocity
+            v_raw = self._velocity_feedback_control(
+                lin_err_abs, self.acc_lin, self.v_max
+            )
+            v_limit = self._turn_rate_limit(
+                lin_err_abs,
+                heading_err_abs,
+                self.w_max / 2.0,
+                self.max_heading_ang,
+                tol=self.lin_error_tol,
+            )
+
+            v_cmd = np.clip(v_raw, 0.0, v_limit)
+
+            # Compute angular velocity
+            w_cmd = self._velocity_feedback_control(
+                heading_err, self.acc_ang, self.w_max
+            )
+
+        # Rotate to correct yaw if yaw tracking is on and XY position is at goal
+        elif abs(ang_err) > self.ang_error_tol and self.track_yaw:
+            # Compute angular velocity
+            w_cmd = self._velocity_feedback_control(
+                ang_err, self.acc_ang, self.w_max
+            )
+        return v_cmd, w_cmd
+
+    def forward(self, xyt, trans):
+        xyt_err = self._compute_error_pose(xyt, trans)
+        return self._feedback_simple(xyt_err)


jimmytyyang:
Hi Austin, 

Currently, I was able to get it work as expected by doing the following things
In nav.py
(1) In `def _agent_state_to_xyt(agent_state):` I just use `float(agent_state.rotation.angle())`
(2) when doing `for _t in np.arange(0.0, duration / dt) * dt`, I only enter the loop once
`if _t == 0: break`
(3) In `MoveForwardWaypointAction`, please add `xyt_waypoint = self._preprocess_action(xyt_waypoint)` after the line `xyt_waypoint = [self._config.forward_step_size, 0.0, 0.0]`

In go_controller.py
(1) In `xyt_global_to_base`, I swap the axis: 
`xyt_base2target[1] = x_diff * base_cos + y_diff * base_sin
 xyt_base2target[0] = x_diff * -base_sin + y_diff * base_cos
`  
(2) In `xyt_base_to_global`, I did
`  x_base2target = xyt_base2target[0] * base_cos - xyt_base2target[1] * base_sin
    y_base2target = xyt_base2target[0] * base_sin + xyt_base2target[1] * base_cos
`

In velocity_controllers.py
(1) In `def __call__(self, xyt_err):`, I do `if lin_err_abs > self.cfg.lin_error_tol:` and `elif abs(ang_err) > self.cfg.ang_error_tol:`

The first two is more like transformation issue you mentioned. We can discuss if these solutions make sense or now and how to better document these. The last one is just a bug for variable name.

Finally, I noticed that the velocity controller will output non-zero value for lin vel when only using velocity control with zero lin vel input. Should we disable scaling?

Thank you! let's us discuss tmr, and it was nice to work with you!

jimmytyyang:
videos 

https://user-images.githubusercontent.com/55121504/220160630-73e3c53d-d0fa-4496-8d33-580ee8aa518e.mp4

exhaustin:
Thank you for debugging this!
I can apply the typo fixes in velocity_controllers.py and push to the tag.

Regarding the other changes however, can you explain the logic behind making these changes?
In any case we should go through them carefully together, as I am worried that if we don't fully understand the problem it will come back and bite us at some point.
I sent you a calendar invite for tomorrow morning.

exhaustin:
Here are my comments on the changes you mentioned, which hopefully will be good starting points for our discussion:

> (1) In `def _agent_state_to_xyt(agent_state):` I just use `float(agent_state.rotation.angle())` 

Have you printed out and compared values for both implementations? I’m suspecting that `float(agent_state.rotation.angle())` will be incorrect 50% of the time when the axis points in the other direction.

>(2) when doing `for _t in np.arange(0.0, duration / dt) * dt`, I only enter the loop once `if _t == 0: break` 

So you are only executing 1 timestep? Did you increase the maximum velocity or timestep somehow to compensate? Doing this seems to take away the entire notion of feedback control.

>(3) In `MoveForwardWaypointAction`, please add `xyt_waypoint = self._preprocess_action(xyt_waypoint)` after the line `xyt_waypoint = [self._config.forward_step_size, 0.0, 0.0]`

This doesn’t seem right. If you do this, xyt_waypoint will be passed into `self._preprocess_action` two times (as `_preprocess_action` is also being called on the input in the parent class `WaypointAction`). Clipping two times shouldn’t have an effect, but scaling two times sounds problematic. Perhaps by doing this you scaled up the numbers twice and went through the clipping, thus achieving maximum values in a non-deliberate manner?

> In go_controller.py (1) In `xyt_global_to_base`, I swap the axis: `xyt_base2target[1] = x_diff * base_cos + y_diff * base_sin xyt_base2target[0] = x_diff * -base_sin + y_diff * base_cos ` (2) In `xyt_base_to_global`, I did ` x_base2target = xyt_base2target[0] * base_cos - xyt_base2target[1] * base_sin y_base2target = xyt_base2target[0] * base_sin + xyt_base2target[1] * base_cos`

So in addition to the coordinate transform, you also swapped the x and y axes. Why?

> Finally, I noticed that the velocity controller will output non-zero value for lin vel when only using velocity control with zero lin vel input. Should we disable scaling?

I think scaling should be disabled by default so the action spaces are more transparent. Users should be able to turn it on if their models predict actions within a normalized space. Maybe a lot of the problems we're seeing comes from the fact that it's actually enabled by default which I did not realize.

Finally a piece of lightweight feedback: If you have experimental changes that you would like to put up for review, it is completely okay to just push commits to the PRs as commits are revert-able and update-able. If the changes are hard to isolate, you can stack another draft PR on top. Descriptions of code changes are not 100% transparent, and requires extra effort from both sides to iterate.

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -186,6 +184,9 @@ class WaypointControlActionConfig(VelocityControlActionConfig):
     max_heading_ang: float = 0.7854  # 45 degrees in rad
     lin_error_tol: float = 0.01
     ang_error_tol: float = 0.025
+    # Stop criteria
+    min_abs_lin_diff: float = 0.01  # meters
```
ykarmesh, line 188: I think we should keep the numbers consistent across `VelocityControlActionConfig` and `WaypointControlActionConfig` , etc. The thresholds for stopping shouldn't change even if you are using a different controller.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1218,6 +1224,12 @@ def step(
 
         return self._get_agent_observation(agent_state_result)
 
+    def _check_stop(self, linear_velocity, angular_velocity) -> bool:
+        """Use a heuristic to determine if an action qualifies as a stop action"""
+        lin_check = abs(linear_velocity) < self._config.min_abs_lin_speed
```
ykarmesh, line 1229: Avoid using hydra configs within a loop. This will cause considerable slow down as noted by Vince in a prior PR.

habitat-lab/habitat/config/default_structured_configs.py
exhaustin: The reason we cannot have a consistent criterion within the current design is that WaypointControl and VelocityControl are not exactly 1-1 mappings.
VelocityControl executes a base velocity command for 0.1 secs.
WaypointControl runs a controller that generates velocity commands and tries to navigate to the goal for 1 seconds. So that's 10 velocity commands, whether the goal is reached or not. We might get zero velocities but that just means we reached the waypoint prematurely, it does not mean that the intent of the policy is to stay in place (so instead, a stop action is when the waypoint is close to the current location)

Of course, I don't know if this is the best design, so happy to iterate on that

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: Wow that's new to me. Interesting design choice from Hydra to have configs resolved dynamically.
Will fix.

exhaustin: @ykarmesh Can you provide a link to the PR you mentioned? Would love to read more on this matter, because it would affect certain designs in my other projects too.

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -155,13 +155,74 @@ class TeleportActionConfig(ActionConfig):
 @attr.s(auto_attribs=True, slots=True)
 class VelocityControlActionConfig(ActionConfig):
     type: str = "VelocityAction"
-    # meters/sec
-    lin_vel_range: List[float] = [0.0, 0.25]
-    # deg/sec
-    ang_vel_range: List[float] = [-10.0, 10.0]
+    lin_vel_range: List[float] = [0.0, 0.25]  # meters/sec
+    ang_vel_range: List[float] = [-10.0, 10.0]  # deg/sec
```
ykarmesh, line 159: If we are using radian inputs, lets also use radians here to prevent any possible errors

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -155,13 +155,74 @@ class TeleportActionConfig(ActionConfig):
...
+    In Navigation tasks only, this action will utilize a velocity controller to
+    move the robot to the desired waypoint.
+    The waypoint is specified in the frame of the robot.
+    """
+    type: str = "WaypointAction"
+    action_duration: float = 1.0  # seconds
+    # Action space range
+    waypoint_lin_range: List[float] = [-0.5, 0.5]  # meters
+    waypoint_ang_range: List[float] = [-180, 180]  # degrees
+    max_max_duration: float = 10.0  # seconds
```
ykarmesh, line 180: Consider changing the name of this variable to be more informative?

exhaustin: Any suggestions?

habitat-lab/habitat/tasks/nav/nav.py
ykarmesh: https://github.com/facebookresearch/habitat-lab/pull/1120

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -198,13 +199,74 @@ class TeleportActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, this action will utilize a velocity controller to
+    move the robot to the desired waypoint.
+    The waypoint is specified in the frame of the robot.
+    """
+    type: str = "WaypointAction"
+    action_duration: float = 1.0  # seconds
+    # Action space range
+    waypoint_lin_range: List[float] = [-0.5, 0.5]  # meters
+    waypoint_ang_range: List[float] = [-np.pi, 180]  # radians
```
ykarmesh, line 223:
```suggestion
    waypoint_ang_range: List[float] = [-np.pi, np.pi]  # radians
```

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1341,310 @@ def step(
...
+                xyt
+            )
+
+            # Apply action and step simulation
+            next_agent_state = self._apply_velocity_action(
+                linear_velocity, angular_velocity, time_step=dt
+            )
+            xyt = self._agent_state_to_xyt(next_agent_state)
+
+            # Complete action early if commanded speed is low
```
ykarmesh, line 1554: This condition is essentially checking if the agent has been commanded to stop by the controller. So it should be above the `self._apply_velocity_action()` function.

exhaustin: It doesn't really matter if we execute the last action or not if the velocity is low.
Having it here will guarantee that we take at least one step and `next_action_state` is populated.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1554,7 +1565,10 @@ def _step_rel_waypoint(
                 linear_velocity, angular_velocity, time_step=self._time_step
             )
             xyt = self._agent_state_to_xyt(next_agent_state)
-
+            # Update the xyt of the controller
+            self.w2v_controller.set_goal(
+                xyt_waypoint, start=xyt, relative=True
+            )
```
exhaustin, line 1571: This doesn't make sense. If you keep setting the goal relative to the current location, then the robot will continuously chase a moving target instead of stopping at 0.25 (in the case of move_forward).
What happens if you don't add this?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1344,6 +1348,13 @@ def _apply_velocity_action(
         # TODO: Make a better way to flag collisions
         self._sim._prev_sim_obs["collided"] = collided  # type: ignore
 
+        # Update the state of the agent
+        agent_observations = self._sim.get_observations_at(
+            position=final_position,
+            rotation=final_rotation,
+            keep_agent_at_new_pose=True,
+        )
```
exhaustin, line 1356: Is the idea here that if we don't call `self._sim.get_observations_at(..., keep_agent_at_new_pose=True)`, the agent doesn't actually move?
If so, we should deduplicate this with `get_agent_observation`.

Also one minor concern, does `self._sim.get_observations_at` incur additional rendering cost? Is that something we should be worried about? @ykarmesh

jimmytyyang: Right, as my comment suggested, the agent does not move if you do not call self._sim.get_observations_at() to update the position. Yeah, we can reuse the function.

habitat-lab/habitat/tasks/nav/nav.py
jimmytyyang: If I do not add this, the robot just move spirally, meaning that it outputs linear and angular velocity at the same time

jimmytyyang: Is this an expected thing the robot moves spirally?

jimmytyyang: Another point is that do we really update the robot local transformation matrix? meaning that if the robot moves at `_t=0`, then the transformation at `_t=0.1` cannot use the previous transformation?

exhaustin: The spiral movement is not expected.
The reason that it generates angular velocity is that the state feedback (next_agent_state) isn't interpreted to be moving forwards. So the controller tries to rotate and compensate.
You circumvented the problem by telling the robot to always move forward, but you lose feedback information and thus the controller is now just blindly moving forwards instead of trying to stabilize to the waypoint. What we should be doing is to find out and correct why the state feedback is wrong or being interpreted wrong.

exhaustin: > Another point is that do we really update the robot local transformation matrix? meaning that if the robot moves at `_t=0`, then the transformation at `_t=0.1` cannot use the previous transformation?

What do you mean by robot local transformation matrix?

jimmytyyang: `xyt_err = xyt_global_to_base(self.xyt_goal, self.xyt_loc)` It seems xyt_err does not output correct values within _t for-loop: 
_t: 0.0 [ 0.50000078 -0.03000045  0.        ]
_t: 0.1 [ 0.50143064 -0.0465724   0.02681698]
_t: 0.2 [ 0.50370458 -0.0598166   0.06015406]
Maybe `xyt_global_to_base()` is doing something strange?

jimmytyyang: wait, but you do update `update_pose_feedback()` each time, which I think it is correct

exhaustin: > 

Yes exactly

exhaustin: > 

Are you printing out values of `xyt_err`?
If so at _t: 0.0 the error should be [0.25, 0, 0].

jimmytyyang: Yes, I am printing out `xyt_err`, and it should be [0.5, 0, 0]? why  [0.25, 0, 0]?

exhaustin: What command are you giving the robot? Are you not calling `move_forward`?

exhaustin: This is what I got from running move_forward on the latest branch:
```
======================
action={'action': 'move_forward_waypoint', 'action_args': None}
===========[controller integration]===========
hab_goal=[0.25 0.   0.  ], agent_goal=[ 14.89055808 -17.44277298   4.03323325]
t=0.0
        hab_xyt=[ 15.04759216 -17.24824715   4.03323325], agent_xyt=[ 15.04759216 -17.24824715   4.03323325]
        agent_xyt_err=[ 2.50000000e-01 -1.16573418e-15  0.00000000e+00]
        vel_cmd=[0.3, -7.480323551899881e-08]
t=0.1
        hab_xyt=[ 15.04759216 -17.24824715   4.03323325], agent_xyt=[ 15.04759216 -17.24824715   4.03323325]
        agent_xyt_err=[ 2.50000000e-01 -1.16573418e-15  0.00000000e+00]
        vel_cmd=[0.3, -7.480323551899881e-08]
t=0.2
        hab_xyt=[ 15.04759216 -17.24824715   4.03323325], agent_xyt=[ 15.04759216 -17.24824715   4.03323325]
        agent_xyt_err=[ 2.50000000e-01 -1.16573418e-15  0.00000000e+00]
        vel_cmd=[0.3, -7.480323551899881e-08]
```
As you can see xyt_err is always [0.25, 0, 0] since you are setting the goal every timestep.

exhaustin: Also the robot is rotating like 90 degrees every time we call `turn_left` or `turn_right`

exhaustin:
Unresolved concerns:
- Does calling `Simulation.get_observation` at every velocity integration timestep incur additional rendering costs?
- `enable_scale_convert` basically denormalizes input velocity control params from an input range of [-1, 1]. Is it a common application? If not it adds a lot of logic and should probably be removed if it's not going to be used.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1195,10 +1243,9 @@ def step(
         self,
         *args: Any,
         task: EmbodiedTask,
-        linear_velocity: float,
-        angular_velocity: float,
+        linear_velocity: float = None,
```
ykarmesh, line 1246: I dont think we will want to have a case where None is passed to the agent. Lets remove it

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -198,13 +207,70 @@ class TeleportActionConfig(ActionConfig):
...
+    In Navigation tasks only, this action will utilize a velocity controller to
+    move the robot to the desired waypoint.
+    The waypoint is specified in the frame of the robot.
+    """
+    type: str = "WaypointAction"
+    action_duration: float = 1.0  # seconds
+    # Action space range
+    waypoint_lin_range: List[float] = [-0.5, 0.5]  # meters
+    waypoint_ang_range: List[float] = [-np.pi, np.pi]  # radians
+    wait_duration_range: List[float] = [0.0, 10.0]  # seconds
```
ykarmesh, line 228: Should the wait_duration_range be [time_step, 10.] so that the controller atleast takes one step in the env?

exhaustin: Good call, although having dependencies between configuration fields seems to be an unnecessary complexity. I'll modify the logic of the loop in `WaypointAction` to always execute for 1 step.

examples/tutorials/velocity_control_sim.ipynb
```diff
@@ -0,0 +1,274 @@
+{
```
vincentpierre, line 1: What does this do? Can you add a markdown cell that describes what this is and what it does ?

habitat-baselines/habitat_baselines/config/instance_imagenav/ddppo_instance_imagenav_v3.yaml
```diff
@@ -0,0 +1,88 @@
+# @package _global_
```
vincentpierre, line 1: For my personal understanding, what is different compared to v2 besides the benchmark ?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -134,6 +137,14 @@ class StopActionConfig(ActionConfig):
     type: str = "StopAction"
 
 
+@attr.s(auto_attribs=True, slots=True)
+class VelocityStopActionConfig(ActionConfig):
```
vincentpierre, line 141: This is the third stop action defined here. Can this be RearrangeStopAction ? If not, why?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -134,6 +137,14 @@ class StopActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, the velocity stop action is an action that takes continuous space input.
```
vincentpierre, line 143: what does `an action that takes continuous space input` mean ?

habitat-lab/habitat/tasks/nav/instance_image_nav_task.py
```diff
@@ -263,3 +263,10 @@ class InstanceImageNavigationTask(ObjectNavigationTask):
     image. Built on top of ObjectNavigationTask. Used to explicitly state a
     type of the task in config.
     """
+
+
+@registry.register_task(name="InstanceImageNav-v2")
+class InstanceImageNavigationTaskV2(InstanceImageNavigationTask):
```
vincentpierre, line 269: Is it really necessary ?

habitat-lab/habitat/tasks/nav/object_nav_task.py
```diff
@@ -183,3 +183,9 @@ class ObjectNavigationTask(NavigationTask):
     r"""An Object Navigation Task class for a task specific methods.
     Used to explicitly state a type of the task in config.
     """
+
+@registry.register_task(name="ObjectNav-v2")
+class ObjectNavigationTaskV2(ObjectNavigationTask):
```
vincentpierre, line 188: Is this new class necessary ?

habitat-lab/requirements.txt
```diff
@@ -14,3 +14,5 @@ imageio>=2.2.0
 imageio-ffmpeg>=0.2.0
 scipy>=1.0.0
 tqdm>=4.0.0
+# Home robot dependency for velocity controllers
+home-robot @ git+ssh://git@github.com/facebookresearch/home-robot.git@habitat-challenge-2023#subdirectory=src/home_robot
```
vincentpierre, line 18: Not a fan of having this dependency here. Can we make it an optional dependency ?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1403,543 @@ def step(
...
+            **kwargs,
+        )
+
+    @property
+    def action_space(self):
+        return EmptySpace()
+
+
+@registry.register_task_action
+class LookDownDiscreteToVelocityAction(WaypointAction):
```
vincentpierre, line 1917: The name of this class is rather confusing, what does it do? Don't we already have a discrete action for looking up/down ? What is the difference between the two implementations ?

habitat-lab/habitat/config/benchmark/nav/instance_imagenav/instance_imagenav_hm3d_v3.yaml
```diff
@@ -0,0 +1,34 @@
+# @package _global_
```
vincentpierre, line 1: Can all of these new benchmarks we used with the gym wrapper ? I would like to be sure they can be.

vincentpierre (commented):
Missing unit tests

habitat-lab/habitat/config/habitat/task/instance_imagenav_v3.yaml
```diff
@@ -0,0 +1,28 @@
+defaults:
```
ykarmesh, line 1: why is this file instance_imagenav_v3 when the task type is v2?

habitat-lab/habitat/config/default_structured_configs.py
ykarmesh: This is the logic inside the step function of the rearrange_stop action.
```
    def step(self, task, *args, is_last_action, **kwargs):
        should_stop = kwargs.get("rearrange_stop", [1.0])
        if should_stop[0] > 0.0:
            rearrange_logger.debug(
                "Rearrange stop action requesting episode stop."
            )
            self.does_want_terminate = True

        if is_last_action:
            return self._sim.step(HabitatSimActions.rearrange_stop)
        else:
            return {}
```
This is the logic in the velocity_stop action:
```
        self, velocity_stop, task: EmbodiedTask, *args: Any, **kwargs: Any
    ):
        r"""Update ``_metric``, this method is called from ``Env`` on each
        ``step``.
        """
        if velocity_stop[0] > 0.0:
            task.is_stop_called = True  # type: ignore

        return self._sim.get_observations_at()  # type: ignore
```
The logic is mostly different between the two.
Also rearrange stop will have to be renamed if we want to use it in Habitat 1.0 tasks so it made sense to me to create a new action.

habitat-baselines/habitat_baselines/config/instance_imagenav/ddppo_instance_imagenav_v3.yaml
jacobkrantz: Just the benchmark, but we add `rl.policy.action_distribution_type: gaussian` for the action space.

habitat-lab/habitat/tasks/nav/object_nav_task.py
ykarmesh: The idea by @devendrachaplot was that having a separate version for the current task will allow us others to refer to this task with a simple name. Otherwise people will be calling it continuous control objectnav which is much more verbose.

habitat-lab/habitat/tasks/nav/instance_image_nav_task.py
ykarmesh: The idea by @devendrachaplot was that having a separate version for the current task will allow us others to refer to this task with a simple name. Otherwise people will be calling it continuous control objectnav which is much more verbose.

habitat-lab/habitat/config/benchmark/nav/instance_imagenav/instance_imagenav_hm3d_v3.yaml
ykarmesh: Sorry I am not sure if I know how to properly do this. Do you have any pointers?

examples/tutorials/velocity_control_sim.ipynb
ykarmesh: @exhaustin

habitat-lab/requirements.txt
ykarmesh: @exhaustin

habitat-lab/habitat/config/default_structured_configs.py
ykarmesh: Modified the text

habitat-lab/habitat/tasks/nav/nav.py
jimmytyyang: Thank you! This is for making sure that we can do sim2real transfer. This function takes the discrete action command and produces the velocity command for the motor so that we can mimic the real world hardware. The original one does not have such capability and just teleport the camera.

habitat-lab/requirements.txt
exhaustin: #1183

examples/tutorials/velocity_control_sim.ipynb
exhaustin: This notebook exists to test & tune the core controller logic.
I included it so Jimmy & Karmesh could get insight into the controller behavior because we were debugging why the integrated version wasn't working back then.
I think we can remove the notebook now the continuous actions are working (and add tests to ensure that). @jimmytyyang @ykarmesh what do you guys think?

ykarmesh: Yeah I dont think I will need it. It will still be good to keep it somewhere internally maybe as a google colab? @vincentpierre Any thoughts?

vincentpierre: Code that we do not keep using will inevitably stop working after a few weeks.
I would prefer if this code was shortened, documented, tested and improved so that an external user could also benefit from it. If that is unrealistic, let's remove it.
If we want to keep it for internal reference for a whole, let's make it a gist or a colab somewhere and link to it in a code comment, but chances are it will be out of date very soon...

habitat-baselines/habitat_baselines/config/instance_imagenav/ddppo_instance_imagenav_v3.yaml
vincentpierre: Would it be possible to use v2 in the default list and then add the benchmark / action type? I think that would be better so that when we change v2 the changes will affect v3 automagically.

habitat-lab/habitat/tasks/nav/instance_image_nav_task.py
vincentpierre: This class does not seem to enforce the continuous action space is used. It looks like it is the same class by another name. We will need to be consistent with how we use it.

habitat-lab/habitat/tasks/nav/nav.py
vincentpierre: Would it be beneficial to just change the existing action to produce a velocity command?
Or maybe rename the old one to LookDownInstantaneousAction ?

habitat-lab/habitat/config/benchmark/nav/instance_imagenav/instance_imagenav_hm3d_v3.yaml
```diff
@@ -0,0 +1,34 @@
...
+        sim_sensors:
+          rgb_sensor:
+            width: 480
+            height: 640
+            hfov: 42
+            position: [0, 1.31, 0]
+          depth_sensor:
+            width: 480
+            height: 640
+            hfov: 58
```
jacobkrantz, line 26: This should be 42.
@ykarmesh, mind fixing?

ykarmesh, line 26:
```suggestion
            hfov: 42
```

habitat-lab/habitat/config/default_structured_configs.py
ykarmesh: @vincentpierre Does this answer your question

habitat-lab/habitat/config/benchmark/nav/instance_imagenav/instance_imagenav_hm3d_v3.yaml
ykarmesh: Done

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -198,13 +214,98 @@ class TeleportActionConfig(ActionConfig):
...
-    # meters/sec
-    lin_vel_range: List[float] = [0.0, 0.25]
-    # deg/sec
-    ang_vel_range: List[float] = [-10.0, 10.0]
+    lin_vel_range: List[float] = [0.0, 0.3]  # meters/sec
+    ang_vel_range: List[float] = [-0.45, 0.45]  # rad/sec
+    ang_vel_range_camera_pitch: List[float] = [-0.45, 0.45]  # rad/sec
+    ang_range_camera_pitch: List[float] = [-1.57, 0.43]  # rad
+    time_step: float = 0.1  # seconds
+    enable_scale_convert: bool = False
```
ykarmesh, line 222: Making this True by default. Also, this cannot be changed by the participants during the challenge.
```suggestion
    enable_scale_convert: bool = True
```

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1197,8 +1258,8 @@ def step(
         task: EmbodiedTask,
         linear_velocity: float,
         angular_velocity: float,
+        camera_pitch_angular_velocity: float = 0.0,
         time_step: Optional[float] = None,
-        allow_sliding: Optional[bool] = None,
```
vincentpierre, line 1201: Is the agent no longer allowed to slide? Is this a breaking change that needs to be documented?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
         EPS = 1e-5
         collided = (dist_moved_after_filter + EPS) < dist_moved_before_filter
 
-        agent_observations = self._sim.get_observations_at(
-            position=final_position,
-            rotation=final_rotation,
+        # TODO: Make a better way to flag collisions
+        self._sim._prev_sim_obs["collided"] = collided  # type: ignore
```
vincentpierre, line 1406: Please don't leave todoes in the code without an owner. What does this do and why do we need a better way to "flag collisions"?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1209,39 +1270,95 @@ def step(
...
+        )
+
+    def _apply_velocity_action(
+        self,
+        linear_velocity: float,
+        angular_velocity: float,
+        time_step: Optional[float] = None,
+    ):
+        """
+        Apply velocity command to simulation, step simulation, and return agent observation
```
vincentpierre, line 1351: Add return typing.
It looks like this returns an agent_state and not an observation

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+            )
+        elif camera_pitch_ang + delta < self._ang_range_camera_pitch[0]:
+            next_camera_pitch_ang = self._ang_range_camera_pitch[0]
+            goal_rigid_state.rotation = mn.Quaternion.rotation(
+                mn.Rad(next_camera_pitch_ang), mn.Vector3(1, 0, 0)
+            )
+
+        # Update all the sensors
+        for sensor in sensor_names:
+            self._sim.agents[0]._sensors[  # type: ignore
```
vincentpierre, line 1473: Why accessing agent 0 ? There is a standard way to grab the agent index in the code. Please use that or this code will not scale to multi-agents.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+                sensor
+            ].node.rotation = goal_rigid_state.rotation
+
+        return self._sim.get_agent_state()
```
vincentpierre, line 1477: Why is there a return on a `apply_` method ?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+
+        # Init goto velocity controller
+        try:
+            from habitat.utils.controller_wrapper import (
+                DiffDriveVelocityController,
+            )
+
+            self.w2v_controller = DiffDriveVelocityController(self._config)
+        except ModuleNotFoundError as exc:
+            additional_error_message = """
```
vincentpierre, line 1546:
```suggestion
            additional_error_message = f"""
            {exc.message}
```

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+                max_duration,
+                [0, 1],
+                [
+                    self._wait_duration_range[0],
+                    self._wait_duration_range[1],
+                ],
+            )
+
+        # Clamp
+        xyt_waypoint_clamped = np.array(
```
vincentpierre, line 1703: If inputs are scaled, the clamping has different ranges.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+        Habitat coordinate system: +y = up
+
+        We map the coordinate systems by assuming the +x axis is shared
+        x -> x
+        y -> -z
+        rz -> -ry
+        """
+        # Convert rotation to rotation vector
+        q = agent_state.rotation
+        if type(q) is mn.Quaternion:
```
vincentpierre, line 1804: in the rest of the code, we use isinstance instead. Please consider changing it

habitat-lab/habitat/utils/controller_wrapper.py
```diff
@@ -0,0 +1,47 @@
...
+        self.controller.set_yaw_tracking(track_yaw)
+
+    def set_goal(
+        self,
+        goal: np.ndarray,
+        start: Optional[np.ndarray] = None,
+        relative: bool = False,
+    ):
+        """Update controller goal
+        goal: Desired robot base SE2 pose in global frame
```
vincentpierre, line 25: docstring for each  argument

habitat-lab/requirements.txt
vincentpierre: Please remove from this file since the additional dependency is added in setup.py as per #1183

examples/shortest_path_follower_example.py
```diff
@@ -78,7 +78,9 @@ def shortest_path_example():
 
                 observations, reward, done, info = env.step(best_action)
                 im = observations["rgb"]
+
                 top_down_map = draw_top_down_map(info, im.shape[0])
+
                 output_im = np.concatenate((im, top_down_map), axis=1)
```
ykarmesh, line 84:
```suggestion
                top_down_map = draw_top_down_map(info, im.shape[0])
                output_im = np.concatenate((im, top_down_map), axis=1)
```

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -200,13 +216,98 @@ class TeleportActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, this discrete action will move the robot forward by
+    a fixed amount determined by the SimulatorConfig.forward_step_size amount.
+    """
+    type: str = "MoveForwardWaypointAction"
+    max_wait_duration: float = 3.0  # seconds
+    forward_step_size: float = 0.25  # meters
+
+
+@attr.s(auto_attribs=True, slots=True)
```
ykarmesh, line 269:
```suggestion
@dataclass
```

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -200,13 +216,98 @@ class TeleportActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, this discrete action will rotate the robot to the left
+    by a fixed amount determined by the SimulatorConfig.turn_angle amount.
+    """
+    type: str = "TurnLeftWaypointAction"
+    max_wait_duration: float = 3.0  # seconds
+    turn_angle: float = np.pi / 6  # rad (30 degrees)
+
+
+@attr.s(auto_attribs=True, slots=True)
```
ykarmesh, line 280:
```suggestion
@dataclass
```

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -200,13 +216,98 @@ class TeleportActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, this discrete action will rotate the robot to the right
+    by a fixed amount determined by the SimulatorConfig.turn_angle amount.
+    """
+    type: str = "TurnRightWaypointAction"
+    max_wait_duration: float = 3.0  # seconds
+    turn_angle: float = np.pi / 6  # rad (30 degrees)
+
+
+@attr.s(auto_attribs=True, slots=True)
```
ykarmesh, line 291:
```suggestion
@dataclass
```

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -200,13 +216,98 @@ class TeleportActionConfig(ActionConfig):
...
+    r"""
+    In Navigation tasks only, this discrete action will rotate the robot to the left
+    by a fixed amount determined by the SimulatorConfig.turn_angle amount.
+    """
+    type: str = "LookUpDiscreteToVelocityAction"
+    max_wait_duration: float = 3.0  # seconds
+    turn_angle: float = np.pi / 6  # rad (30 degrees)
+
+
+@attr.s(auto_attribs=True, slots=True)
```
ykarmesh, line 302:
```suggestion
@dataclass
```

habitat-lab/habitat/core/benchmark.py
```diff
@@ -24,14 +24,21 @@ class Benchmark:
     r"""Benchmark for evaluating agents in environments."""
 
     def __init__(
-        self, config_paths: Optional[str] = None, eval_remote: bool = False
+        self,
+        config_paths: Optional[str] = None,
+        eval_remote: bool = False,
+        action_space: Optional[str] = None,
```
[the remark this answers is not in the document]
ykarmesh: Yes, I know this is a hack but I specifically didnt want to create a list of overrides because that would allow the challenge participants to pass any number of config changes to the simulator which potentially opens the gate of the challenge to cheating.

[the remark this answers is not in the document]
ykarmesh: I am open to discussions about this though.

habitat-lab/habitat/tasks/nav/nav.py
ykarmesh: This change was done by @exhaustin. One thing to note is that the velocity action was mostly not in use in habitat 1.0, so the number of people affected by this would be minimal. But that doesnt mean that we should stop allowing this completely. It might be useful to improve training performance.

habitat-baselines/habitat_baselines/config/instance_imagenav/ddppo_instance_imagenav_v3.yaml
ykarmesh: @jacobkrantz What do you think about this?

examples/tutorials/velocity_control_sim.ipynb
jimmytyyang: Thank you, based on the above discussion, I will remove the file. Thank you everyone!

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1084,6 +1084,37 @@ def step(self, task: EmbodiedTask, *args: Any, **kwargs: Any):
         return self._sim.get_observations_at()  # type: ignore
 
 
+@registry.register_task_action
+class VelocityStopAction(SimulatorTaskAction):
+    name: str = "velocity_stop"
```
mathfac, line 1089: Need doc string, why this action is continuous as it's only discrete stop call.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1084,6 +1084,37 @@ def step(self, task: EmbodiedTask, *args: Any, **kwargs: Any):
         return self._sim.get_observations_at()  # type: ignore
 
 
+@registry.register_task_action
```
mathfac, line 1087: File `habitat-lab/habitat/tasks/nav/nav.py` is already overloaded and too long. Let's move out all continuous action related declaration into separate file something like `continuous_nav.py` or `velocity_nav.py.`

habitat-lab/habitat/tasks/nav/nav.py
mathfac, line 1089: Need doc string, why this action is continuous as it's only discrete stop call.

habitat-lab/habitat/config/benchmark/nav/instance_imagenav/instance_imagenav_hm3d_v3.yaml
jimmytyyang: Hi @vincentpierre, it seems that there is no existing code that tests this on the main branch. Could you please provide more pointers, thank you! @ykarmesh I think this code can be used with the gym wrapper

jimmytyyang: my point is that it seems that we need to create a new test script that covers all the existing yamls

habitat-lab/habitat/tasks/nav/nav.py
mathfac: If it's only related to continuous control we can remove it.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+            xyt_waypoint, self._config.max_wait_duration, *args, **kwargs
+        )
+
+    @property
+    def action_space(self):
+        return EmptySpace()
+
+
+@registry.register_task_action
+class TurnLeftWaypointAction(WaypointAction):
```
mathfac, line 1849: All this classes needs doc strings.  As well as description of different action spaces and what is Waypoint.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1402,553 @@ def step(
...
+            xyt_waypoint,
+            self._config.max_wait_duration,
+            delta_camera_pitch_angle,
+            *args,
+            **kwargs,
+        )
+
+    @property
+    def action_space(self):
+        return EmptySpace()
```
mathfac, line 1951: Why it's EmptySpace?

habitat-lab/habitat/sims/habitat_simulator/actions.py
```diff
@@ -197,8 +197,25 @@ def get(self):
 @registry.register_action_space_configuration(name="velocitycontrol")
 class HabitatSimVelocityCtrlActionSpaceConfiguration(ActionSpaceConfiguration):
     def get(self):
+        if not HabitatSimActions.has_action("VELOCITY_CTRL"):
+            HabitatSimActions.extend_action_space("VELOCITY_CTRL")
         return {
             HabitatSimActions.VELOCITY_CTRL: habitat_sim.ActionSpec(
                 "velocity_control"
             ),
+            # The perfect actions are needed for the oracle planner
```
mathfac, line 206: Is there a way to avoid it? Where exactly it is used in the code?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -201,12 +217,105 @@ class TeleportActionConfig(ActionConfig):
...
+        default_factory=lambda: [-1.57, 0.43]
+    )
+    time_step: float = 0.1  # seconds
+    enable_scale_convert: bool = True
+
+
+@dataclass
+class WaypointControlActionConfig(VelocityControlActionConfig):
+    r"""
+    In Navigation tasks only, this action will utilize a velocity controller to
```
mathfac, line 238: That would be great to mention with what types of robot this parameters work and that it enables/suitable for sim2real.

mathfac (commented):
Really like new sim2real capabilities we are adding here with new action spaces. The integration here is a little bit too aggressive: 
1. That would be great to separate the new continuous control capabilities from the desecrate regular.
2. Make it explicit what type of robot/sim2real experiment it can support.
3. Add home_robot as a dependency if that isn't too heavy one.
4. Add integration test that can check that functionality is running ok.

habitat-lab/habitat/config/habitat/task/objectnav_v2.yaml
```diff
@@ -0,0 +1,31 @@
+defaults:
```
mathfac, line 1: Should we add both tasks to https://github.com/facebookresearch/habitat-lab/blob/main/DATASETS.md as well?

habitat-lab/habitat/core/benchmark.py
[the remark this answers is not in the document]
mathfac: Yes, we actually shouldn't promote user do overwrites this way.

habitat-lab/setup.py
```diff
@@ -24,13 +24,21 @@ def get_package_version():
     return VERSION
 
 
+EXTRA_DEPS = {
+    # Home robot dependency for velocity controllers
+    "home_robot": [
+        "home-robot @ git+ssh://git@github.com/facebookresearch/home-robot.git@habitat-challenge-2023#subdirectory=src/home_robot",
```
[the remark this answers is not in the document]
vincentpierre: I am going to push back just a little bit.
This is a dependency on a branch of a repo of a package that isn't on pypi. This isn't best practice for our users. I would rather it be an extra dependency.

habitat-lab/habitat/config/benchmark/nav/objectnav/objectnav_v2_hm3d_stretch.yaml
```diff
@@ -0,0 +1,37 @@
+# @package _global_
```
ykarmesh, line 1: Make file naming consistent with instance imagenav yaml

habitat-lab/habitat/sims/habitat_simulator/actions.py
ykarmesh: The shortest path agent these actions to be available to plan the path. But it might be okay to stop supporting that with this new version of the navigation tasks. The shortest path agent can only be used by 1 of the three new action spaces anyway. That action space is discrete_waypoint_controller which is trying to replicate the old habitat 1.0 actions. If anyone wants to use the shortest path planner, they could just use the older action space and get the outputs from there.

Let me know what you think?

habitat-lab/habitat/config/habitat/task/objectnav_v2.yaml
ykarmesh: The task configs arent referenced on DATASETS.md file. However, the dataset configs are referenced. The dataset configs for the new episode datasets have already been referenced in that README on the main branch. They currently are pointing to a path in the main branch which doesnt exist but they should start working once this branch is merged.

habitat-lab/setup.py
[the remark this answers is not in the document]
ykarmesh: Following up on our conversation in the meeting. The command is pointing to a tag on the home-robot repository so this should be fine. 

Also we are printing this error msg when the home robot lib is not found
```
            Missing dependencies for waypoint type actions.
            Install habitat-lab with the 'home_robot' option to enable this feature.
            pip install -e "habitat-lab[home_robot]
```

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: @ykarmesh can you comment on this? 
My understanding is that we're expecting the policy to have some sort of a softmax output with scores for each discrete action, and since "stop" is one of the possible actions it will also be determined from a value ranging from -1 to 1.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1209,39 +1270,95 @@ def step(
...
+            self._ang_vel_range[0],
+            self._ang_vel_range[1],
+        )
+        camera_pitch_angular_velocity_clamped = np.clip(
+            camera_pitch_angular_velocity,
+            self._ang_vel_range_camera_pitch[0],
+            self._ang_vel_range_camera_pitch[1],
         )
 
-        # Stop is called if both linear/angular speed are below their threshold
```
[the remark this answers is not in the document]
exhaustin: @jimmytyyang Are there users of VelocityAction that would be affected by this change? I remember having a discussion between us and @ykarmesh where the conclusion is that implicit stops by thresholding are not desired.

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1209,39 +1270,95 @@ def step(
...
+        angular_velocity_clamped = np.clip(
+            angular_velocity,
+            self._ang_vel_range[0],
+            self._ang_vel_range[1],
+        )
+        camera_pitch_angular_velocity_clamped = np.clip(
+            camera_pitch_angular_velocity,
+            self._ang_vel_range_camera_pitch[0],
+            self._ang_vel_range_camera_pitch[1],
         )
```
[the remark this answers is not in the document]
exhaustin: We actually don't clamp the inputs during scaling, so they would exceed system limits if the input is not within [-1, 1]. Is that something of concern?

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: This `_apply_velocity_action` method was originally within the `step` method of `VelocityAction`, implemented by @jimmytyyang.

I wanted to wrap all of this into a helper method because 1) the process of applying the velocity and integrating the sim is surprisingly complicated and having all of this in `step` will make the interface method unreadable; and 2) I wanted WaypointAction to also be able to use this. However I didn't have much context on how any of this worked so I just threw everything into a method, including the part where the next state of the agent was being queried at the end, which also seemed non-trivial.

Would love to know if there's a clean way to apply velocity so that we can directly query the next state. with `sim.get_agent_state()`

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: If `_enable_scale_convert is True`, then they would have been denormalized in the previous block of code.

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: We are re-raising the error with `raise ModuleNotFoundError(additional_error_message) from exc`, which should display the original error message. Is there another reason we would need to explicitly print out the error message?

habitat-lab/habitat/tasks/nav/nav.py
exhaustin: Because the action type has no parameters. Like calling `MoveFowardWaypointAction` always moves the robot forward by the amount specified in configs.

Personally I think it makes more sense to define a `CameraDiscreteAction` and have parameters specify whether you want to look down, look up, etc. But I don't know habitat well enough so I just went with whatever's there.

habitat-lab/habitat/config/benchmark/nav/objectnav/objectnav_v2_hm3d_stretch.yaml
jimmytyyang: Done!

jimmytyyang:
> 1. desecrate

Hi @mathfac, I have a few questions about his:
(1) What do you mean by "new continuous control capabilities from the desecrate regular."?
(2) Right now habitat-lab does not have the true hardware code, could you please provide more examples on "Add integration test"? Thank you!

habitat-lab/habitat/tasks/nav/nav.py
jimmytyyang: Hi @mathfac. I think the logics is that once the velocity is below the certain threshold, the robot just calls the Stop action. We want to make this consistent across the velocity control

habitat-lab/habitat/tasks/nav/nav.py
wangyenjen: @exhaustin, if that's the case, would you consider leaving a TODO note?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -199,14 +215,125 @@ class TeleportActionConfig(ActionConfig):
...
+    """
+    type: str = "TurnRightWaypointAction"
+    max_wait_duration: float = 3.0  # seconds
+    turn_angle: float = np.pi / 6  # rad (30 degrees)
+
+
+@dataclass
+class LookUpDiscreteToVelocityActionConfig(WaypointControlActionConfig):
+    r"""
+    In Navigation tasks only, this discrete action will rotate the robot to the left
```
wangyenjen, line 320: Is the doc string, `In Navigation tasks only, this discrete action will rotate the robot to the left`, correct?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -199,14 +215,125 @@ class TeleportActionConfig(ActionConfig):
...
+    """
+    type: str = "LookUpDiscreteToVelocityAction"
+    max_wait_duration: float = 3.0  # seconds
+    turn_angle: float = np.pi / 6  # rad (30 degrees)
+
+
+@dataclass
+class LookDownDiscreteToVelocityActionConfig(WaypointControlActionConfig):
+    r"""
+    In Navigation tasks only, this discrete action will rotate the robot to the right
```
wangyenjen, line 331: Is the doc string, `In Navigation tasks only, this discrete action will rotate the robot to the right`, correct?

wangyenjen (commented):
Hi @jimmytyyang,

I want to express my gratitude for your valuable contribution to the continuous controller for the navigation task. However, I noticed that you have made numerous modifications in `nav.py`. I would like to suggest considering @mathfac's suggestion of splitting `nav.py` into two files. Alternatively, we could add an additional flag to enable the new feature while ensuring the original discrete controller can still function properly. What are your thoughts on this?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1408,553 @@ def step(
...
+            camera_pitch_angular_velocity = (
+                self.w2v_controller.velocity_feedback_control(
+                    camera_pitch_angular_err,
+                    self._acc_ang_camera_pitch,
+                    self._w_max_camera_pitch,
+                )
+            )
+
+            # Apply action and step simulation
+            next_agent_state = self._apply_camera_pitch_velocity_action(
```
wangyenjen, line 1777:
```suggestion
           self._apply_camera_pitch_velocity_action(
```

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1289,16 +1408,553 @@ def step(
...
+        ``step``.
+        """
+        xyt_waypoint = np.array([self._forward_step_size, 0.0, 0.0])
+        return self._step_rel_waypoint(
+            xyt_waypoint, self._config.max_wait_duration, *args, **kwargs
+        )
+
+    @property
+    def action_space(self):
+        return EmptySpace()
```
wangyenjen, line 1851: Are there any plans for introducing a new action space? If so, it would be helpful to leave a TODO note regarding this.

wangyenjen (changes requested):

PR #1132, opened 2023-02-12
Agent Factory

ASzot:
## Motivation and Context

Refactored trainer code to interface with a new agent abstraction which wraps the rollouts, policy, and updaters.

Note that the changes are not applied to `ver_trainer` yet.

## Types of changes

### DIFF
--- a/habitat-baselines/habitat_baselines/rl/ppo/trainer_agent.py
+++ b/habitat-baselines/habitat_baselines/rl/ppo/trainer_agent.py
@@ -1,0 +1,283 @@
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple
+
+import gym.spaces as spaces
+import numpy as np
+import torch
+from torch.optim.lr_scheduler import LambdaLR
+
+from habitat import logger
+from habitat_baselines.common.baseline_registry import baseline_registry
+from habitat_baselines.common.rollout_storage import (  # noqa: F401.
+    RolloutStorage,
+)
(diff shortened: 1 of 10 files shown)


habitat-baselines/habitat_baselines/rl/ppo/trainer_agent.py
```diff
@@ -0,0 +1,283 @@
...
+            self.envs.num_envs,
+            obs_space,
+            self.policy_action_space,
+            self._ppo_cfg.hidden_size,
+            num_recurrent_layers=self.actor_critic.num_recurrent_layers,
+            is_double_buffered=self._ppo_cfg.use_double_buffered_sampler,
+        )
+        self.rollouts.to(self.device)
+
+        self._lr_sched = None
```
Skylion007, line 103: Why set it to None first?

habitat-baselines/habitat_baselines/rl/ppo/trainer_agent.py
```diff
@@ -0,0 +1,283 @@
...
+    HierarchicalPolicy,
+)
+from habitat_baselines.rl.ppo import PPO  # noqa: F401.
+from habitat_baselines.rl.ppo.policy import Policy
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
+
+
+class TrainerAgent:
```
Skylion007, line 27: Can this be an attrs class with slots? It would simplify this logic a bit and potentially make it a bit more efficient.

PR #1160, opened 2023-02-24
Velocity controller updates

ykarmesh:
## Motivation and Context

We need to have a continuous equivalent of the discrete STOP action that can be trained with the new ObjectNav policy.
I have added a new action called velocity_stop to refer to the continuous action space STOP. The new stop needs an input argument which should be +ve if the agent needs to stop. 

Other changes:
- Removed the old heuristic for stopping
- Removed the `cs.store` related to the VelocityControlActionConfig since it was being repeated twice.
- Changed ActionSpace to spaces.Dict() since ActionSpace is considered a discrete action space where you can choose between the different actions provided to it

## How Has This Been Tested

By starting the training of an agent.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/nav/nav.py
+++ b/habitat-lab/habitat/tasks/nav/nav.py
@@ -1086,6 +1086,34 @@ class StopAction(SimulatorTaskAction):
 
 
 @registry.register_task_action
+class VelocityStopAction(SimulatorTaskAction):
+    name: str = "velocity_stop"
+
+    def reset(self, task: EmbodiedTask, *args: Any, **kwargs: Any):
+        task.is_stop_called = False  # type: ignore
+
+    def step(self, velocity_stop, task: EmbodiedTask, *args: Any,  **kwargs: Any):
+        r"""Update ``_metric``, this method is called from ``Env`` on each
+        ``step``.
+        """
+        if velocity_stop[0] > 0.0:
+            task.is_stop_called = True  # type: ignore
+
+        return self._sim.get_observations_at()  # type: ignore
+
+    @property
+    def action_space(self):
+        return spaces.Dict(
+            {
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1298,10 +1312,10 @@ def _apply_velocity_action(
...
 
         # Map velocity actions
-        self.vel_control.linear_velocity = np.array(
+        self.vel_control.linear_velocity = mn.Vector3(
             [0.0, 0.0, -linear_velocity]
         )
-        self.vel_control.angular_velocity = np.array(
+        self.vel_control.angular_velocity = mn.Vector3(
             [0.0, angular_velocity, 0.0]
         )
```
exhaustin, line 1320: Out of curiosity, what difference does this make? Is it just for typing?

exhaustin (approved):
Looks good, interesting choice of making a continuous stop action. Is this to prevent extra action preprocessing (thresholding) code on the user side?

habitat-lab/habitat/tasks/nav/nav.py
ykarmesh: The angular_velocity and linear_velocity are already in np.array format. When you do np.array([0.0, 0.0, -linear_velocity]), it gives an error. I saw they were using mn.Vector3 in Habitat 2.0 controller so I also tried it and that worked. Basically mn.Vector3 is somehow handling that assertion.

ykarmesh:
> Looks good, interesting choice of making a continuous stop action. Is this to prevent extra action preprocessing (thresholding) code on the user side?

Yes, It is being used in Habitat 2.0 so I just decided to use it here as well to prevent changes on the habitat-baselines side

PR #1162, opened 2023-02-27
Object substring search fix

JHurricane96:
## Motivation and Context

Optimization for loading objects. Replace expensive substring searches with O(1) string matching.
## How Has This Been Tested

Manually by running episode generation and rearrangement tasks.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/utils/common.py
+++ b/habitat-lab/habitat/utils/common.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 from os import makedirs
 from os import path as osp
 from typing import List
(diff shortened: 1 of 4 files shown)


habitat-lab/habitat/utils/common.py
```diff
@@ -30,6 +31,13 @@ def check_make_dir(directory_path: str) -> bool:
     return True
 
 
+def get_project_root() -> str:
```
ASzot, line 34: This will break if the location of this method changes or if the relative location of the data path changes. We need to fix this because I think it is very likely users will have a modified data path. Can you instead get the path from the object manager?

JHurricane96: Fixed this to get the path from config.

ASzot (approved):
LGTM

PR #1172, opened 2023-03-01
@dataclass-style structured configs to @attrs-style

rpartsey:
## Motivation and Context

- Finish @dataclass-style structured configs to @attr.s-style @attr.s(auto_attribs=True, slots=True).
- Migrated code to modern attrs API

## Types of changes

## Checklist


vincentpierre (approved):
LGTM, did you do some testings to make sure we do not have issues with overwriting configs?

habitat-lab/habitat/datasets/rearrange/run_episode_generator.py
```diff
@@ -26,77 +26,70 @@
     from habitat.config import DictConfig
 
 
-@dataclass
+@attrs.define(auto_attribs=True)
 class SceneSamplerParamsConfig:
     scene: str = "v3_sc1_staging_00"
-    scene_sets: List[Any] = field(default_factory=list)
+    scene_sets: List[Any] = []
```
Skylion007, line 32: I am against doing this as this technically bad behavior. I would rather we stick to dataclasses and wrap it with this: https://pypi.org/project/dataslots/

Skylion007 (changes requested):
We should consider sticking with dataclasses. They are better supported and I don't see any performance gains from doing it here. The are some nice features like validators, but we aren't using them currently. One of the nice things about having these files as dataclasses is that we can define them without Hydra. Not having a attrs.Factory of fields initializer creates bugprone behavior since our default args are mutable.

rpartsey:
This PR can actually be decomposed into two: 
1.  structured config `@dataclass`-style definition to `@attr.s`-style definition (update code that wasn't updated in https://github.com/facebookresearch/habitat-lab/pull/1120)
2. migrating our code to modern attrs API (should still be valid)

rpartsey:
> LGTM, did you do some testings to make sure we do not have issues with overwriting configs?

I didn't observe anything during development if structured configs (parent and child) are defined the same style. Though, I didn't test extensively.

habitat-lab/habitat/datasets/rearrange/run_episode_generator.py
Skylion007: The main issue here is that the list can be edited and it will affect the DEFAULT parameters of the config which is unexpected behavior and bugprone.

rpartsey:
Closing in favour of https://github.com/facebookresearch/habitat-lab/pull/1182.

PR #1189, opened 2023-03-08
Episode generation fixes

JHurricane96:
More episode generation fixes!

- Reset sim after every episode
- Adjust y-axis displacement of objects before snapping
- Check if all 4 corners of objects are on receptacle
- Don't throw error when target object is removed due to failing stability tests
- Fix check_if_point_on_surface for TriangleMeshReceptacle: need to first transform to local coordinate frame
- Add height and size filters for receptacle surfaces
- Enhance navigability filter for receptacle surfaces by checking all 4 corners

## Motivation and Context

Fixing and speeding up episode generation.

## How Has This Been Tested

Manually by running episode generation and verifying output visually.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
@@ -9,6 +9,7 @@ import random
 import time
 from collections import defaultdict
 from typing import Dict, List, Optional, Tuple
+from habitat.tasks.rearrange.utils import get_aabb
 
 import magnum as mn
 import numpy as np
@@ -271,11 +272,8 @@ class ObjectSampler:
                         habitat_sim.utils.common.random_quaternion()
                     )
             if isinstance(receptacle, TriangleMeshReceptacle):
-                new_object_height = (
-                    new_object.root_scene_node.cumulative_bb.size()[1]
-                )
                 new_object.translation = new_object.translation + mn.Vector3(
-                    0, new_object_height / 2, 0
+                    0, 0.05, 0
                 )
 
             if isinstance(receptacle, OnTopOfReceptacle):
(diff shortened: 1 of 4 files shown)


habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -271,11 +272,8 @@ def sample_placement(
                         habitat_sim.utils.common.random_quaternion()
                     )
             if isinstance(receptacle, TriangleMeshReceptacle):
-                new_object_height = (
-                    new_object.root_scene_node.cumulative_bb.size()[1]
-                )
                 new_object.translation = new_object.translation + mn.Vector3(
-                    0, new_object_height / 2, 0
+                    0, 0.05, 0
```
ASzot, line 276: Why can you assume a fixed offset? Don't you need to adjust by the object height?

habitat-lab/habitat/datasets/rearrange/samplers/object_target_sampler.py
```diff
@@ -69,6 +69,7 @@ def sample(
             self.object_instance_set, target_receptacles, goal_receptacles
         ):
             if object_to_containing_receptacle[use_target.handle] != use_recep:
+                return None
```
ASzot, line 72: Remove the exception raising after.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -738,17 +758,33 @@ def get_navigable_receptacles(
...
+            global_bounds.front_bottom_right
         ]
-        for point in recep_points:
-            if is_accessible(sim, point, nav_to_min_distance):
-                navigable_receptacles.append(receptacle)
-                logger.info(f"Receptacle {receptacle.parent_object_handle}, {receptacle_obj.translation} is navigable.")
-                break
+        is_navigable = all(is_accessible(sim, point, nav_to_min_distance) for point in recep_points)
+
+        if is_navigable:
```
ASzot, line 785: Change from `is_navigable` to something like `all_is_accessible` since these points are not navigable, they are only accessible.

ASzot (approved):
With the exception of above comments, LGTM.

habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
JHurricane96: Alex and I had a discussion about this, and the issue seems to be with the receptacle surface annotations clipping a bit into the receptacle mesh, and it's unrelated to the object height. Quoting Alex's response here:

> I agree that in this case moving the object first is not likely to cause an issue, but I don't think this is generally true. For example, objects in a drawer or on a shelf could be bumped up inside of something else.
The [snap_down](https://github.com/facebookresearch/habitat-lab/blob/864d6b503b61806fb7370c68453448dece593704/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py#L231) function should already account for the height of the object. It projects the COM downward and then adjusts for the object's size. The only case in which this would be a problem is if the mesh receptacle surface is actually inside of the collision shape. Maybe we should check for this issue instead.
Alternatively, I might be more comfortable with a small fixed scalar offset for the COM, something like 10cm, which is very unlikely to push an object's COM into another part of the mesh.

habitat-lab/habitat/datasets/rearrange/samplers/object_target_sampler.py
JHurricane96: Removed.

PR #1194, opened 2023-03-10
[Bug Fix] fix sensor metrics

jimmytyyang:
## Motivation and Context

 The variable `self._max_accum_force` is created, but it is never be used in `ForceTerminate`. Fix the bug here so that the episode would be terminated if the accumulate force exceeds the threshold  `self._max_accum_force`.

## How Has This Been Tested

 Run the CI test

## Types of changes

## Checklist


ASzot (approved):
Good catch. Thanks for finding this!

PR #1208, opened 2023-03-20
[Development] add new collision detection for Spot/Stretch robot

jimmytyyang:
## Motivation and Context

 The Spot robot from Boston Dynamics has an oblong shape. This is problematic since we assume the agent's shape is a cylinder, which prevents us from getting a correct mesh for Spot and makes it easy to collide with the obstacles. Our solution is to have multiple cylinders aligned to approximate an oblong shape. See an illustration figure below. 

Screenshot 2023-03-20 at 4 35 36 PM

In addition, we also benchmark the Spot training speed and found that the new collision shape introduces a 10% of speed decrease (from 1257 to 1112 FPS), which is acceptable given that we still get 1k FPS.

Moreover, we design a new class of action space, called `BaseVelSpotAction`, instead of refactoring the original `BaseVelAction`. This is to prevent changing the behavior (training speed) of the original codebase.
 
This is a joint work with @aclegg3 !

## How Has This Been Tested

 Ran the VER training on the fair cluster to benchmark the speed, and `interactive_play.py`

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -28,8 +28,8 @@ habitat:
       base_velocity_spot:
         allow_dyn_slide: False
         cylinder_deviate: 0.3
-        lin_collision_threshold: 0.04
-        ang_collision_threshold: 0.01
+        lin_collision_threshold: 1e-5
+        ang_collision_threshold: 1e-5
   simulator:
     type: RearrangeSim-v0
     seed: 100
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -28,12 +28,10 @@ habitat:
       base_velocity_non_cylinder:
         allow_dyn_slide: False
         # There is a collision if the difference between the clamped NavMesh position and target position
-        # is more than than lin_collision_threshold/ang_collision_threshold for any point
-        lin_collision_threshold: 1e-5
-        ang_collision_threshold: 1e-5
+        # is more than than collision_threshold for any point
+        collision_threshold: 1e-5
         # The x and y locations of the clamped NavMesh position
-        x_offset: [0.3, -0.3]
-        y_offset: [0.0, 0.0]
+        navmesh_offset: [[0.3, 0], [-0.3, 0]]
         # If we allow the robot to move laterally
         enable_lateral_move: True
   simulator:
@@ -41,6 +39,8 @@ habitat:
     seed: 100
     additional_object_paths:
       - "data/objects/ycb/configs/"
+      # - "/Users/jimmytyyang/Downloads/test_scene_dataset/object_datasets/amazon_berkeley/configs/"
+      # - "/Users/jimmytyyang/Downloads/test_scene_dataset/object_datasets/google_object_dataset/configs/"
     agents:
       main_agent:
         radius: 0.3
@@ -59,3 +59,7 @@ habitat:
       enable_physics: True
   dataset:
     data_path: data/datasets/replica_cad/rearrange/v1/{split}/all_receptacles_10k_1k.json.gz
+    # data_path: /Users/jimmytyyang/floor_planner/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_5_0303.json.gz
+    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_1_0303_102344307.json.gz
+    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_scene3_obj10_en10.json.gz
+    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/fp_episodes/rearrange/microtrain/rearrange_floorplanner.json.gz
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -34,7 +34,7 @@ habitat:
         # is more than than collision_threshold for any point
         collision_threshold: 1e-5
         # The x and y locations of the clamped NavMesh position
-        navmesh_offset: [[0.3, 0], [-0.3, 0]]
+        navmesh_offset: [[0.0, 0.0], [0.25, 0.0], [-0.25, 0.0]]
         # If we allow the robot to move laterally
         enable_lateral_move: True
   simulator:
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -3,7 +3,7 @@
 defaults:
   - /habitat: habitat_config_base
   - /habitat/task/actions:
-    - base_velocity_spot
+    - base_velocity_non_cylinder
   - /habitat/simulator/agents@habitat.simulator.agents.main_agent: rgbd_head_rgbd_arm_agent
   - /habitat/task/rearrange: play
   - /habitat/dataset/rearrangement: replica_cad
@@ -25,11 +25,13 @@ habitat:
         disable_grip: False
         delta_pos_limit: 0.0125
         ee_ctrl_lim: 0.015
-      base_velocity_spot:
+      base_velocity_non_cylinder:
         allow_dyn_slide: False
         cylinder_deviate: 0.3
         lin_collision_threshold: 1e-5
         ang_collision_threshold: 1e-5
+        x_offset: [0.3, -0.3]
+        y_offset: [0.0, 0.0]
   simulator:
     type: RearrangeSim-v0
     seed: 100
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -28,7 +28,7 @@ habitat:
       base_velocity_non_cylinder:
         allow_dyn_slide: False
         # There is a collision if the difference between the clamped NavMesh position and target position
-        # is too great than lin_collision_threshold/ang_collision_threshold for any point
+        # is more than than lin_collision_threshold/ang_collision_threshold for any point
         lin_collision_threshold: 1e-5
         ang_collision_threshold: 1e-5
         # The x and y locations of the clamped NavMesh position
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -2,6 +2,8 @@
 
 defaults:
   - /habitat: habitat_config_base
+  - /habitat/task/actions:
+    - base_velocity_spot
   - /habitat/simulator/agents@habitat.simulator.agents.main_agent: rgbd_head_rgbd_arm_agent
   - /habitat/task/rearrange: play
   - /habitat/dataset/rearrangement: replica_cad
@@ -23,6 +25,11 @@ habitat:
         disable_grip: False
         delta_pos_limit: 0.0125
         ee_ctrl_lim: 0.015
+      base_velocity_spot:
+        allow_dyn_slide: False
+        cylinder_deviate: 0.3
+        lin_collision_threshold: 0.04
+        ang_collision_threshold: 0.01
   simulator:
     type: RearrangeSim-v0
     seed: 100
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -27,10 +27,15 @@ habitat:
         ee_ctrl_lim: 0.015
       base_velocity_non_cylinder:
         allow_dyn_slide: False
+        # There is a collision if the difference between the clamped NavMesh position and target position
+        # is too great than lin_collision_threshold/ang_collision_threshold for any point
         lin_collision_threshold: 1e-5
         ang_collision_threshold: 1e-5
+        # The x and y locations of the clamped NavMesh position
         x_offset: [0.3, -0.3]
         y_offset: [0.0, 0.0]
+        # If we allow the robot to move laterally
+        enable_lateral_move: True
   simulator:
     type: RearrangeSim-v0
     seed: 100
@@ -38,7 +43,7 @@ habitat:
       - "data/objects/ycb/configs/"
     agents:
       main_agent:
-        radius: 0.55
+        radius: 0.3
         articulated_agent_urdf: data/robots/hab_spot_arm/urdf/hab_spot_arm.urdf
         articulated_agent_type: "SpotRobot"
         sim_sensors:
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -27,7 +27,6 @@ habitat:
         ee_ctrl_lim: 0.015
       base_velocity_non_cylinder:
         allow_dyn_slide: False
-        cylinder_deviate: 0.3
         lin_collision_threshold: 1e-5
         ang_collision_threshold: 1e-5
         x_offset: [0.3, -0.3]
--- a/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
+++ b/habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
@@ -39,8 +39,6 @@ habitat:
     seed: 100
     additional_object_paths:
       - "data/objects/ycb/configs/"
-      # - "/Users/jimmytyyang/Downloads/test_scene_dataset/object_datasets/amazon_berkeley/configs/"
-      # - "/Users/jimmytyyang/Downloads/test_scene_dataset/object_datasets/google_object_dataset/configs/"
     agents:
       main_agent:
         radius: 0.3
@@ -59,7 +57,3 @@ habitat:
       enable_physics: True
   dataset:
     data_path: data/datasets/replica_cad/rearrange/v1/{split}/all_receptacles_10k_1k.json.gz
-    # data_path: /Users/jimmytyyang/floor_planner/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_5_0303.json.gz
-    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_1_0303_102344307.json.gz
-    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/rearrange_floorplanner/train/rearrange_floorplanner_scene3_obj10_en10.json.gz
-    # data_path: /Users/jimmytyyang/habitat_lab_0301/habitat-lab/data/datasets/fp_episodes/rearrange/microtrain/rearrange_floorplanner.json.gz
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -260,12 +260,10 @@ class BaseVelocityNonCylinderActionConfig(ActionConfig):
     # If we allow the robot to move back or not
     allow_back: bool = True
     # There is a collision if the difference between the clamped NavMesh position and target position
-    # is more than lin_collision_threshold/ang_collision_threshold for any point.
-    lin_collision_threshold: float = 1e-5
-    ang_collision_threshold: float = 1e-5
+    # is more than collision_threshold for any point.
+    collision_threshold: float = 1e-5
     # The x and y locations of the clamped NavMesh position
-    x_offset: Optional[List[float]] = None
-    y_offset: Optional[List[float]] = None
+    navmesh_offset: Optional[List[float]] = None
     # If we allow the robot to move laterally.
     enable_lateral_move: bool = False
 
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -245,6 +245,21 @@ class BaseVelocityActionConfig(ActionConfig):
 
 
 @dataclass
+class BaseVelocitySpotActionConfig(ActionConfig):
+    r"""
+    In Rearrangement only. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
+    """
+    type: str = "BaseVelSpotAction"
+    lin_speed: float = 10.0
+    ang_speed: float = 10.0
+    allow_dyn_slide: bool = False
+    allow_back: bool = True
+    cylinder_deviate: float = 0.0
+    lin_collision_threshold: float = 0.0
+    ang_collision_threshold: float = 0.0
+
+
+@dataclass
 class HumanoidJointActionConfig(ActionConfig):
     r"""
     In Rearrangement only. Corresponds to actions to change the humanoid joints. Contains the parameter num_joints, indicating the joints that can be modified.
@@ -1536,6 +1551,12 @@ cs.store(
     group="habitat/task/actions",
     name="base_velocity",
     node=BaseVelocityActionConfig,
+)
+cs.store(
+    package="habitat.task.actions.base_velocity_spot",
+    group="habitat/task/actions",
+    name="base_velocity_spot",
+    node=BaseVelocitySpotActionConfig,
 )
 cs.store(
     package="habitat.task.actions.humanoidjoint_action",
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -255,8 +255,8 @@ class BaseVelocitySpotActionConfig(ActionConfig):
     allow_dyn_slide: bool = False
     allow_back: bool = True
     cylinder_deviate: float = 0.0
-    lin_collision_threshold: float = 0.0
-    ang_collision_threshold: float = 0.0
+    lin_collision_threshold: float = 1e-5
+    ang_collision_threshold: float = 1e-5
 
 
 @dataclass
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -247,17 +247,26 @@ class BaseVelocityActionConfig(ActionConfig):
 @dataclass
 class BaseVelocityNonCylinderActionConfig(ActionConfig):
     r"""
-    In Rearrangement only. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
+    In Rearrangement only for the non cylinder shape of the robot. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
     """
     type: str = "BaseVelNonCylinderAction"
+    # The max linear speed of the robot
     lin_speed: float = 10.0
+    # The max angular speed of the robot
     ang_speed: float = 10.0
+    # If we want to do sliding or not
     allow_dyn_slide: bool = False
+    # If we allow the robot to move back or not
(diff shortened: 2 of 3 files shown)


jimmytyyang:
Added collision detection conditions

Screenshot 2023-03-20 at 4 55 19 PM

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+            target_rigid_state.rotation.to_matrix(), end_pos
+        )
+        self.cur_articulated_agent.sim_obj.transformation = target_trans
+
+        did_coll = self.spot_collison_check(trans, target_trans, end_pos)
+
+        if not self._allow_dyn_slide:
+            # Check if in the new articulated_agent state the arm collides with anything.
+            # If so we have to revert back to the previous transform
+            self._sim.internal_step(-1)
```
aclegg3, line 596: Shouldn't be stepping the simulator here.
This is general comment across all robot code as I assume this is being done elsewhere also.
If we need to reset the collision cache or something we should instead use a discrete collision detection. If there is another reason to do this, it should be factored out so we don't advance sim time arbitrarily.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+        )
+
+        # For step filter of the front pos
+        end_front_pos = self._sim.step_filter(front_cur_pos, front_goal_pos)
+        end_front_pos[1] = front_cur_pos[1]
+        front_goal_pos[1] = front_cur_pos[1]
+        # For step filter of the rear pos
+        end_rear_pos = self._sim.step_filter(rear_cur_pos, rear_goal_pos)
+        end_rear_pos[1] = front_cur_pos[1]
+        rear_goal_pos[1] = front_cur_pos[1]
```
aclegg3, line 542: I'm assuming you want to project all of these points to the XZ plane so distances are not polluted by changes in height.
```suggestion
        end_rear_pos[1] = rear_cur_pos[1]
        rear_goal_pos[1] = rear_cur_pos[1]
```

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+        # For step filter of the center pos
+        end_pos[1] = trans.translation[1]
+
+        # Robot moving distance
```
aclegg3, line 546:
```suggestion
        # Planar move distance clamped by NavMesh
```

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+        front_move = (end_front_pos - front_goal_pos).length()
+        rear_move = (end_rear_pos - rear_goal_pos).length()
+
+        # For detection of linear or angualr velocity
+        center_move = (end_pos - trans.translation).length()
+
+        # If the acutal target and the resulting moving position
+        # is too samll, then there is a collision
```
aclegg3, line 554:
```suggestion
        # There is a collision if the difference between the clamped NavMesh position and target position is too great for any point.
```

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
```diff
@@ -23,6 +25,11 @@ habitat:
         disable_grip: False
         delta_pos_limit: 0.0125
         ee_ctrl_lim: 0.015
+      base_velocity_spot:
+        allow_dyn_slide: False
+        cylinder_deviate: 0.3
+        lin_collision_threshold: 0.04
+        ang_collision_threshold: 0.01
```
aclegg3, line 32: This seems dangerous, especially with continuous control.
Currently we enforce a maximum error of 1e-5 (see [agent action controls](https://github.com/facebookresearch/habitat-sim/blob/8f9cc41a7d87894032da70c95f6b427c7faf113d/src_python/habitat_sim/agent/controls/object_controls.py#L75)).
I feel like we should have the same requirements here. Any planar error is the result of a collision, right?

aclegg3 (changes requested):
This approach seems functional for Spot, but I think we should revisit design a bit before merging.

Instead of a per-robot solution, I feel this could be abstracted to a general robot step filter function with per-robot config variations. Each robot should provide a list of 2D offsets which will all be sequentially checked. For Spot, this would be front, middle, back. For Stretch this would be 1 point offset from the center. Everything else could live in the base level robot module. Happy to catch up on this offline. :)

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Thanks for this!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Thanks for this!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Good catch! Thanks for this!

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
jimmytyyang: Thank you for pointing out this! I make it to be small!

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
```diff
@@ -23,6 +25,11 @@ habitat:
         disable_grip: False
         delta_pos_limit: 0.0125
         ee_ctrl_lim: 0.015
+      base_velocity_spot:
+        allow_dyn_slide: False
+        cylinder_deviate: 0.3
+        lin_collision_threshold: 0.04
```
akshararai: could you add short comments on what each of these parameters mean?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -244,6 +244,21 @@ class BaseVelocityActionConfig(ActionConfig):
     allow_back: bool = True
 
 
+@dataclass
+class BaseVelocitySpotActionConfig(ActionConfig):
+    r"""
+    In Rearrangement only. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
+    """
+    type: str = "BaseVelSpotAction"
+    lin_speed: float = 10.0
```
akshararai, line 253: silly question, but is this meant to be the max_lin_speed? If we can't rename the variable, can we add a comment clarifying this?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -244,6 +244,21 @@ class BaseVelocityActionConfig(ActionConfig):
     allow_back: bool = True
 
 
+@dataclass
+class BaseVelocitySpotActionConfig(ActionConfig):
+    r"""
+    In Rearrangement only. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
```
akshararai, line 250: ``
```suggestion
    In Rearrangement only. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation. While Spot can move laterally as well, we ignore this degree of freedom, and only allow longitudinal and angular motion.
```

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
             return {}
 
 
+@registry.register_task_action
+class BaseVelSpotAction(ArticulatedAgentAction):
```
akshararai: Should this class inherit from BaseVelAction? Then maybe some of the parameters below do not need to be re-defined?

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+        self._allow_dyn_slide = self._config.get("allow_dyn_slide", True)
+        self._lin_speed = self._config.lin_speed
+        self._ang_speed = self._config.ang_speed
+        self._allow_back = self._config.allow_back
+        self._cylinder_deviate = self._config.cylinder_deviate
+        self._lin_collision_threshold = self._config.lin_collision_threshold
+        self._ang_collision_threshold = self._config.ang_collision_threshold
+
+    @property
+    def action_space(self):
```
akshararai, line 499: is this the same as action_space() from BaseVelAction?

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,163 @@ def step(self, *args, is_last_action, **kwargs):
...
+            "vel": self.cur_articulated_agent.sim_obj.joint_velocities,
+            "pos": self.cur_articulated_agent.sim_obj.joint_positions,
+        }
+
+    def _set_articulated_agent_state(self, set_dat):
+        self.cur_articulated_agent.sim_obj.joint_positions = set_dat["forces"]
+        self.cur_articulated_agent.sim_obj.joint_velocities = set_dat["vel"]
+        self.cur_articulated_agent.sim_obj.joint_forces = set_dat["pos"]
+
+    def spot_collison_check(self, trans, target_trans, end_pos):
```
akshararai: I believe that this would be the only new function in this class (over what is already defined in BaseVelAction)?

akshararai: Add documentation of what the inputs of this function are.

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
```diff
@@ -23,6 +25,12 @@ habitat:
         disable_grip: False
         delta_pos_limit: 0.0125
         ee_ctrl_lim: 0.015
+      base_velocity_non_cylinder:
+        allow_dyn_slide: False
+        lin_collision_threshold: 1e-5
```
akshararai, line 30: can we have very short comments on what these parameters mean?

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,161 @@ def step(self, *args, is_last_action, **kwargs):
             return {}
 
 
+@registry.register_task_action
+class BaseVelNonCylinderAction(ArticulatedAgentAction):
```
akshararai, line 473: Should this class inherit from BaseVelAction?

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,161 @@ def step(self, *args, is_last_action, **kwargs):
...
+        return spaces.Dict(
+            {
+                self._action_arg_prefix
+                + "base_vel": spaces.Box(
+                    shape=(2,), low=-lim, high=lim, dtype=np.float32
+                )
+            }
+        )
+
+    def collision_check(
```
akshararai, line 510: can you add comment on what each of these inputs mean?

jimmytyyang:
> This approach seems functional for Spot, but I think we should revisit design a bit before merging.
> 
> Instead of a per-robot solution, I feel this could be abstracted to a general robot step filter function with per-robot config variations. Each robot should provide a list of 2D offsets which will all be sequentially checked. For Spot, this would be front, middle, back. For Stretch this would be 1 point offset from the center. Everything else could live in the base level robot module. Happy to catch up on this offline. :)

Hi Alex. Thank you so much for this! I refactored the entire function to something more generic `BaseVelNonCylinderAction()`, which deals with the non-cylinder body shape. Now it takes two additional arguments `self._config.x_offset` and `self._config.y_offset`, which specify the offset positions of the new cylinder to the original cylinder. In addition, I choose not to change the original action class `BaseVelAction` since we want to be cautious about the speed impact of each commit, as Andrew suggested.

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
jimmytyyang: Thank you Akshara! I will add the comments.

habitat-lab/habitat/config/default_structured_configs.py
jimmytyyang: Yes, thank you! I added the comments

habitat-lab/habitat/config/default_structured_configs.py
jimmytyyang: Cool! thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: good point! I will think about this

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Yes, it is.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Done! thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Done, thank you!

jimmytyyang:
@akshararai and @aclegg3 I refactored the code and now it looks more generic. I also add literal movement as an option so that the robot can move literally.

jimmytyyang:
https://user-images.githubusercontent.com/55121504/227041848-7ef0e0ba-e51e-4355-a1e0-eb237b0f9eec.mp4

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+            target_rigid_state.rotation.to_matrix(), end_pos
+        )
+
+        did_coll, new_target_trans = self.collision_check(
+            trans, target_trans, end_pos, target_rigid_state
+        )
+        self.cur_articulated_agent.sim_obj.transformation = new_target_trans
+
+        if not self._allow_dyn_slide:
+            if did_coll:
```
akshararai, line 602: do we need `self._sim.internal_step(-1)` here? There are some subtle differences from update_base() in BaseVelAction(). Could you clarify why those changes are needed? I'm trying to minimize code repeats as much as possible, and introducing any bugs.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+                self.cur_articulated_agent.sim_obj.transformation = trans
```
akshararai, line 603: We should not allow a step if there was collision and sliding is off.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+
+        if (lin_diff > 0) and center_move != 0:
+            return True, new_target_trans
+        elif (ang_diff > 0) and center_move == 0:
+            return True, new_target_trans
+        else:
+            return False, new_target_trans
+
+    def update_base(self):
+        ctrl_freq = self._sim.ctrl_freq
```
akshararai, line 577: Overall I am a little unclear why we need this over the update_base() in BaseVelAction. I would try to re-use code as much as possible, and only add very necessary changes. If needed, re-structure the original update_base() to be more functional, so that the functions can be re-used.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+            lin_vel, ang_vel = kwargs[self._action_arg_prefix + "base_vel"]
+
+        lin_vel = np.clip(lin_vel, -1, 1) * self._lin_speed
+        lateral_lin_vel = np.clip(lateral_lin_vel, -1, 1) * self._lin_speed
+        ang_vel = np.clip(ang_vel, -1, 1) * self._ang_speed
+        if not self._allow_back:
+            lin_vel = np.maximum(lin_vel, 0)
+
+        self.base_vel_ctrl.linear_velocity = mn.Vector3(
+            lin_vel, 0, -lateral_lin_vel
```
akshararai, line 631: why `-lateral_lin_vel`?

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -244,6 +244,31 @@ class BaseVelocityActionConfig(ActionConfig):
...
 
 
+@dataclass
+class BaseVelocityNonCylinderActionConfig(ActionConfig):
+    r"""
+    In Rearrangement only for the non cylinder shape of the robot. Corresponds to the base velocity. Contains two continuous actions, the first one controls forward and backward motion, the second the rotation.
+    """
+    type: str = "BaseVelNonCylinderAction"
+    # The max linear speed of the robot
+    lin_speed: float = 10.0
```
akshararai, line 254: if we allow lateral velocity, can we add a max speed for that as well, instead of sharing lin_speed in the 2 dimensions?

akshararai (changes requested):

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+    """
+    The articulated agent base motion is constrained to the NavMesh and controlled with velocity commands integrated with the VelocityControl interface.
+
+    Optionally cull states with active collisions if config parameter `allow_dyn_slide` is True
+    """
+
+    def __init__(self, *args, config, sim: RearrangeSim, **kwargs):
+        super().__init__(*args, config=config, sim=sim, **kwargs)
+        self._lin_collision_threshold = self._config.lin_collision_threshold
+        self._ang_collision_threshold = self._config.ang_collision_threshold
```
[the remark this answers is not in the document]
jimmytyyang: Hi Akshara! Sorry, I think they are not in the base class. These two variables are for the function class.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+            return spaces.Dict(
+                {
+                    self._action_arg_prefix
+                    + "base_vel": spaces.Box(
+                        shape=(3,), low=-lim, high=lim, dtype=np.float32
+                    )
+                }
+            )
+        else:
+            return spaces.Dict(
```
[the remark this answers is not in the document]
jimmytyyang: Hi Akshara! Sorry, I do not think they are the same since there is a if-statement here. Thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: yes, this is exactly what I am doing here. Thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: This is due to the coordinate of Habitat, so we add negative sign here. Thank you!

habitat-lab/habitat/config/default_structured_configs.py
jimmytyyang: Done, thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Hi, Akshara! The current design of the original class `BaseVelAction` prevents me from using the inheritance. I will need to do a refactor of `BaseVelAction`. Thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+        rigid_state = habitat_sim.RigidState(
+            mn.Quaternion.from_matrix(trans.rotation()), trans.translation
+        )
+
+        target_rigid_state = self.base_vel_ctrl.integrate_transform(
+            1 / ctrl_freq, rigid_state
+        )
+        end_pos = self._sim.step_filter(
+            rigid_state.translation, target_rigid_state.translation
+        )
```
aclegg3, line 600: I don't think it is true that the center position should always be a checked against the navmesh. I feel (0,0) should be added to the cylinder offsets if you want to use the center (e.g. Stretch has 1 navmesh point and it isn't the center). 

There could be special handling to check for (0,0) and speed it up instead of doing the matrix multiplication if that is important for perf.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+
+        # For detection of linear or angualr velocities
+        # There is a collision if the difference between the clamped NavMesh position and target position is too great for any point.
+        lin_diff = len([v for v in move if v > self._lin_collision_threshold])
+        ang_diff = len([v for v in move if v > self._ang_collision_threshold])
+
+        if (lin_diff > 0) and center_move != 0:
+            return True, new_target_trans
+        elif (ang_diff > 0) and center_move == 0:
+            return True, new_target_trans
```
aclegg3, line 583: I'm questioning this separation between linear and angular error. It seems to me like you don't need any knowledge of the center position or this distinction. Unless we empirically see a need for different error values, I suggest simplifying and unifying these.

aclegg3 (changes requested):
Much closer, thanks for iterating on this. :) 
I have a couple of design thoughts still. Happy to chat if it would help.

habitat-lab/habitat/config/benchmark/rearrange/play_spot.yaml
```diff
@@ -23,14 +25,25 @@ habitat:
...
         ee_ctrl_lim: 0.015
+      base_velocity_non_cylinder:
+        allow_dyn_slide: False
+        # There is a collision if the difference between the clamped NavMesh position and target position
+        # is too great than lin_collision_threshold/ang_collision_threshold for any point
+        lin_collision_threshold: 1e-5
+        ang_collision_threshold: 1e-5
+        # The x and y locations of the clamped NavMesh position
+        x_offset: [0.3, -0.3]
+        y_offset: [0.0, 0.0]
```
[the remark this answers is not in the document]
jimmytyyang: Thank you!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Hi Alex, I think this line code is needed since we need this to compute the position the other cylinders. In addition, the robot will move based on the end_pos if there is no collision. Moreover, there will be a case that the center of the robot actually needs to be checked. Thank you, and I am happy to discuss more!

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Make sense! We want to make sure that we do not arbitrarily advance the simulation steps.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
jimmytyyang: Good catch! Thank you, Alex! I unified them.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -469,6 +469,178 @@ def step(self, *args, is_last_action, **kwargs):
...
+
+        # Planar move distance clamped by NavMesh
+        move = []
+        for i in range(num_check_cylinder):
+            move.append((end_pos[i] - goal_pos[i]).length())
+        center_move = (center_end_pos - trans.translation).length()
+
+        # Wrap the move direction if we use sliding
+        # Find the largest diff moving direction, which means that there is a collision in that cylinder
+        max_idx = np.argmax(move + [center_move])
```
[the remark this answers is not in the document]
jimmytyyang: Yes, it will be applied no matter what the value of `if_sliding ` is. If `if_sliding` is true, then we will slide the robot, based on the computation from this function `collision_check `

akshararai (approved):
thanks for making these changes!

jimmytyyang:
sliding is on in the video
https://user-images.githubusercontent.com/55121504/227624558-9b7562a2-d563-4e89-a8cd-7e3056ad36d5.mp4

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -470,6 +470,218 @@ def step(self, *args, is_last_action, **kwargs):
...
+                target_trans.transform_point(
+                    np.array(
+                        [
+                            self._navmesh_offset[i][0],
+                            0,
+                            self._navmesh_offset[i][1],
+                        ]
+                    )
+                )
+            )
```
aclegg3, line 555: Nit: slightly more concise. You'll need to re-run pre-commit for format correction.
```suggestion
nav_pos_3d = [np.array([xz[0], 0, xz[1]]) for xz in self._navmesh_offset]
cur_pos = [trans.transform_point(xyz) for xyz in nav_pos_3d]
goal_pos = [target_trans.transform_point(xyz) for xyz in nav_pos_3d]
```

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -470,6 +470,218 @@ def step(self, *args, is_last_action, **kwargs):
...
+            self.cur_articulated_agent.sim_obj.transformation = (
+                new_target_trans
+            )
+        else:
+            if did_coll:
+                self.cur_articulated_agent.sim_obj.transformation = trans
+            else:
+                self.cur_articulated_agent.sim_obj.transformation = (
+                    target_trans
+                )
```
aclegg3, line 633: If `new_target_trans` returned by `collision_check` was always the correct final transform (see above suggestion), this could be reduced to:
```suggestion
    self.cur_articulated_agent.sim_obj.transformation = new_target_trans
```

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -470,6 +470,218 @@ def step(self, *args, is_last_action, **kwargs):
...
+            new_target_trans = None
+
+        # For detection of linear or angualr velocities
+        # There is a collision if the difference between the clamped NavMesh position and target position is too great for any point.
+        diff = len([v for v in move if v > self._collision_threshold])
+
+        if diff > 0:
+            return True, new_target_trans
+        else:
+            return False, new_target_trans
```
aclegg3, line 591: Instead of None, consider always returning the finalized transform for easier downstream use:
```suggestion
        
        # For detection of linear or angualr velocities
        # There is a collision if the difference between the clamped NavMesh position and target position is too great for any point.
        diff = len([v for v in move if v > self._collision_threshold])
        
        if diff > 0:
            if compute_sliding:
                max_idx = np.argmax(move)
                move_vec = end_pos[max_idx] - cur_pos[max_idx]
                new_end_pos = trans.translation + move_vec
                return True, mn.Matrix4.from_(
                    target_rigid_state.rotation.to_matrix(), new_end_pos
                )
            return True, trans
        else:
            return False, target_trans
```

aclegg3 (approved):
Logic looks good, thanks for the iterations :) 
Remaining suggestions here are just aimed to simplify the implementation a bit.

Though my suggestions don't use it, I still think returning the `did_collide` boolean is useful. :+1: 

Note: you'll want to incorporate these locally and re-run pre-commit to fix formatting.

jimmytyyang:
> Logic looks good, thanks for the iterations :) Remaining suggestions here are just aimed to simplify the implementation a bit.
> 
> Though my suggestions don't use it, I still think returning the `did_collide` boolean is useful. 👍
> 
> Note: you'll want to incorporate these locally and re-run pre-commit to fix formatting.

thank you alex for working with me on this! I incorporated with your suggestions!

PR #1216, opened 2023-03-23
[Experiment] Humanoid rest pose siro

xavierpuigf:
## Motivation and Context

Adds a reset state for the human, to avoid the T shape. This could also go as a reset skill, which is forced to run at the beginning of the episode. The current solution causes minimal disruption to the training flow.

## How Has This Been Tested

Trained policies with humanoid

## Types of changes

## Checklist


ASzot (approved):
LGTM. No need to get tests to pass.

PR #1249, opened 2023-04-07
Add support for glTF multi-receptacle assets

aclegg3:
## Motivation and Context

Current `TriangleMeshReceptacle` parsing supports only single mesh PLY files. This PR adds support for multi-mesh glTF files.

In the multi-mesh case, a unique `TriangleMeshReceptacle` instance is created for each sub-mesh.

For example, if 3 sub-meshes exist in the `1234.glb` file referenced from receptacle metadata like:
```
"receptacle_mesh_1234": {
      "name": "receptacle_mesh_1234",
      "mesh_filepath": "my_rec_meshes/1234.glb"
  }
```
The parser will produce 3 TriangleMeshReceptacles named:
```
"receptacle_mesh_1234.0000"
"receptacle_mesh_1234.0001"
"receptacle_mesh_1234.0002"
```

NOTE some un-quiet-able logging is expected from receptacle loading and import initialization.
This will be fixed soon in a follow-up change to Magnum APIs introducing log quieting functionality.
E.g.:
```
Trade::GltfImporter::scene(): node 0 extras modelObb property is Utility::JsonToken::Type::Object, skipping
...
PluginManager::Manager: duplicate static plugin AnySceneImporter, ignoring
```
## How Has This Been Tested

Locally with .glb and .ply files.
Should update test_assets in habitat-sim to include new .glb receptacles.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
@@ -535,25 +535,25 @@ def import_tri_mesh(mesh_file: str) -> List[mn.trade.MeshData]:
 
     :param mesh_file: The input meshes file. NOTE: must contain only triangles.
     """
-    _importer = _manager.load_and_instantiate("AnySceneImporter")
-    _importer.open_file(mesh_file)
+    importer = _manager.load_and_instantiate("AnySceneImporter")
+    importer.open_file(mesh_file)
 
     mesh_data: List[mn.trade.MeshData] = []
 
     # import mesh data and pre-process
     mesh_data = [
-        filter_interleave_mesh(_importer.mesh(mesh_ix))
-        for mesh_ix in range(_importer.mesh_count)
+        filter_interleave_mesh(importer.mesh(mesh_ix))
+        for mesh_ix in range(importer.mesh_count)
     ]
 
     # if there is a scene defined, apply any transformations
-    if _importer.scene_count > 0:
-        scene_id = _importer.default_scene
+    if importer.scene_count > 0:
+        scene_id = importer.default_scene
         # If there's no default scene, load the first one
         if scene_id == -1:
             scene_id = 0
 
-        scene = _importer.scene(scene_id)
+        scene = importer.scene(scene_id)
 
         # Mesh referenced by mesh_assignments[i] has a corresponding transform in
         # mesh_transformations[i]. Association to a particular node ID is stored in
--- a/habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
@@ -7,7 +7,7 @@
 import os
 import random
 from abc import ABC, abstractmethod
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -505,33 +505,161 @@ def get_all_scenedataset_receptacles(
...
 
-    mesh_ix = 0
-    mesh = importer.mesh(mesh_ix)
-    assert (
-        mesh.primitive == mn.MeshPrimitive.TRIANGLES
-    ), "Must be a triangle mesh."
+    # Import mesh data from a scene format (e.g. glTF)
+    scene_ix = importer.default_scene
+    if scene_ix != -1:
+        scene_data = importer.scene(scene_ix)
```
mosra, line 548: With https://github.com/facebookresearch/habitat-sim/pull/2066 this whole code could be replaced with something like this, assuming the goal is to get back a list of meshes that have their positions transformed according to the scene hierarchy. Basically it's about porting [this example snippet](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1SceneTools.html#ab1505270c538ebe54e23c78676d7a780) to Python:

```py
meshes = [importer.mesh(i) for i in range(importer.mesh_count)]

# Mesh referenced by mesh_assignments[i] has a corresponding transform in
# mesh_transformations[i]. Association to a particular node ID is stored in 
# scene.mapping(mn.trade.SceneField.MESH)[i], but it's not needed for anything
# here.
mesh_assignments: cr.containers.StridedArrayView1D = \
    scene.field(mn.trade.SceneField.MESH)
mesh_transformations: List[mn.Matrix4] = \
    mn.scenetools.flatten_transformation_hierarchy3d(scene, mn.trade.SceneField.MESH)
assert len(mesh_assignments) == len(mesh_transformations)

# A mesh can be referenced by multiple nodes, so this can't operate in-place.
flattened_meshes: List[mn.trade.MeshData] = []
for mesh_id, transformation in zip(mesh_assignments, mesh_transformations):
    flattened_meshes += [mn.meshtools.transform3d(mesh_id, transformation)]
```

The `flattened_meshes` is then very close to the `mesh_data` array here. You may however still want to store the `MeshData` instances or (references to) the position arrays from them instead of a `List[mn.Vector3]`, as that's *significantly* more memory-efficient (as far as I can see, each `mn.Vector3` instance stored in a list is 64 *bytes*, a lot more than the three floats per position when stored inside `MeshData`). 

If you'd want to reduce the memory use even more, you could discard also all unused attributes except positions. I.e. do this in the mesh loading code:

```py
mesh = ...
# This operation affects only the metadata -- the remaining attributes stay in 
# the vertex buffer but are no longer referenced and thus treated like padding
mesh = mn.meshtools.filter_only_attributes(mesh, [mn.trade.MeshAttribute.POSITION])
# This forces a repack of the mesh, throwing away all unused data (by default 
# it keeps the original layout including all padding if it's already interleaved)
mesh = mn.meshtools.interleave(mesh, mn.meshtools.InterleaveFlag.NONE)
```

(Wrote this off the top of my head, can't guarantee it compiles or works. But should be pretty close to a working state.)

aclegg3: Thanks for the API updates and suggestions here.
I refactored the code to use the new APIs and took your suggestion to store the `MeshData` directly instead of lists.

However, I ran into some issues with the memory improvement suggestions. I think these could be added in a later PR, but for reference:
1. `mesh = mn.meshtools.filter_only_attributes(mesh, [mn.trade.MeshAttribute.POSITION])`
Seems to corrupt the `indices`:
```
len(self.mesh_data.indices) = 6
len(self.mesh_data.attribute(mn.trade.MeshAttribute.POSITION)) = 4
index = 0
self.mesh_data.indices[index] = 32349
```
2. `mesh = mn.meshtools.interleave(mesh, mn.meshtools.InterleaveFlag.NONE)` maybe need some additional plugin or compile flags?
```
AttributeError: module '_magnum.meshtools' has no attribute 'InterleaveFlag'
```

aclegg3: Also, I'm getting the following terminal output even with `export MAGNUM_LOG=quiet`:

```
PluginManager::Manager: duplicate static plugin AssimpImporter, ignoring
PluginManager::Manager: duplicate static plugin AnySceneImporter, ignoring
PluginManager::Manager: duplicate static plugin AnyImageImporter, ignoring
Trade::GltfImporter::scene(): node 0 extras modelNormal property is Utility::JsonToken::Type::Object, skipping
Trade::GltfImporter::scene(): node 0 extras modelObb property is Utility::JsonToken::Type::Object, skipping
Trade::GltfImporter::scene(): node 1 extras modelNormal property is Utility::JsonToken::Type::Object, skipping
Trade::GltfImporter::scene(): node 1 extras modelObb property is Utility::JsonToken::Type::Object, skipping
Trade::GltfImporter::scene(): node 2 extras modelNormal property is Utility::JsonToken::Type::Object, skipping
Trade::GltfImporter::scene(): node 2 extras modelObb property is Utility::JsonToken::Type::Object, skipping

```

mosra: > `export MAGNUM_LOG=quiet`

Unlike with `HABITAT_LOG`, this affects only the startup log when creating a GL context, not everything. Implementing that would either mean `getenv()` on every print (which is very slow), or need some global "logging context" singleton which -- while completely fine in Habitat -- is a rather nasty thing to have in a library / middleware setting and would introduce more problems than it would solve.

From the application perspective, silencing the output is possible with scoped redirection, such as this in C++ ([relevant docs](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html#Utility-Debug-scoped-output)):

```cpp
{
    Mn::Warning redirectWarnings{nullptr};
    // all warnings coming from Magnum are silenced now
}

// all warnings coming from Magnum are printed again
```

Hm, so maybe I could expose something similar to Python as well. I'm thinking it could be doable with a context manager and a `with` statement, although I'd need to look up how to do that with pybind. Would this make sense?

```py
import magnum as mn

with mn.redirect_warnings(None):
    # all warnings coming from Magnum are silenced now
    
# all warnings coming from Magnum are printed again
```

Another idea I had was to introduce `Flag::Quiet` on the importer interface, and make the plugins respect it. There's `Flag::Verbose` already. I'll add that, this seems good to have in any case.

> `Trade::GltfImporter::scene(): node 0 extras modelNormal property is Utility::JsonToken::Type::Object, skipping`

I was asking about these on Slack a while ago: https://cvmlp.slack.com/archives/C048W359BFC/p1677178392534799?thread_ts=1677014872.077569&cid=C048W359BFC

My goal is to make them all easily available through Python, so you could for example get them as `scene.field(importer.scene_field_for_name('modelNormal'))`, which also means having them implicitly preserved though all possible scene transformations and exported back when you save a file again. There's two options as I said in that thread -- either those get changed to not be in nested objects so the existing importer code can fetch them, or I update the glTF import/export code to handle them, somehow. I didn't get to the latter yet.

mosra: > `module '_magnum.meshtools' has no attribute 'InterleaveFlag'`

It's `InterleaveFlags`, sorry. (I just typed that code snippet out from memory.)

> Seems to corrupt the `indices`

Uh oh. Working on getting that fixed :wink:  The root cause is that `meshtools.filter_*_attributes()` modify the metadata but reference the original mesh, and Python wasn't made aware of the original still being referenced. I'm making some additional changes to have these accidents caught automatically to prevent this from happening again.

I think this should work for now, I'll ping you once I have the overal robustness improved:

```py
mesh = ...
# `mesh_filtered` now refers to data in `mesh`
# (`mesh_filtered.vertex_data_flags` isn't `OWNED`)
# until i make a proper fix, it's important to use a different variable
# which should tell Python's GC to not delete the original
mesh_filtered = mn.meshtools.filter_only_attributes(mesh, [mn.trade.MeshAttribute.POSITION])

# makes the mesh standalone / self-owned again
# (`mesh.vertex_data_flags` is `OWNED`)
mesh = mn.meshtools.interleave(mesh_filtered, mn.meshtools.InterleaveFlags.NONE)
```

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -505,33 +496,58 @@ def get_all_scenedataset_receptacles(
...
+        mesh_transformations: List[
+            mn.Matrix4
+        ] = mn.scenetools.flatten_transformation_hierarchy3d(
+            scene, mn.trade.SceneField.MESH
+        )
+        assert len(mesh_assignments) == len(mesh_transformations)
+
+        # A mesh can be referenced by multiple nodes, so this can't operate in-place.
+        flattened_meshes: List[mn.trade.MeshData] = [
+            mn.meshtools.transform3d(importer.mesh(mesh_id), transformation)
```
mosra, line 535: If the same mesh is referenced more than once (such as is the case with chairs, etc.), this will unnecessarily perform the (complex) import multiple times. So probably better to do what you do in the `else` clause here as well, and only then update the meshes with the flattened transforms:

```py
    mesh_data: List[mn.trade.MeshData] = [
        importer.mesh(mesh_id) for mesh_id in range(importer.mesh_count)
    ]
    
    if importer.scene_count > 0:
        ...
    
        # A mesh can be referenced by multiple nodes, so this can't operate in-place
        # i.e., len(mesh_data) likely changes after this step
        mesh_data = [
            mn.meshtools.transform3d(mesh_data[mesh_id], transformation)
            for mesh_id, transformation in zip(
                mesh_assignments, mesh_transformations
            )
        ]
        
    # no else here
```

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -505,33 +496,58 @@ def get_all_scenedataset_receptacles(
...
+            for mesh_id, transformation in zip(
+                mesh_assignments, mesh_transformations
+            )
+        ]
+
+        for mesh in flattened_meshes:
+            assert (
+                mesh.primitive == mn.MeshPrimitive.TRIANGLES
+            ), "Must be a triangle mesh."
```
mosra, line 544: There's also a new [meshtools.generate_indices()](https://doc.magnum.graphics/python/magnum/meshtools/#generate_indices) utility that can convert `TRIANGLE_STRIP` and `TRIANGLE_FAN` primitives to `TRIANGLES`. Those aren't that common, but some datasets have them, so it'd be an additional step for robustness:

```py
            if mesh.primitive in [mn.MeshPrimitive.TRIANGLE_STRIP, mn.MeshPrimitive.TRIANGLE_FAN]:
                mesh = mn.meshtools.generate_indices(mesh)
            assert (
                mesh.primitive == mn.MeshPrimitive.TRIANGLES
            ), "Must be a triangle mesh."
```

[See C++ docs for details.](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a8b0d9114105a36fbd2147ee129e2017d)

Hm, also probably best to do this during the initial import at the very top, again to not have to do that several times for the same mesh in case of duplicated chairs and such.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: Thanks, this appears to be working now.

aclegg3: > Hm, so maybe I could expose something similar to Python as well. ...
> ```python
> with mn.redirect_warnings(None):
>     # all warnings coming from Magnum are silenced now
> ```
> ...
> Another idea I had was to introduce `Flag::Quiet` on the importer interface, and make the plugins respect it. ...

Either of these should functionally solve the problem from my perspective. In general a solution to quieting logs would be useful once we get to scaled training or running batched processes. :+1: 

> > `Trade::GltfImporter::scene(): node 0 extras modelNormal property is Utility::JsonToken::Type::Object, skipping`
> 
> I was asking about these on Slack a while ago: https://cvmlp.slack.com/archives/C048W359BFC/p1677178392534799?thread_ts=1677014872.077569&cid=C048W359BFC

I generally agree that accessing these nested objects would be valuable in the long run, but currently I just want to suppress the noise. :slightly_smiling_face:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: Thanks, applied this on the initial import step

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -506,6 +506,25 @@ def import_tri_mesh(mesh_file: str) -> List[mn.trade.MeshData]:
     importer = manager.load_and_instantiate("AnySceneImporter")
     importer.open_file(mesh_file)
 
+    def filter_interleave_mesh(mesh: mn.trade.MeshData) -> mn.trade.MeshData:
+        """
+        Filter all but position data and interleave a mesh to reduce overall memory footprint.
+        """
+        # `mesh_filtered` now refers to data in `mesh`
+        # (`mesh_filtered.vertex_data_flags` isn't `OWNED`)
+        # until a proper fix, it's important to use a different variable
```
mosra, line 527: I have the fix almost done, just need to wait until my CIs pass :)

If you can hold off with merging this until tomorrow, I should manage to open a PR upgrading the relevant repos on the sim side by then. Possibly even including the `Flag::Quiet` on the importers (which seems easier from the two options I outlined).

aclegg3: Yeah, I think that sounds reasonable. Aim to merge this tomorrow with remaining improvements. Thanks for the support here. :slightly_smiling_face:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -502,9 +506,8 @@ def import_tri_mesh(mesh_file: str) -> List[mn.trade.MeshData]:
 
     :param mesh_file: The input meshes file. NOTE: must contain only triangles.
     """
-    manager = mn.trade.ImporterManager()
-    importer = manager.load_and_instantiate("AnySceneImporter")
-    importer.open_file(mesh_file)
+
+    _importer.open_file(mesh_file)
```
mosra, line 510: Last suggestion (sorry) -- now that you have the importer global as well, it might be good to call `_importer.close()` at the end to free up the memory it used for the last file. Otherwise this is done implicitly only when the instance gets GC'd or when `open_file()` is called again.

You could also keep it local like before and have just the manager global. Compared to plugin manager setup, the actual plugin instantiation is quite lightweight. But both approaches are fine, it's just a matter of preference.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -505,33 +500,82 @@ def get_all_scenedataset_receptacles(
...
-    importer.open_file(ply_file)
-
-    # TODO: We don't support mesh merging or multi-mesh parsing currently
-    if importer.mesh_count > 1:
-        raise NotImplementedError(
-            "Importing multi-mesh receptacles (mesh merging or multi-mesh parsing) is not supported."
+
+    _importer.open_file(mesh_file)
+
+    def filter_interleave_mesh(mesh: mn.trade.MeshData) -> mn.trade.MeshData:
```
vincentpierre, line 512: Can this be put outside of `import_tri_mesh` and be made a more general utility? `import_tri_mesh` is already a fairly large method.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
vincentpierre: I would prefer we did not call close on a global importer if possible, but if the speed gains are substantials, I am okay with it.

vincentpierre (commented):
I

vincentpierre (approved):

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: The importer still exists after it is closed, it should just clear the data cache instead of sticking around after its last use.

aclegg3: I moved the importer out of global scope. That should be fine. I think it was just the manager that shouldn't be recreated often.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -505,34 +499,85 @@ def get_all_scenedataset_receptacles(
...
+    return mesh
+
+
+def import_tri_mesh(mesh_file: str) -> List[mn.trade.MeshData]:
+    """
+    Returns a list of MeshData objects from a mesh asset using magnum trade importer.
+
+    :param mesh_file: The input meshes file. NOTE: must contain only triangles.
+    """
+    _importer = _manager.load_and_instantiate("AnySceneImporter")
```
mosra, line 538:
```suggestion
    importer = _manager.load_and_instantiate("AnySceneImporter")
```

This probably counts as trolling now, sorry.

mosra (approved):
:trollface:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
mosra: Just to clarify on this -- the `close()` doesn't make the global importer unusable or in some unspecified state. It's just freeing internal memory, which would otherwise be freed on the next call to `open_file()`, which might also happen never.

PR #1251, opened 2023-04-07
Removing Habitat v1 style actions and move to v2 style

vincentpierre:
## Motivation and Context

Following up on our discussion last Tuesday, I made this PR that moves SOME of our sim actions to lab.
Basically, I changed the logic so that actions no longer return observations (except when they do) and the actions no longer step the simulation physics (except when they do). They do when the action is implemented in sim (only move forward, turn left and turn right) because otherwise we cannot do pathfinding and others.
Overall the code for adding new actions is cleaner (see examples/new_actions.py ) and the stepping is also simpler. Please have a look.

Why not move all the actions to lab ? We need sim actions for navigation to calculate the number of actions needed to solve a navigation task in the episode generator. Hence, we need the sim actions to be used in the task.
Couldn't you just reimplement the actions in lab the same way they are in sim? Yes, but I would need to make sure the action configuration are the same in sim and lab, which can be problematic since it forces the sim config to know about the navigation action configs. 
Why are we trying to change the way actions are used? We are moving towards having the lab actions as a default since they are easier to use, they are the only actions habitat2.0 uses. 

## How Has This Been Tested

Unit tests pass, plus I used a script to make sure the new actions (like tilt camera) do the same thing.

## Types of changes

## Checklist


habitat-lab/habitat/core/embodied_task.py
```diff
@@ -295,45 +295,43 @@ def reset(self, episode: Episode):
...
             action_name in self.actions
         ), f"Can't find '{action_name}' action in {self.actions.keys()}."
         task_action = self.actions[action_name]
-        observations.update(
-            task_action.step(
-                **action["action_args"],
-                task=self,
-                is_last_action=is_last_action,
-            )
+        return task_action.step(
```
vincentpierre, line 308: Most of the time, task_action will return None, but for some legacy reason, it will sometimes return the observations. This is because sim.step(action) returns observations. In order to avoid rendering observations twice, we will use the observations computed in task_action when they are available.

habitat-lab/habitat/core/embodied_task.py
```diff
@@ -295,45 +295,43 @@ def reset(self, episode: Episode):
...
                     action,
                     episode,
-                    i == len(action_name) - 1,
                 )
         else:
-            self._step_single_action(
-                observations, action_name, action, episode
+            observations = self._step_single_action(
+                action_name, action, episode
             )
```
vincentpierre, line 328: After this code, if observations is None, then we will need to get the observations from the simulator and lab sensors. If it is not None, this means the environment was stepped by the action, so we DO NOT call sim.step()

habitat-lab/habitat/core/registry.py
```diff
@@ -181,23 +181,6 @@ def register_dataset(cls, to_register=None, *, name: Optional[str] = None):
             "dataset", to_register, name, assert_type=Dataset
         )
 
-    @classmethod
-    def register_action_space_configuration(
```
vincentpierre, line 185: We deprecate registering new action spaces for sim. If you want to add new actions, you can directly add them to the action space of lab. YOU CAN NO LONGER create new sim actions. The only sim actions we have will be hardcoded. They are `move forward` `turn left` and `turn right`. We use them for pathfinding and other tools.

habitat-lab/habitat/sims/habitat_simulator/habitat_simulator.py
```diff
@@ -384,9 +384,28 @@ def create_sim_config(
             sensor_specifications.append(sim_sensor_cfg)
 
         agent_config.sensor_specifications = sensor_specifications
-        agent_config.action_space = registry.get_action_space_configuration(
-            self.habitat_config.action_space_config
-        )(self.habitat_config).get()
+
+        agent_config.action_space = {
```
vincentpierre, line 388: This is where the sim action space is hard coded.

habitat-lab/habitat/tasks/rearrange/actions/actions.py
```diff
@@ -60,19 +60,14 @@ def reset(self, *args, **kwargs):
         super().reset(*args, **kwargs)
         self.does_want_terminate = False
 
-    def step(self, task, *args, is_last_action, **kwargs):
+    def step(self, task, *args, **kwargs):
```
vincentpierre, line 63: Getting rid of `is_last_action` for simplicity.

test/test_sensors.py
```diff
@@ -591,14 +583,6 @@ def test_noise_models_rgbd():
             "RedwoodDepthNoiseModel"
         )
 
-        config.habitat.simulator.action_space_config = "pyrobotnoisy"
```
vincentpierre, line 594: Deprecating pyrobotnoisy and its test

habitat-lab/habitat/core/embodied_task.py
```diff
@@ -295,45 +295,43 @@ def reset(self, episode: Episode):
...
 
+        self._sim.step_physics(1.0 / 60.0)  # type:ignore
```
ASzot, line 330: What is this call? Why is this using a fixed 60Hz? This didn't seem to be here before, why add it?

ASzot:
All the HabitatSim action registration code in`habitat-lab/habitat/tasks/rearrange/__init__.py` should also be removed.

habitat-lab/habitat/core/embodied_task.py
vincentpierre: The simulator method always steps physics with a dt of 1/60 (see [here](https://github.com/facebookresearch/habitat-sim/blob/main/src_python/habitat_sim/simulator.py#L503))
Before, each action was calling sim.step which is INCREDIBLY INEFFICIENT WHEN THERE ARE MORE THAN ONE ACTION, so instead, the actions now just perform the action and so not step the simulator. The stepping of the simulator happens in `EmbodiedTask`. Since there is no action to perform and the simulator just the physics needs to be stepped, we do not call sim.step (since this method takes as input an action that does not exist) but instead, we manually step the physics with sim.step_physics.

vincentpierre:
> All the HabitatSim action registration code in `habitat-lab/habitat/tasks/rearrange/__init__.py` should also be removed.

Thanks, I think I just forgot about those. I removed them now.

ASzot (approved):

examples/new_actions.py
```diff
@@ -12,169 +12,132 @@
...
+    # Get the state of the agent
+    agent_state = sim.get_agent_state()
+    # Convert from np.quaternion (quaternion.quaternion) to mn.Quaternion
+    normalized_quaternion = agent_state.rotation
+    agent_mn_quat = mn.Quaternion(
+        normalized_quaternion.imag, normalized_quaternion.real
     )
-    strafe_angle = np.deg2rad(strafe_angle)
+    forward = agent_mn_quat.transform_vector(-mn.Vector3.z_axis())
+    strafe_angle = np.deg2rad(strafe_angle_degrees)
```
Skylion007, line 47: Nit: I think you can do this entire with Magnum calls.

examples/new_actions.py
```diff
@@ -12,169 +12,132 @@
...
+        self._sim = sim
+        self._move_amount = config.move_amount
+        self._noise_amount = config.noise_amount
+
     def _get_uuid(self, *args, **kwargs) -> str:
         return "strafe_left"
 
     def step(self, *args, **kwargs):
-        return self._sim.step(HabitatSimActions.STRAFE_LEFT)
+        print(
```
Skylion007, line 81: Should we be trying to use a logger? here

habitat-lab/habitat/core/embodied_task.py
```diff
@@ -295,45 +295,43 @@ def reset(self, episode: Episode):
...
+            **action["action_args"],
+            task=self,
         )
 
     def step(self, action: Dict[str, Any], episode: Episode):
         action_name = action["action"]
         if "action_args" not in action or action["action_args"] is None:
             action["action_args"] = {}
-        observations: Any = {}
+        observations: Any = None
```
Skylion007, line 317:
```suggestion
        observations: Optional = None
```

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1040,6 +1040,24 @@ def update_metric(
...
+    def __init__(self, *args, config, sim, **kwargs):
+        super().__init__(*args, config=config, sim=sim, **kwargs)
+        self._sim = sim
+        self._tilt_angle = config.tilt_angle
+
+    def _move_camera_vertical(self, amount: float):
+        assert (
+            len(self._sim.agents) == 1  # type: ignore
+        ), "For navigation tasks, there can be only one agent in the scene"
+        sensor_names = list(self._sim.agents[0]._sensors.keys())  # type: ignore
```
Skylion007, line 1053: Why assign a temporary list here? Just iterate through the keys / dict directly?

habitat-lab/habitat/tasks/nav/nav.py
```diff
@@ -1040,6 +1040,24 @@ def update_metric(
...
+        for sensor_name in sensor_names:
+            sensor = self._sim.agents[0]._sensors[sensor_name].node  # type: ignore
+            sensor.rotation = sensor.rotation * mn.Quaternion.rotation(
+                mn.Rad(np.deg2rad(amount)), mn.Vector3.x_axis()
```
Skylion007, line 1057: You can use Magnum degree to rad here

examples/new_actions.py
vincentpierre: Besides random uniform and deg2rad, everything is magnum calls.
These were there in the original implementation, so I think I will leave them here.

habitat-lab/habitat/tasks/nav/nav.py
vincentpierre: I am changing the dictionary in the for loop. You have to make a copy of the keys to be able to iterate.

habitat-lab/habitat/core/embodied_task.py
aclegg3: Agreed this step size should be configured somewhere rather than locked at 60Hz.
True this is the default, but better that it is configured than a magic number.

aclegg3 (approved):
Overall looks good to me.

PR #1254, opened 2023-04-10
test_sandbox_driver attempt

rpartsey:
## Motivation and Context

This is an attempt to implement SandboxDriver image-based unit test described in this [doc](https://docs.google.com/document/d/1KUYO1jJwsTdT8yAJEmCrdpzrgFCu2yhtZf9lWxdasNE/edit#bookmark=kix.197svbuffvez).

But following instructions from `BatchReplayRendererTest::testIntegration` as suggested in the doc `renderer.render([mutable_image_view])`  fails with `# AssertionError: ESP_CHECK failed: Renderer was not created with a background render thread, cannot do async drawing`.

## Types of changes

## Checklist

### DIFF
--- a/test/test_sandbox_app.py
+++ b/test/test_sandbox_app.py
@@ -1,0 +1,148 @@
+from collections import namedtuple
+from os import path as osp
+
+import magnum as mn
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+from magnum.platform.glfw import Application
+
+import habitat
+import habitat_sim
+from examples.siro_sandbox.sandbox_app import (
+    SandboxDriver,
+    update_habitat_config,
+)
+from habitat.gui.gui_application import GuiApplication
+from habitat.gui.gui_input import GuiInput
+from habitat.gui.replay_gui_app_renderer import ReplayGuiAppRenderer
+
+# import os
+# import sys
+# cwd = os.getcwd()
+# sys.path.insert(0, cwd)
+# print()
+# print(sys.path)
+# print()
+
+
+DEFAULT_APP_WINDOW_WIDTH = 1280
+DEFAULT_APP_WINDOW_HEIGHT = 720
+DEFAULT_POSE_PATH = "data/humanoids/humanoid_data/walking_motion_processed.pkl"
+DEFAULT_CFG = "benchmark/rearrange/rearrange_easy_human_and_fetch.yaml"
+DEFAULT_CFG_OPTS = ["habitat.dataset.split=minival"]
+
+SandboxDriverArgs = namedtuple(
+    "SandboxDriverArgs",
+    [
+        "cfg",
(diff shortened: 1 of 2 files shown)


test/test_sandbox_app.py
```diff
@@ -0,0 +1,148 @@
...
+    glfw_config = Application.Configuration()
+    glfw_config.title = "Sandbox App"
+    glfw_config.size = (args.width, args.height)
+    gui_app_wrapper = GuiApplication(glfw_config, 30)
+    framebuffer_size = gui_app_wrapper.get_framebuffer_size()
+
+    # init SandboxDriver
+    sim_input = GuiInput()
+    driver = SandboxDriver(args, config, gui_input=sim_input)
+    app_renderer = ReplayGuiAppRenderer(
```
eundersander, line 93: It's probably not appropriate to use ReplayGuiAppRenderer for this test, because it assumes it can render to the "default framebuffer". A better option is probably to create your own replay renderer using `create_classic_replay_renderer`.

test/test_sandbox_app.py
```diff
@@ -0,0 +1,148 @@
...
+        dtype=np.uint8,
+    )
+    mutable_image_view = mn.MutableImageView2D(
+        mn.PixelFormat.RGB8_UNORM, framebuffer_size, buffer
+    )
+    # app_renderer._replay_renderer.set_sensor_transform(
+    #     0, app_renderer._sensor_uuid, app_renderer.cam_transform
+    # )
+    # AssertionError: ESP_CHECK failed: Renderer was not created with a background render thread, cannot do async drawing
+    # app_renderer._replay_renderer.render([mutable_image_view])
```
eundersander, line 128: Once you switch to a standalone replay renderer, you may still his this same error. @0mdc should weigh in as he created these renderers and knows about the various `render` methods.

eundersander (commented):
Added some quick comments.

rpartsey:
> Added some quick comments.

Thank you @eundersander , I have tried like that but I faced with the same error.

test/test_sandbox_app.py
0mdc: A few things could be happening here.

1. Make sure that you are using an up-to-date habitat-sim. This error was occurring prior to [this fix](https://github.com/facebookresearch/habitat-sim/pull/2043).
2. `ReplayGuiAppRenderer` is windowed, hence it already has a context. It tries to create a standalone context here. Therefore, it fails. The following solutions could fix this:
    * Creating path so that `ReplayGuiAppRenderer` can be run in headless mode.
    * I can add a python binding to copy the framebuffer into an `MutableImageView`. This already exists in the backend. In this case, you would have to call `replay_renderer.render(framebuffer)`, then something like `replay_renderer.copy_color_into(image_views)`.

PR #1257, opened 2023-04-10
Aks wandb job name

akshararai:
## Motivation and Context

Adding a job name to configs to make it easy to tell the different runs apart.
Adding wandb job names to correspond to checkpoint folders so that we can match runs with checkpoints.


xavierpuigf (approved):
LGTM

PR #1259, opened 2023-04-11
Sandbox tool on-screen text

rpartsey:
## Motivation and Context

- Implemented `TextDrawer` class for on-screen text drawing
- Added default font
- Integrated `TextDrawer` into the SIRo Sandbox Tool 

Example of the on-screen text display:
Screenshot 2023-04-22 at 00 47 42

## How Has This Been Tested
Running Sandbox tool locally.

## Types of changes

- **Development** 

## Checklist


habitat-lab/habitat/gui/replay_gui_app_renderer.py
```diff
@@ -57,11 +66,27 @@ def post_sim_update(self, post_sim_update_dict):
...
-            self._need_render = True
+        if "info" in post_sim_update_dict:
+            # get needed metrics and format text string
+            # to be drawn on the GUI app screen
+            info = post_sim_update_dict["info"]
+            self._text_to_draw = (
+                pformat(
+                    {
+                        "App metrics": info["app"],
+                        "Env metrics": info["env"],
```
eundersander, line 83: This is a bit too specific to put into ReplayGuiAppRenderer. E.g. the assumption that we only want to print app and env metrics as opposed to arbitrary text.

How about just give the Driver direct access to the TextDrawer? Similar to how we give direct access to the Debug Line Render. See my thoughts below on the interface; we have to separate adding/drawing text from actually drawing to the screen.

habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -0,0 +1,94 @@
...
+        mn.gl.Renderer.enable(mn.gl.Renderer.Feature.BLENDING)
+        mn.gl.Renderer.set_blend_function(
+            mn.gl.Renderer.BlendFunction.ONE,
+            mn.gl.Renderer.BlendFunction.ONE_MINUS_SOURCE_ALPHA,
+        )
+        mn.gl.Renderer.set_blend_equation(
+            mn.gl.Renderer.BlendEquation.ADD, mn.gl.Renderer.BlendEquation.ADD
+        )
+
+    def draw_text(self, text_to_draw):
```
eundersander, line 87: Can you add documentation for how to use this function? E.g. what is text_to_draw, how do I specify where text gets printed on-screen, etc.

eundersander: I can think of two possible interfaces that would be nice to expose to user code (Drivers):
(1) `draw_text(str, on_screen_position)`, where `on_screen_position` does something similar to CSS to help position the text, e.g. a combination of left|right and top|bottom alignment plus a pixel offset.
(2) `add_text_line(str)`, where we just let the user code provide multiple lines of text, and then the GuiRenderer decides how to arrange them on-screen.

We want user code (Drivers) to be able to call `draw_text/add_text_line` from anywhere in sim code, at any time. The TextDrawer needs to save these requests internally and then "flush" them later (actually drawing to OpenGL). Again, this is exactly the paradigm used by DebugLineRender.

Open to other ideas here on the interface for specifying/arranging the text!

eundersander: And just a heads up that someday in the future, we will want 3D-positioned text, e.g. `draw_text(str, world_position)`, which will use the current camera and project to screen space to come up with the 2D screen position.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -402,12 +408,20 @@ def depth_to_rgb(obs):
...
+    driver = SandboxDriver(args, config, gui_app_wrapper.get_sim_input())
+
+    # instantiate renderer:
     # note this must be created after GuiApplication due to OpenGL stuff
     app_renderer = ReplayGuiAppRenderer(
         framebuffer_size.x, framebuffer_size.y, args.use_batch_renderer
     )
+    image_drawer = ImageFramebufferDrawer(max_width=1024, max_height=1024)
+    text_drawer = TextDrawer(framebuffer_size)
+    app_renderer.set_image_and_text_drawers(image_drawer, text_drawer)
```
eundersander, line 423: This bootstrap code is getting pretty complicated. We want it to be easy to use GuiApplication and related classes. Maybe ReplayGuiAppRenderer should construct image and text drawers internally in its constructor, with constructor arguments to help construct them.

habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -0,0 +1,94 @@
...
+        framebuffer_size: mn.Vector2i,
+        relative_path_to_font: str = DEFAULT_FONT_PATH,
+        max_display_text_chars: int = MAX_DISPLAY_TEXT_CHARS,
+        text_delta_from_center: float = TEXT_DELTA_FROM_CENTER,
+        display_font_size: float = DISPLAY_FONT_SIZE,
+    ) -> None:
+        self._framebuffer_size = framebuffer_size
+
+        # Load a TrueTypeFont plugin and open the font file
+        self._display_font = text.FontManager().load_and_instantiate(
```
eundersander, line 37: @rpartsey can you post a link to where you borrowed this code from?

@mosra I wanted you to have a look at this Magnum text-drawing code.

eundersander (commented):
I left some comments!

habitat-lab/habitat/gui/fonts/README.txt
```diff
@@ -0,0 +1,100 @@
...
+# draw text
+shader.bind_vector_texture(glyph_cache.texture)
+shader.transformation_projection_matrix = window_text_transform
+shader.color = [1.0, 1.0, 1.0]
+window_text.render(
+    f"""
+Hello World
+    """
+)
+shader.draw(window_text.mesh)
```
mosra, line 100: Is the code snippet here useful for anything? :) Just the license blurb makes sense, the rest not really, and similarly for the original in hab-sim.

Also, given the repo depends on habitat-sim anyway, would it make sense to just inherit / reference the font file from there? Or is that too complicated to do?

mosra (commented):
The text rendering code looks like copied verbatim from the Python viewer in habitat-sim, and [there it was fine](https://github.com/facebookresearch/habitat-sim/pull/1862) so I think this is too.

My only comment / question is about sharing the font file and the unnecessary cruft in the README (which, again, was the case in habitat-sim already).

jimmytyyang (approved):
please fix the pre-commit. thank you!

habitat-lab/habitat/gui/fonts/README.txt
rpartsey: >Is the code snippet here useful for anything? :) Just the license blurb makes sense, the rest not really, and similarly for the original in hab-sim.

Yeah, it is a copy-paste. I'll clean it up.

> Also, given the repo depends on habitat-sim anyway, would it make sense to just inherit / reference the font file from there? Or is that too complicated to do?

I also thought about that. From what I gather, neither python viewer utility is part of habitat-sim package (it lives in the `examples/` folder of the repo) nor `font/` directory  (it lives in `data` folder of the repo). But font is packed in C++ viewer, @mosra if it is possible to somehow extract / reference font from there I can try to do that (but I don't know if it is possible)

rpartsey:
> The text rendering code looks like copied verbatim from the Python viewer in habitat-sim, and [there it was fine](https://github.com/facebookresearch/habitat-sim/pull/1862) so I think this is too.

Exactly!

habitat-lab/habitat/gui/text_drawer.py
rpartsey: All text drawing code borrowed / copy-pasted from [habitat-sim/examples/viewer.py](https://github.com/facebookresearch/habitat-sim/blob/100dc8fb140507a3b63565f69e5da2bcbf91f426/examples/viewer.py#L112).

habitat-lab/habitat/gui/fonts/README.txt
mosra: Ah, if it's not packaged then it doesn't make sense.

Getting it out of the C++ viewer would be technically possible but very hacky, and I don't think you'd want to depend on that executable either (as far as I'm aware, it's getting deprecated in favor of the Python version).

So just keep a copy here. The font is tiny, it doesn't matter that much :)

habitat-lab/habitat/gui/text_drawer.py
rpartsey: Done! I hope my implementation looks good to you @eundersander.
I've combined your suggestions `draw_text(str, on_screen_position)`  and `add_text_line(str)` into a single function with kwargs [add_text](https://github.com/facebookresearch/habitat-lab/blob/09a3274e79c9d6f8a534f68f21111e22f804c407/habitat-lab/habitat/gui/text_drawer.py#L84). If no kwargs are set it is equivalent to `add_text_line(str)` , but with kwargs it can be configured like `draw_text(str, on_screen_position)`.

rpartsey:
> We want user code (Drivers) to be able to call draw_text/add_text_line from anywhere in sim code, at any time. The TextDrawer needs to save these requests internally and then "flush" them later (actually drawing to OpenGL). Again, this is exactly the paradigm used by DebugLineRender.

@mosra could you please take a look if update (multi line) text rendering implementation (suggested by Eric in his comment above) looks good to you?

examples/siro_sandbox/sandbox_app.py
rpartsey: Done

habitat-lab/habitat/gui/replay_gui_app_renderer.py
rpartsey: Done

habitat-lab/habitat/gui/gui_application.py
```diff
@@ -178,12 +179,17 @@ def draw_event(self):
             self._last_draw_event_time = curr_time
             if self._debug_counter >= 100:
                 elapsed = curr_time - self._debug_timer
-                sps = self._debug_counter / elapsed
-                print("sps: {:.1f}".format(sps))
+                self._debug_sps = self._debug_counter / elapsed
+                print("[debug] sps: {:.1f}".format(self._debug_sps))
```
eundersander, line 183: We could just get rid of this console spam since you're printing on-screen, yeah?

habitat-lab/habitat/gui/replay_gui_app_renderer.py
```diff
@@ -57,11 +76,27 @@ def post_sim_update(self, post_sim_update_dict):
...
 
+        if len(keyframes):
+            self._need_render = True
+
         if "debug_images" in post_sim_update_dict:
             self._debug_images = post_sim_update_dict["debug_images"]
 
-        if len(keyframes):
-            self._need_render = True
+        if "info" in post_sim_update_dict:
```
eundersander, line 85: given the driver can already print exactly what it wants where it wants, why is this whole `if "info" in post_sim_update_dict` block needed? Can it be removed?

habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -0,0 +1,119 @@
...
+        ) @ mn.Matrix3.translation(
+            mn.Vector2(self._framebuffer_size) * mn.Vector2(align_x, align_y)
+            + mn.Vector2(text_delta_x, text_delta_y)
+        )
+        self._text_transform_pairs.append((text_to_add, window_text_transform))
+
+    def draw_text(self):
+        """Draws collected text on the screen"""
+        for text_to_draw, transform in self._text_transform_pairs:
+            self._shader.bind_vector_texture(self._glyph_cache.texture)
```
eundersander, line 113: This could be done just once before the loop.

habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -0,0 +1,119 @@
...
+            self._window_text.reserve(self._max_display_text_chars)
```
eundersander, line 114: @mosra what's the right thing to do here? Do one `reserve(max_display_text_chars)` at init time? Or maybe `reserve(len(text_to_draw))` every time here?

Can we also add a runtime check that we're not going over `max_display_text_chars`?

eundersander (commented):
I added some more thoughts.

habitat-lab/habitat/gui/text_drawer.py
mosra: One reserve at init time, yes. The mapping from characters to glyphs isn't always 1:1 (depends on the font) so in general it's not as simple as `len()`.

That being said, the `Text` APIs could use a UX refresh, and I have some updates scheduled there as part of my work on other projects.

mosra: Er, I mean, the code as is here is not ideal perf-wise because it overwrites a single mesh several times over. But that's what you get with the current API, it wasn't made with such "immediate-mode use" in mind. Instead one was supposed to have several dedicated `text.Renderer2D` instances that get updated/rerendered if they actually need an update (for example it probably wouldn't be needed to refresh the values every frame as such rapid changes would be impossible to spot by the human eye anyway).

But as I said, I have some updates scheduled, and making this use case better-supported (such as being able to efficiently fill a larger mesh from smaller text pieces) is among them.

Skylion007 (approved):

habitat-lab/habitat/gui/replay_gui_app_renderer.py
rpartsey: Sorry, my bad. This had to be removed.

PR #1263, opened 2023-04-12
Sandbox tool camera rotation and pitch control

rpartsey:
## Motivation and Context

Added 2 ways to control camera yaw and pitch: 
- WASD keys
- hold Q and move mouse

TODO: decide which one has better UX

## How Has This Been Tested

By launching Sandbox Tool locally.

## Types of changes

- **Refactoring**

## Checklist

### DIFF
--- a/habitat-lab/habitat/gui/gui_application.py
+++ b/habitat-lab/habitat/gui/gui_application.py
@@ -107,8 +107,12 @@ class InputHandlerApplication(Application):
 
     def mouse_move_event(self, event: Application.MouseMoveEvent) -> None:
         mouse_pos = self.get_mouse_position(event.position)
+        relative_mouse_position = self.get_mouse_position(
+            event.relative_position
+        )
         for wrapper in self._gui_inputs:
             wrapper._mouse_position = mouse_pos
+            wrapper._relative_mouse_position = relative_mouse_position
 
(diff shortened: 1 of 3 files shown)


habitat-lab/habitat/gui/gui_application.py
```diff
@@ -107,8 +107,12 @@ def get_mouse_position(
 
     def mouse_move_event(self, event: Application.MouseMoveEvent) -> None:
         mouse_pos = self.get_mouse_position(event.position)
+        relative_mouse_position = self.get_mouse_position(
```
eundersander, line 110: One thing I'm trying to do with this input wrapper is gracefully handle the case where we poll for input from the OS at a different rate from processing it by the sim.

Specifically, (1) what if we get multiple calls to mouse_move_event between calls to sim_update, or (2) what if we do multiple calls to sim_update between calls to mouse_move_event?

In the case of (1), we want the accumulated mouse movement available inside sim_update. For (2), we want the first sim_update to get the mouse movement but the second sim_update should get zero mouse movement.

TLDR make _relative_mouse_position work exactly like _mouse_scroll_offset.

eundersander (approved):
Small request on how we handle relative mouse movement. After that, LGTM!

PR #1264, opened 2023-04-12
[Development] Siro fix of nan issue

jimmytyyang:
## Motivation and Context

 The function returns NaN when asking for starting location of the robot

## How Has This Been Tested

 Running the training script: https://wandb.ai/andrew-colab/hab3_spot_human_63?workspace=user-jimmytyyang

## Checklist


akshararai (approved):
Can you add a print statement here like "Nan start_pos was sampled" so we can get an idea of why this is happening? Then during evaluation, we can try to put a breakpoint when this happens and make sure something funky is not going on.

PR #1276, opened 2023-04-18
SIRo eval sweep

akshararai:
Added `eval_sweep.sh` which takes as input a training sweep folder like `pop_play/2023-04-17/16-19-04`.  It loops over all subdirectories in this folder and uses the training config `config.yaml` to set all eval parameters. This ensures that the train and eval parameters are the same.
It creates a new wandb run with the name `eval_<training_wandb_name>`, saves eval videos to the same training sweep folder. This way we have corresponding training and eval wand runs, as well as checkpoints and videos.

### DIFF
--- a/eval_sweep.sh
+++ b/eval_sweep.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+base_dir="/checkpoint/akshararai/hab3"
+read -p -r "Enter sweep directory name (example: pop_play/2023-04-17/16-19-04):" SWEEP_SUBDIR
+SWEEP_DIR="$base_dir/$SWEEP_SUBDIR"
+echo "$SWEEP_DIR"
+#SWEEP_DIR="/checkpoint/akshararai/hab3/pop_play/2023-04-17/16-19-04/"
+echo "Sweep directories:"
+for dir in "$SWEEP_DIR"/*/; do
+	subdir=$(basename "$dir")
+    echo "$subdir"
+    wandb_name="eval_$SWEEP_SUBDIR"
+    python habitat-baselines/habitat_baselines/run.py -m --config-name config.yaml --config-path "${dir}"/.hydra/ habitat_baselines.eval_ckpt_path_dir="${dir}"/checkpoints/latest.pth habitat_baselines.video_dir="${dir}"/video habitat_baselines.wb.run_name="${wandb_name}" habitat_baselines.evaluate=True habitat_baselines.test_episode_count=20 habitat_baselines.num_environments=16
+done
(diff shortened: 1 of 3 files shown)


habitat-baselines/habitat_baselines/config/experiments_hab3/pop_play_kinematic_oracle.yaml
```diff
@@ -85,7 +85,7 @@ habitat_baselines:
     run_name: ${hydra:job.name}_${now:%Y-%m-%d}_${now:%H-%M-%S}
 
   eval:
-    should_load_ckpt: False
+    should_load_ckpt: True
```
xavierpuigf, line 88: Does this affect training at all?

eval_sweep.sh
```diff
@@ -0,0 +1,15 @@
...
+read -p "Enter sweep directory name (example: pop_play/2023-04-17/16-19-04):" SWEEP_SUBDIR
+SWEEP_DIR="$base_dir/$SWEEP_SUBDIR"
+echo "$SWEEP_DIR"
+#SWEEP_DIR="/checkpoint/akshararai/hab3/pop_play/2023-04-17/16-19-04/"
+echo "Sweep directories:"
+for dir in "$SWEEP_DIR"/*/; do
+	subdir=$(basename "$dir")
+    echo "$subdir"
+    wandb_name="eval_$SWEEP_SUBDIR"
+    python habitat-baselines/habitat_baselines/run.py -m --config-name config.yaml --config-path "${dir}"/.hydra/ habitat_baselines.eval_ckpt_path_dir="${dir}"/checkpoints/latest.pth habitat_baselines.video_dir="${dir}"/video habitat_baselines.wb.run_name="${wandb_name}" habitat_baselines.evaluate=True habitat_baselines.test_episode_count=20 habitat_baselines.num_environments=16
```
xavierpuigf, line 14: I would add here `should_load_ckpt=True` just in case

xavierpuigf (approved):
Left minor comments, but otherwise looks good. Thanks for doing this!

habitat-baselines/habitat_baselines/config/experiments_hab3/pop_play_kinematic_oracle.yaml
akshararai: nope it doesn't

akshararai:
> Is there any case where the dir loop will look at more than one folder?
Yes if you start a sweep, hydra will create folders per combination is the sweep. So if you sweep over 3 seeds, dir/ will iclude 3 folders: 0/ 1/ 2/ that contain the checkpoints of each seed.

PR #1305, opened 2023-05-01
Spot nav

jimmytyyang:
## Motivation and Context

 Spot navigation in FP


jimmytyyang:
Spot robot navigation
(1) Backward action for the oracle skill
(2) Do not need to point towards the object when arriving the grasping point
(3) Collision check for picking/placing locations given robot's navmesh

Example
https://user-images.githubusercontent.com/55121504/235453593-358732d8-74ec-47d4-8041-fb24c16d20f6.mp4

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -267,3 +269,310 @@ def step(self, *args, is_last_action, **kwargs):
...
+                0.0,
+                self._config.spawn_max_dist_to_obj,
+                self._sim,
+                self._config.num_spawn_attempts,
+                1,
+                self.cur_articulated_agent,
+                self._config.navmesh_offset_check,
+            )
+            if np.isnan(start_pos).any():
+                print("start_pos contains NaN @oracle_nav_spot_action.py.")
```
Skylion007, line 365: Should use logger

Skylion007 (approved):

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -431,7 +437,53 @@ def place_robot_at_closest_point(
...
+        # Do transformation to get the location
+        center_pos_list = [
+            sim.articulated_agent.sim_obj.transformation.transform_point(xyz)
+            for xyz in nav_pos_3d
+        ]
+        for center_pos in center_pos_list:
+            # Update the transformation of the agent
+            sim.articulated_agent.base_pos = center_pos
+            # Get the transformation
+            trans = sim.articulated_agent.sim_obj.transformation
```
eundersander, line 464: I think this line and the line above are a bit wasteful/sloppy. Why move the agent `base_pos` at all? It seems like you're using it as a helper to construct the transformation you want. Could you do something like:
```
trans = mn.Matrix4(sim.articulated_agent.sim_obj.transformation)  # make a copy of agent trans
for center_pos in center_pos_list:
    trans.translation = center_pos
    cur_pos = [trans.transform_point(...) ...]
```

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -431,7 +437,53 @@ def place_robot_at_closest_point(
...
+            cur_pos = [trans.transform_point(xyz) for xyz in nav_pos_3d]
+            # Project the height
+            cur_pos = [
+                np.array([xz[0], cache_pos[1], xz[2]]) for xz in cur_pos
+            ]
+            is_navigable = [
+                sim.pathfinder.is_navigable(pos) for pos in cur_pos
```
eundersander, line 471: Again, it would be nice to early out on the first failed `pathfinder.is_navigable`.

eundersander (commented):
I left some comments. Good progress!

habitat-baselines/habitat_baselines/rl/hrl/skills/oracle_nav.py
```diff
@@ -44,10 +44,14 @@ def __init__(
...
         )
-
-        self._oracle_nav_ac_idx, _ = find_action_range(
-            action_space, "oracle_nav_action"
-        )
+        try:
+            self._oracle_nav_ac_idx, _ = find_action_range(
+                action_space, "oracle_nav_action"
+            )
+        except KeyError:
```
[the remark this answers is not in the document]
jimmytyyang: Thank you! I added a fix for this.

habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -326,6 +326,47 @@ class OracleNavActionConfig(ActionConfig):
     num_spawn_attempts: int = 200
 
 
+@dataclass
+class OracleNavSpotActionConfig(ActionConfig):
```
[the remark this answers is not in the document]
jimmytyyang: Thank you! How about "OracleNavWithBackUpActionConfig"?

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -267,3 +269,310 @@ def step(self, *args, is_last_action, **kwargs):
...
+        trans = self._sim.articulated_agent.sim_obj.transformation
+        nav_pos_3d = [
+            np.array([xz[0], 0.0, xz[1]]) for xz in self._config.navmesh_offset
+        ]
+        cur_pos = [trans.transform_point(xyz) for xyz in nav_pos_3d]
+        cur_pos = [
+            np.array([xz[0], self._sim.articulated_agent.base_pos[1], xz[2]])
+            for xz in cur_pos
+        ]
+        is_navigable = [
```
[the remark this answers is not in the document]
jimmytyyang: thanks!

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -267,3 +269,310 @@ def step(self, *args, is_last_action, **kwargs):
                 raise ValueError(
                     "Unrecognized motion type for oracle nav action"
                 )
+
+
+class velocity_control:
```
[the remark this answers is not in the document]
jimmytyyang: thank you!

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -267,3 +269,310 @@ def step(self, *args, is_last_action, **kwargs):
...
+                self._sim.articulated_agent.base_transformation.transform_vector(
+                    forward
+                )
+            )
+            robot_forward = robot_forward[[0, 2]]
+            angle = get_angle(robot_forward, rel_pos)
+            vel = OracleNavAction._compute_turn(
+                rel_pos, self._config.turn_velocity, robot_forward
+            )
+            self._sim = vc.act(self._sim, vel)
```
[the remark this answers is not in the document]
jimmytyyang: I expose the control parameter, thank you!

habitat-lab/habitat/tasks/rearrange/multi_task/pddl_sim_state.py
```diff
@@ -183,8 +183,11 @@ def set_state(
                 agent=agent,
             )
 
-            agent.base_pos = start_pos
-            agent.base_rot = start_rot
+            # We only teleport the agent when it is allowed
+            if sim_info.enable_teleport_agent_pos:
```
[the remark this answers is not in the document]
jimmytyyang: @eundersander  Thank you for the comments. I think this should be fine here. Also, set_state also needs to set the grasping function, so it will not make it a no-op. But @ASzot could also commend on this, thank you!

habitat-lab/habitat/tasks/rearrange/utils.py
jimmytyyang: thanks, this is fixed!

habitat-lab/habitat/tasks/rearrange/utils.py
jimmytyyang: thanks!

jimmytyyang:
@eundersander Hi Eric, thanks for the review. I addressed all the issues, especially for teleporting the robot. Now everything is using trans copy method you suggested. thank you!

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -267,3 +267,299 @@ def step(self, *args, is_last_action, **kwargs):
...
+        )
+        # Get the base transformation
+        base_T = self.cur_articulated_agent.base_transformation
+        # Get the current path
+        curr_path_points = self._path_to_point(final_nav_targ)
+        # Get the robot position
+        robot_pos = np.array(self.cur_articulated_agent.base_pos)
+
+        if curr_path_points is None:
+            raise Exception
```
Skylion007, line 447: This seems like a bug. You probably want to raise an instationed instance not a class. And probably should be a RuntimeError

eundersander (commented):
Another minor comment and let's wait to hear from Andrew about the PDDL bit.

SIRO_README.md
```diff
@@ -27,7 +27,8 @@ FP is actually four distinct pieces: Floorplanner scenes, Amazon-Berkeley object
 1. Download Floorplanner Scenes: [fpss_osmm.zip](https://drive.google.com/file/d/1-utUMfUbbzg_zUE5GcGNdk1UEK6lSbXe/view?usp=sharing)
 2. Download [Amazon and Google object archives](https://drive.google.com/drive/folders/1x6i3sDYheCWoi59lv27ZyPG4Ii2GhEZB)
 3. Download FP episodes: [floorplanner.zip](https://drive.google.com/file/d/1Guxn6v2SC5kAtouwMs1DkBJMK3WW0die/view?usp=sharing)
-4. Extract these into `habitat-lab/data` as follows: 
+3-1. (May 1 2023 update): [floorplanner.zip](https://drive.google.com/file/d/1MqhYVnnry3zHj3oqsqbUqy29aAJXz-C5/view?usp=sharing) We generate 100 episodes (nav-pick-place for two objects) using 18 scenes. This dataset is called `s_minitrain.json.gz`
```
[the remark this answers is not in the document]
jimmytyyang: Thanks! I removed the old version

habitat-baselines/habitat_baselines/config/rearrange/rl_hierarchical_oracle_nav_spot_fp.yaml
```diff
@@ -9,16 +9,32 @@
...
+        # If the agent is allowed to slide if the agent needs to do rotation
+        enable_rotation_check_for_dyn_slide: False
+        # The distance threshold of treating collision
+        collision_threshold: 0.02
+        # The navmesh offset for the action control
+        navmesh_offset: [[0.0, 0.0], [0.25, 0.0], [-0.25, 0.0]]
+        # The navmesh offset for the agent placement
+        navmesh_offset_for_agent_placement: [[0.0, 0.0], [0.25, 0.0], [-0.25, 0.0]]
+        enable_lateral_move: False
+    robot_at_thresh: 3.0
```
ASzot, line 35: Why are you modifying this?

habitat-baselines/habitat_baselines/rl/hrl/hl/fixed_policy.py
```diff
@@ -37,8 +37,10 @@ def _parse_solution_actions(self, solution):
                 hl_action.name,
                 [x.name for x in hl_action.param_values],
             )
-            # Filter out the correct action for the robot.
+            # Filter out the correct action for the robot
             robot_id = "robot_" + self._agent_name.split("_")[1]
+            if robot_id == "robot_agent":
```
ASzot, line 42: What is this for? Why do we have to deal with this special case?

habitat-lab/habitat/tasks/rearrange/multi_task/pddl_sim_state.py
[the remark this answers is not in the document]
ASzot: I agree with Eric that this is very risky and I don't like changing this part of the code that all other systems rely on. I also don't think this needs to be a special case. 

I don't understand why Spot does not want to have the teleport condition in the PDDL? If you don't want Spot to teleport for navigation then you should use the "oracle" nav instead of the "teleport" nav or set `apply_postconds: False` for the navigation skill.

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -431,7 +437,53 @@ def place_robot_at_closest_point(
             f"Object {target_position} is out of bounds but trying to set robot position to {agent_pos}"
         )
 
+    # Get the normal agent position
+    agent_pos = sim.safe_snap_point(target_position)
```
ASzot, line 441: Put all the added logic in the `if` statement for `navmesh_offset`. This code is called very often and there shouldn't be any unnecessary operations if `navmesh_offset` is None.

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -431,7 +437,53 @@ def place_robot_at_closest_point(
...
     desired_angle = get_angle_to_pos(np.array(target_position - agent_pos))
+    # Cache the initial location of the agent
+    cache_pos = sim.articulated_agent.base_pos
+
+    # Make a copy of agent trans
+    trans = mn.Matrix4(sim.articulated_agent.sim_obj.transformation)
+
+    accept_agent_center = []
+
+    if navmesh_offset is not None:
```
ASzot, line 451: Put this in a separate method for cleaner code.

ASzot (changes requested):
You should remove `enable_teleport_agent_pos`.

habitat-lab/habitat/tasks/rearrange/multi_task/pddl_sim_state.py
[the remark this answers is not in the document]
jimmytyyang: @ASzot Thanks for this. The reason for not doing teleporting for Spot is because that Spot would be too closed to the target receptacle and thus prone to have collision.

Example of this 
https://user-images.githubusercontent.com/55121504/236859124-a8fdbc45-ab09-40c1-bd1c-6892b7aeda76.mp4

You can see that Spot navigates to the target receptacle and then was teleported. 

Based on 60 testing episodes, I observed that by removing this post-condition, we can reduce the collision rate from 0.36% to 0.11%, which is over 2 times smaller.

I have been already using `oracle_nav` (`nav_to_obj`). But when I turn off `apply_postconds`, it seems that the robot is not navigated anymore. See the video below

https://user-images.githubusercontent.com/55121504/236860287-a40a6702-dd63-4856-b712-bf81fe006033.mp4

I will work this a bit, thanks!

habitat-baselines/habitat_baselines/config/rearrange/rl_hierarchical_oracle_nav_spot_fp.yaml
jimmytyyang: Spot navigation code requires considering the non-cylinder navmesh. Also changing `robot_at_thresh` allows Spot to navigate in the tight space to halt early without getting too closed to the target location.

habitat-baselines/habitat_baselines/rl/hrl/hl/fixed_policy.py
jimmytyyang: It seems that in the config yaml, the agent is called "robot_agent", but the pddl solution yaml is called "robot_0". Maybe we want to have two solution yamls one for single agent, and the other one for multi-agent?

habitat-lab/habitat/tasks/rearrange/utils.py
jimmytyyang: Thank you, done!

habitat-baselines/habitat_baselines/rl/hrl/hl/fixed_policy.py
ASzot: Yes, I think the better thing would be to have two solution yamls.

jimmytyyang:
> You should remove `enable_teleport_agent_pos`.

Thank you, Andrew! This is done, and I turn off `apply_postcond`.

jimmytyyang:
Support multi-agent setup

ASzot (approved):

PR #1308, opened 2023-05-03
Add gfx-replay capture to SIRo sandbox app.

0mdc:
## Motivation and Context

This provides the sandbox app with the capability to record gfx-replay captures.

When starting the sandbox app with the `--enable-gfx-replay-save` argument, every frame is recorded.
Upon pressing the period key, the gfx-replay file is saved at the location defined in `--gfx-replay-save-path`.

Note that currently, captures have to be done from the start so that all `creation` and `load` keyframes are captured. If we wanted to fix this, we would need to change the recorder so that it can provide the history of loads and creations.

## How Has This Been Tested

Tested locally. Capture validated in Blender.

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -18,7 +18,7 @@ sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)
 
 import argparse
 from functools import wraps
-from typing import Any
+from typing import Any, List
 
 import magnum as mn
 import numpy as np
@@ -102,6 +102,10 @@ class SandboxDriver(GuiAppDriver):
             # limit pith angle to +/- 90 degrees
             self._min_lookat_offset_pitch = -np.pi / 2 + 1e-5
             self._max_lookat_offset_pitch = np.pi / 2 - 1e-5
+
+        self._enable_gfx_replay_save: bool = args.enable_gfx_replay_save
+        self._gfx_replay_save_path: str = args.gfx_replay_save_path
+        self._recording_keyframes: List[str] = []
 
     @property
     def lookat_offset_yaw(self):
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

examples/siro_sandbox/sandbox_app.py
```diff
@@ -334,11 +361,12 @@ def sim_update(self, dt):
...
 
-        post_sim_update_dict[
-            "keyframes"
-        ] = (
+        keyframes = (
             self.get_sim().gfx_replay_manager.write_incremental_saved_keyframes_to_string_array()
         )
+        post_sim_update_dict["keyframes"] = keyframes
+        if self._enable_gfx_replay_save:
+            self._recording_keyframes.append(keyframes[0])
```
eundersander, line 369: How about assert that `keyframes` length is 1 (or better yet, just append all the frames in `keyframes`)?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -457,6 +485,19 @@ def parse_debug_third_person(args, framebuffer_size):
...
         help="If specified, use the specified viewport height for the debug third-person camera",
     )
+    parser.add_argument(
+        "--enable-gfx-replay-save",
+        action="store_true",
+        default=False,
+        help="Save the gfx-replay keyframes to file. Use --gfx-replay-save-path to specify the save location.",
+    )
+    parser.add_argument(
+        "--gfx-replay-save-path",
```
eundersander, line 495: I don't feel strongly, but maybe it would be simpler to enable gfx-replay-save if and only if this save-path is specified? And thus we only need one new arg instead of two.

eundersander (approved):
I left some minor comments but otherwise LGTM!

examples/siro_sandbox/sandbox_app.py
0mdc: I don't feel strongly either. Due to dependencies, I'll merge this into main as-is. We can change the command line arguments as needed.

PR #1312, opened 2023-05-04
Sandbox tool humanoid first-person control

rpartsey:
## Motivation and Context

First-person Mouse-look camera and walk forward/back keyboard controls.

https://user-images.githubusercontent.com/37709869/236301605-42af209d-7482-441d-adb2-bc873880c8c4.mov

## How Has This Been Tested
Running Sandbox tool locally.

## Types of changes

- **Development** 

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -74,25 +74,26 @@ class SandboxDriver(GuiAppDriver):
         self.ctrl_helper.on_environment_reset()
 
         self.cam_zoom_dist = 1.0
+        self._max_zoom_dist = 50.0
+        self._min_zoom_dist = 0.1
+
         self.gui_input = gui_input
 
         self._debug_line_render = None
         self._debug_images = args.debug_images
 
-        self.lookat = None
-        # lookat offset yaw (spin left/right) and pitch (up/down)
-        # to enable camera rotation and pitch control
-        # (computed from previously hardcoded mn.Vector3(0.5, 1, 0.5).normalized())
-        self._lookat_offset_yaw = 0.785
-        self._lookat_offset_pitch = 0.955
-
         self._held_target_obj_idx = None
         self._viz_anim_fraction = 0.0
 
         self.lookat = None
-        self._first_person_mode = False
-        self._first_person_yaw = 0.0
-        self._first_person_pitch = 0.0
+        self._first_person_mode = True
+        if self._first_person_mode:
+            self._lookat_offset_yaw = 0.0
+            self._lookat_offset_pitch = 0.0
+        else:
+            # (computed from previously hardcoded mn.Vector3(0.5, 1, 0.5).normalized())
+            self._lookat_offset_yaw = 0.785
+            self._lookat_offset_pitch = 0.955
 
     @property
     def lookat_offset_yaw(self):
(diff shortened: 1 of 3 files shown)


rpartsey:
Agreed with @eundersander to merge this PR as is. We will come back to FPV enhancements soon.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -460,38 +477,44 @@ def sim_update(self, dt):
...
+            lookat -= mn.Vector3(0, 0.2, 0)
+
         offset = mn.Vector3(
             np.cos(self.lookat_offset_yaw) * np.cos(self.lookat_offset_pitch),
             np.sin(self.lookat_offset_pitch),
             np.sin(self.lookat_offset_yaw) * np.cos(self.lookat_offset_pitch),
         )
+
         cam_transform = mn.Matrix4.look_at(
             lookat + offset.normalized() * self.cam_zoom_dist,
```
eundersander, line 513: Note my comments here: https://cvmlp.slack.com/archives/D03BA68KGHW/p1683043849141029

An FPS camera should have a fixed position, unlike here, where the position changes as the offset changes.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -460,38 +477,44 @@ def sim_update(self, dt):
...
-        if agent_idx is not None:
+        if agent_idx is None:
+            self._free_camera_lookat_control()
+            lookat = self.lookat
+        else:
             art_obj = (
                 self.get_sim().agents_mgr[agent_idx].articulated_agent.sim_obj
             )
             robot_root = art_obj.transformation
             lookat = robot_root.translation + mn.Vector3(0, 1, 0)
```
eundersander, line 499: I feel like the FPS camera is current too high; ideally it should be exactly at human eye height. It feels like it is somewhat above the head right now.

eundersander (commented):
Leaving a review here as promised even though we already merged this.

examples/siro_sandbox/sandbox_app.py
rpartsey: Yes, remember that. Currently, this is a quick hack: if zoom dist is set to a small fixed value this is almost equivalent to rotating fixed camera at the spot. This allows to reuse a lot of code without duplication.

Otherwise, I have to create two separate look-at matrices: one for 3-rd person view where the camera is rotating around the agent and the other one for FPS where the camera is fixed. And all the yaw/pitch controls will have an opposite effect on each camera (increasing yaw/pitch angle will make agent look up/left in on case and down/right in the other), right?

examples/siro_sandbox/sandbox_app.py
rpartsey: During the development I've tried different camera positions. I can mount the camera exactly at the same location as the agent's head depth camera. It will be exactly at human eye height but a bit in front (5-10cm) of agents head. Then if we press walk back key we will see part agents head in the FPV when it will be moving backward. If this sounds good to you I can do that.

examples/siro_sandbox/sandbox_app.py
eundersander: I agree with your assessment. It would be cleaner to eventually have the FPS camera math implemented correctly, instead of your quick hack. But this is low priority.

examples/siro_sandbox/sandbox_app.py
eundersander: Eh, I don't want the head to appear on-screen. We can leave this as-is for now. Maybe we'll look into hiding the humanoid in the main viewport eventually, such that we can place the camera exactly where we want without this issue.

PR #1314, opened 2023-05-05
SIRo Evaluation Fixes

ASzot:
* Bug fixes for SIRo evaluation from @xavierpuigf.
* ZSC script.

### DIFF
--- a/habitat-baselines/habitat_baselines/config/experiments_hab3/eval_zsc_kinematic_oracle.yaml
+++ b/habitat-baselines/habitat_baselines/config/experiments_hab3/eval_zsc_kinematic_oracle.yaml
@@ -1,0 +1,46 @@
+# @package _global_
+
+defaults:
+  - pop_play_kinematic_oracle
+  - /habitat/task/lab_sensors:
+    - multi_agent_all_predicates
+    - all_predicates
+  - override /habitat_baselines/rl/policy@habitat_baselines.rl.policy.agent_1: hab3_planner
+  - _self_
+
+habitat_baselines:
+  evaluate: True
+  rl:
+    agent:
+      num_pool_agents_per_type: [1,1]
+  eval:
+    should_load_ckpt: True
+habitat:
+  gym:
+    obs_keys:
+      # Include all possible inputs. The agents will filter them out.
+      - agent_0_head_depth
+      - agent_0_relative_resting_position
+      - agent_0_obj_start_sensor
+      - agent_0_obj_goal_sensor
+      - agent_0_obj_start_gps_compass
+      - agent_0_obj_goal_gps_compass
+      - agent_0_joint
+      - agent_0_is_holding
+      - agent_0_ee_pos
+      - agent_0_localization_sensor
+      - agent_0_other_agent_gps
+      - agent_0_multi_agent_all_predicates
+      - agent_1_head_depth
+      - agent_1_relative_resting_position
+      - agent_1_obj_start_sensor
+      - agent_1_obj_goal_sensor
+      - agent_1_obj_start_gps_compass
+      - agent_1_obj_goal_gps_compass
+      - agent_1_joint
+      - agent_1_is_holding
+      - agent_1_ee_pos
+      - agent_1_localization_sensor
+      - agent_1_other_agent_gps
+      - agent_1_multi_agent_all_predicates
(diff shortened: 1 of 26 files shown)


habitat-baselines/habitat_baselines/config/habitat_baselines/rl/policy/hab3_planner.yaml
```diff
@@ -15,4 +15,9 @@ hierarchical_policy:
     allow_other_place: False
     max_search_depth: 8
     is_reactive: False
+    # This is the number of non-overlapping steps to sub-sample from each
+    # planned sequence (so the agent only does a part of the overall task).
+    plan_split_len: 4
```
xavierpuigf, line 20: Does this split the plan? Shouldn't we split the goals instead?

habitat-baselines/habitat_baselines/rl/hrl/hl/high_level_policy.py
```diff
@@ -33,6 +33,10 @@ def __init__(
         self._device = None
         self._agent_name = agent_name
 
+    @property
+    def should_load_agent_state(self):
```
xavierpuigf, line 37: I think this will give eval problems, it should not always return False.

habitat-baselines/habitat_baselines/config/habitat_baselines/rl/policy/hab3_planner.yaml
ASzot: Yeah good point. That's a better way to do it. Let me fix the code to have it like that instead.

habitat-baselines/habitat_baselines/rl/hrl/hl/high_level_policy.py
ASzot: I think this is the better way to do it because the NN HL policy will override this to return True.

habitat-baselines/habitat_baselines/rl/hrl/hl/planner_policy.py
```diff
@@ -143,29 +142,41 @@ def _get_all_plans(self, pred_vals):
         for path in paths:
             all_ac_seqs.append([node.action for node in path])
         # Sort by the length of the action sequence
-        return sorted(all_ac_seqs, key=lambda x: len(x))
+        full_plans = sorted(all_ac_seqs, key=lambda x: len(x))
```
Skylion007, line 145:
```suggestion
        full_plans = sorted(all_ac_seqs, key=len)
```

akshararai (approved):
looks good, thank you for waiting for me to review!

habitat-baselines/habitat_baselines/config/experiments_hab3/eval_zsc_kinematic_oracle.yaml
```diff
@@ -0,0 +1,46 @@
+# @package _global_
+
+defaults:
+  - pop_play_kinematic_oracle
+  - /habitat/task/lab_sensors:
+    - multi_agent_all_predicates
+    - all_predicates
```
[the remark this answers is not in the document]
ASzot: Yes, unfortunately we do need both because the planner expects `all_predicates` just based off the key. Maybe in the future we should refactor this.

habitat-baselines/habitat_baselines/rl/multi_agent/bdp_access_mgr.py
```diff
@@ -52,9 +52,23 @@ def _sample_active_idxs(self):
         num_envs = self._agents[0]._num_envs
         device = self._agents[0]._device
 
-        behav_ids = torch.randint(
-            0, self._pop_config.behavior_latent_dim, (num_envs,), device=device
-        )
+        if self._pop_config.force_all_agents:
+            if num_envs != self._pop_config.behavior_latent_dim:
+                raise ValueError(
+                    f"Must have num_envs={num_envs} equal to behavior latent dim={self._pop_config.behavior_latent_dim}"
```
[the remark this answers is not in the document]
ASzot: So we can have 1 latent ID per environment to evaluate against all agents in the training population (otherwise I would have to have some sampling interval to iterate through the behavior IDs during eval).

PR #1324, opened 2023-05-09
Sandbox: fix for humanoid flying-into-the-sky bug

eundersander:
## Motivation and Context

This bug happened if you walked into a wall long enough.

My fix is to rethink how we query the navmesh and how we interpret the result.

I also updated the readme with a tip for enabling `allow_sliding`, which is off in at least some of our configs.

## How Has This Been Tested

## Types of changes

PR into SIRo branch

## Checklist

### DIFF
--- a/examples/siro_sandbox/controllers.py
+++ b/examples/siro_sandbox/controllers.py
@@ -13,7 +13,6 @@ import numpy as np
 import torch
 
 import habitat.gym.gym_wrapper as gym_wrapper
-import habitat_sim
 from habitat.articulated_agent_controllers import HumanoidRearrangeController
 from habitat.gui.gui_input import GuiInput
 from habitat.tasks.rearrange.actions.actions import ArmEEAction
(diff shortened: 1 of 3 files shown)


examples/siro_sandbox/controllers.py
```diff
@@ -423,25 +422,6 @@ def _update_grasp(self, grasp_object_id, drop_pos):
...
-        # https://github.com/facebookresearch/habitat-lab/blob/2b7eaf9266a8916c0851c79043236ff2e55eaca0/habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py#L153
-        articulated_agent = self.get_articulated_agent()
-        trans = articulated_agent.sim_obj.transformation
-        rigid_state = habitat_sim.RigidState(
-            mn.Quaternion.from_matrix(trans.rotation()), trans.translation
-        )
-        target_rigid_state_trans = (
-            self._humanoid_controller.obj_transform_base.translation
-        )
-        end_pos = self._env._sim.step_filter(
```
eundersander, line 437: Here's roughly the problem with how we were calling this:
* Start and end positions are at pelvis height, not feet/navmesh height. This is generally risky.
* Normally, `step_filter`'s return value is a position that has been snapped down to the navmesh.
    * Hence the need to do `end_pos -= articulated_agent.params.base_offset` to get a translation suitable to assign to `self._humanoid_controller.obj_transform_base.translation` (i.e. raise `end_pos` back to pelvis height).
* However, sometimes start or end positions can't be mapped to the navmesh.
    * This happens often in the sandbox app, namely when you try to walk into a wall.
    * In this case, `step_filter` just returns the start pos, which was at pelvis height. So, it is incorrect to apply `articulated_agent.params.base_offset` in that case (`end_pos` gets raised above pelvis height).

rpartsey (approved):
Thanks for this fix and detailed explanation @eundersander! Tested locally works great!

PR #1333, opened 2023-05-11
Sandbox policy-driven agents

rpartsey:
## Motivation and Context

- Implement policy-driven (PDDL planner + oracle skill) agents (Humanoid and Spot).
- Use baselines config instead of benchmark config to setup Sandbox app. Baselines config should contain all the information not only about the benchmark but also about baseline agents that we should use to configure controllers. 
- implement [Revised object-grasping and placing](https://docs.google.com/document/d/1KUYO1jJwsTdT8yAJEmCrdpzrgFCu2yhtZf9lWxdasNE/edit#bookmark=kix.x377g3s361bn)
    - Add object grasp/place proximity threshold CLI argument.
- Remove nav hints to the objects grasped by the other agent.

## How Has This Been Tested

Running Sandbox app locally.

## Types of changes

- **Refactoring**

## Checklist

### DIFF
--- a/examples/siro_sandbox/README.md
+++ b/examples/siro_sandbox/README.md
@@ -11,10 +11,10 @@ This is a 3D interactive GUI app for testing various pieces of SIRo, e.g. rearra
 * When using Floorplanner scenes (see below), the app has very bad runtime perf on older Macbooks (2021 is fine; 2019 is bad).
 * Spot robot stops and doesn't move once it collides with any object (try pressing `M` to reset to a next episode).
 
-## Running HITL eval with a user-controlled humanoid and policy-driven Fetch or Spot
+## Running HITL eval with a user-controlled humanoid and policy-driven Spot
 
 * Make sure you've followed the [SIRo install instructions](../../SIRO_README.md#installation), including grabbing latest habitat-sim `main`.
-* To use Fetch, run:
+<!-- * To use Fetch, run:
 ```
 HABITAT_SIM_LOG=warning MAGNUM_LOG=warning \
 python examples/siro_sandbox/sandbox_app.py \
@@ -24,17 +24,18 @@ python examples/siro_sandbox/sandbox_app.py \
 --cfg benchmark/rearrange/rearrange_easy_human_and_fetch.yaml \
 --cfg-opts habitat.dataset.split=minival \
 --sample-random-baseline-base-vel
-```
+``` -->
 * To use Spot, run:
 ```
 HABITAT_SIM_LOG=warning MAGNUM_LOG=warning \
 python examples/siro_sandbox/sandbox_app.py \
 --disable-inverse-kinematics \
 --never-end \
---gui-controlled-agent-index 0 \
---cfg benchmark/rearrange/rearrange_easy_human_and_spot.yaml \
---cfg-opts habitat.dataset.split=minival \
---sample-random-baseline-base-vel
+--gui-controlled-agent-index 1 \
+--cfg experiments_hab3/pop_play_kinematic_oracle_humanoid_spot_fp.yaml \
+--cfg-opts \
+habitat_baselines.evaluate=True \
+habitat_baselines.num_environments=1
 ```
 * Solo user-controlled humanoid mode, with sliding enabled:
 ```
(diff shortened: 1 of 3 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -883,7 +887,8 @@ def parse_debug_third_person(args, framebuffer_size):
         debug_third_person_height,
     ) = parse_debug_third_person(args, framebuffer_size)
 
-    config = habitat.get_config(args.cfg, args.cfg_opts)
+    config = get_baselines_config(args.cfg, args.cfg_opts)
```
rpartsey, line 890: Use baselines config instead of benchmark config. Baselines config should contain all the information not only about the benchmark but also about baseline agent that we should use to configure controllers.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -65,10 +67,12 @@ def __init__(self, args, config, gui_input):
             config.habitat.simulator.habitat_sim_v0.enable_gfx_replay_save = (
                 True
             )
+            config.habitat.simulator.concur_render = False
```
rpartsey, line 70: Without this I get `ESP_CHECK failed: Renderer was not created with a background render thread, cannot do async drawing` error.

examples/siro_sandbox/README.md
```diff
@@ -43,9 +44,10 @@ python examples/siro_sandbox/sandbox_app.py \
 --disable-inverse-kinematics \
 --gui-controlled-agent-index 0 \
 --never-end \
---cfg benchmark/rearrange/rearrange_easy_human.yaml \
---cfg-opts habitat.dataset.split=minival \
-habitat.simulator.habitat_sim_v0.allow_sliding=True
+--cfg experiments_hab3/single_agent_pddl_planner_kinematic_oracle_humanoid.yaml \
```
eundersander, line 47: It looks like you lost the `habitat.simulator.habitat_sim_v0.allow_sliding=True`. I found this flag valuable for easier movement.

examples/siro_sandbox/sandbox_app.py
eundersander: Seems like a fine workaround for now. We can investigate later if concurrent rendering (background-thread rendering) will help the sandbox app runtime perf.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -194,10 +210,41 @@ def visualize_task(self):
...
+                        if dist_to_obj < self._can_grasp_place_threshold:
+                            self._held_target_obj_idx = None
+                            # object_drop_height = 0.15
+                            drop_pos = (
+                                goal_position
+                                # + mn.Vector3(0, object_drop_height, 0)
+                            )
+
             except ValueError:
                 pass
```
eundersander, line 239: What's the need for this? If a held object is not in `idxs`?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -194,10 +210,41 @@ def visualize_task(self):
...
+                        )
+
+                    assert self._held_target_obj_idx is not None
+                    if self.gui_input.get_key_down(GuiInput.KeyNS.SPACE):
+                        translation = self.get_agent_translation()
+                        dist_to_obj = np.linalg.norm(
+                            goal_position - translation
+                        )
+                        if dist_to_obj < self._can_grasp_place_threshold:
+                            self._held_target_obj_idx = None
```
eundersander, line 231: It's confusing to see this object-drop logic in `visualize_task`. Let's rename the function (`visualize_task_and_set_act_hints`?).

examples/siro_sandbox/sandbox_app.py
```diff
@@ -937,15 +962,40 @@ def parse_debug_third_person(args, framebuffer_size):
...
-        args.gui_controlled_agent_index is None
-        or args.gui_controlled_agent_index >= 0
-        and args.gui_controlled_agent_index
-        < len(config.habitat.simulator.agents)
-    ), (
-        f"--gui-controlled-agent-index argument value ({args.gui_controlled_agent_index}) "
-        f"must be >= 0 and < number of agents ({len(config.habitat.simulator.agents)})"
-    )
+        if args.gui_controlled_agent_index is not None:
+            assert (
```
eundersander, line 966: Best practices for Python IMO is we should only use asserts for cases we *never* expect to happen. For routine errors coming from bad user input, an exception is better. See discussion here: https://stackoverflow.com/questions/944592/best-practice-for-using-assert

An example of how this matters: one can disable asserts as a (minor) runtime perf optimization; your program should still work correctly in that case.

eundersander (approved):
I left some minor comments that I hope you can address, otherwise LGTM!

examples/siro_sandbox/sandbox_app.py
rpartsey: Good catch! Legacy code, can be removed.

PR #1336, opened 2023-05-12
SIRo Train Speed and Hyperparams

ASzot:


habitat-baselines/habitat_baselines/rl/hrl/hierarchical_policy.py
```diff
@@ -190,7 +203,9 @@ def num_recurrent_layers(self):
         if self._high_level_policy.num_recurrent_layers != 0:
             return self._high_level_policy.num_recurrent_layers
         else:
-            return self._skills[0].num_recurrent_layers
+            return self._skills[
```
xavierpuigf, line 206: When is 0 not in _skills? Or it is for generality?

habitat-lab/habitat/config/habitat/task/rearrange/multi_agent_tidy_house_base.yaml
```diff
@@ -32,29 +32,45 @@ type: RearrangeCompositeTask-v0
...
+
+  # agent 1 navigation settings
   agent_1_base_velocity:
     agent_index : 1
+    lin_speed: 40.0
+    ang_speed: 20.0
+  agent_1_oracle_nav_action:
+    agent_index : 1
+    spawn_max_dist_to_obj: -1.0
+    lin_speed: 40.0
```
xavierpuigf, line 63: What is the difference between specifying the lin_speed here vs in base_velocity?

habitat-baselines/habitat_baselines/rl/hrl/hierarchical_policy.py
ASzot: `0` would actually still work here. I originally tried changing the index to based on PyTorch tensors, but even without that change, I think this is still best for generality (since we just want to get any skill here).

akshararai (approved):
I wonder if there are more self.config... uses that we can remove and speed up training. Thanks for fixing these!

PR #1338, opened 2023-05-14
human speed controllable

xavierpuigf:
## Motivation and Context
Adds the ability to control the humanoid speed with a config.

## How Has This Been Tested

Ran policies with planner

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
+++ b/habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
@@ -96,10 +96,11 @@ class HumanoidRearrangeController:
             walk_data["stop_pose"]["joints"].reshape(-1),
             mn.Matrix4(walk_data["stop_pose"]["transform"]),
         )
-        self.draw_fps = draw_fps
         self.dist_per_step_size = (
             self.walk_motion.displacement[-1] / self.walk_motion.num_poses
         )
+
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -96,11 +96,12 @@ def __init__(
             walk_data["stop_pose"]["joints"].reshape(-1),
             mn.Matrix4(walk_data["stop_pose"]["transform"]),
         )
-        self.draw_fps = draw_fps
         self.dist_per_step_size = (
             self.walk_motion.displacement[-1] / self.walk_motion.num_poses
         )
 
+        self.draw_fps = draw_fps
```
akshararai, line 103: I think its confusing to define draw_fps 2 ways - through lin_speed and through direct parameter. I would remove one of the two.

akshararai (approved):
remove draw_fps from function inputs, and looks good to me!

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
xavierpuigf: Should I set some default value in the init function for draw_fps? What happens if for some reason, we don't call that funcion (for instance in sandbox)

ASzot (approved):

PR #1341, opened 2023-05-17
SIRo: Human-in-the-loop tutorial

0mdc:
## Motivation and Context

This changeset adds an intro sequence ("tutorial") to the sandbox tool to introduce the user to the scene and goals in a HITL context.

Polishing of the animation sequence will be done in subsequent PRs, if needed.

https://github.com/facebookresearch/habitat-lab/assets/110583667/7b267da0-9a3b-4448-af2b-369eff786fc3

## How Has This Been Tested

Tested in the sandbox tool.


examples/siro_sandbox/sandbox_app.py
```diff
@@ -57,6 +60,67 @@ def wrapper(*args, **kwds):
     return wrapper
 
 
+class SandboxState(Enum):
+    SIMULATION = 1
```
eundersander, line 64: Nit: how about a better name than simulation? I guess it's the phase where the user can control the humanoid to complete the task.

Btw, I understand you are not stepping the simulation during the tutorial, hence the motivation to call this state "simulation". My point is really that we'll likely have other states in the future besides tutorial and this one.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -636,35 +711,134 @@ def sim_update(self, dt):
...
-            lookat = self.lookat
-        else:
-            art_obj = (
-                self.get_sim().agents_mgr[agent_idx].articulated_agent.sim_obj
+        lookat = self._create_camera_lookat()
+        self.cam_transform = mn.Matrix4.look_at(
+            lookat[0], lookat[1], mn.Vector3(0, 1, 0)
+        )
+
+    def _generate_tutorial(self) -> List[TutorialStage]:
```
eundersander, line 719: This looks like a lot of logic. Can we move this to a separate helper class and file? `SandboxDriver` is becoming bloated.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -698,7 +872,8 @@ def depth_to_rgb(obs):
             np.flipud(image) for image in debug_images
         ]
 
-        self._update_text()
+        if self._sandbox_state == SandboxState.SIMULATION:
+            self._update_text()
```
eundersander, line 876: I think we'll need help text during the tutorial in the future. Let's refactor this so that it's easy for us to add tutorial-specific help text in the future.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -728,6 +903,69 @@ def evaluate_cubic_bezier(ctrl_pts, t):
     return result
 
 
+def _ease_fn_in_out_quat(t: float):
```
eundersander, line 906: Move this math helper outside this file. Let's avoid bloating `SandboxDriver` further.

eundersander (commented):
Good progress! Let's discuss/address my comments below before merging. Also! Can you attach a video to the PR description?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -121,6 +128,22 @@ def __init__(self, args, config, gui_input):
 
         self._cursor_style = None
 
+        self._sandbox_state = (
```
rpartsey, line 131: As far as I remember, we agreed to play the tutorial at the beginning of each episode (see Eric's message [here](https://cvmlp.slack.com/archives/C0460NTKM4G/p1683922450164749?thread_ts=1683909138.185399&cid=C0460NTKM4G)). CC: @eundersander.

Thus, we need to have `self._show_tutorial = args.show_tutorial` attribute too. To know if we are in the "show tutorial" mode. Because, right now the app switches to `SandboxState.CONTROLLING_AGENT` at the end of the tutorial and tutorial doesn't start at the episode reset (pressing M key).

rpartsey (approved):
Left some comments. LGTM!
Note, that you have merge conflicts.

PR #1342, opened 2023-05-17
SIRo sandbox: camera polish; object-placing polish; fix for EGL_BAD_ACCESS error

eundersander:
## Motivation and Context

* camera polish
* object-placing polish: preserve the original upright orientation of the object on place
* fix for EGL_BAD_ACCESS error on some platforms

## How Has This Been Tested

## Types of changes

PR into SIRo

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -13,8 +13,6 @@ import ctypes
 # must call this before importing habitat or magnum! avoids EGL_BAD_ACCESS error on some platforms
 import sys
 
-from habitat_baselines.config.default import get_config as get_baselines_config
-
 flags = sys.getdlopenflags()
 sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)
 
@@ -40,6 +38,7 @@ from habitat.gui.gui_application import GuiAppDriver, GuiApplication
 from habitat.gui.gui_input import GuiInput
 from habitat.gui.replay_gui_app_renderer import ReplayGuiAppRenderer
 from habitat.gui.text_drawer import TextOnScreenAlignment
+from habitat_baselines.config.default import get_config as get_baselines_config
 
 # Please reach out to the paper authors to obtain this file
 DEFAULT_POSE_PATH = "data/humanoids/humanoid_data/walking_motion_processed.pkl"
@@ -113,7 +112,9 @@ class SandboxDriver(GuiAppDriver):
         self._first_person_mode = args.first_person_mode
         if self._first_person_mode:
             self._lookat_offset_yaw = 0.0
-            self._lookat_offset_pitch = 0.0
+            self._lookat_offset_pitch = float(
+                mn.Rad(mn.Deg(20.0))
+            )  # look slightly down
             self._min_lookat_offset_pitch = (
                 -max(min(np.radians(args.max_look_up_angle), np.pi / 2), 0)
                 + 1e-5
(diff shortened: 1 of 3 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -40,6 +38,7 @@
 from habitat.gui.gui_input import GuiInput
 from habitat.gui.replay_gui_app_renderer import ReplayGuiAppRenderer
 from habitat.gui.text_drawer import TextOnScreenAlignment
+from habitat_baselines.config.default import get_config as get_baselines_config
```
eundersander, line 41: Moving this fixed EGL_BAD_ACCESS error

examples/siro_sandbox/sandbox_app.py
```diff
@@ -399,12 +400,12 @@ def _camera_pitch_and_yaw_wasd_control(self):
             self._lookat_offset_yaw += cam_rot_angle
 
     def _camera_pitch_and_yaw_mouse_control(self):
-        enable_mouse_control = self.gui_input.get_key(GuiInput.KeyNS.R) and (
-            (
-                self._first_person_mode
-                and self._cursor_style == Application.Cursor.HIDDEN_LOCKED
-            )
-            or (not self._first_person_mode)
+        enable_mouse_control = (
```
eundersander, line 403: For first-person mode, if the user left-clicks to hide the cursor, it makes sense to me to just always do mouse-look, without having to hold the R key.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -538,14 +539,12 @@ def _save_recorded_keyframes_to_file(self):
     def _update_cursor_style(self, post_sim_update_dict):
         do_update_cursor = False
         if self._cursor_style is None:
-            if self._first_person_mode:
-                self._cursor_style = Application.Cursor.HIDDEN_LOCKED
-            else:
-                self._cursor_style = Application.Cursor.ARROW
+            self._cursor_style = Application.Cursor.ARROW
```
eundersander, line 542: Defaulting to arrow cursor for first-person mode. This change is essentially Ruta's recent request: that we disable mouse-look.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -859,7 +859,7 @@ def parse_debug_third_person(args, framebuffer_size):
     )
     parser.add_argument(
         "--can-grasp-place-threshold",
-        default=1.0,
+        default=1.2,
```
eundersander, line 862: I noticed an episode where I was unable to get close enough to the fridge to place my object. Ruta can test and tune this further in the future.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -538,14 +539,12 @@ def _save_recorded_keyframes_to_file(self):
...
-            else:
-                self._cursor_style = Application.Cursor.ARROW
+            self._cursor_style = Application.Cursor.ARROW
             do_update_cursor = True
         else:
-            if self._first_person_mode and self.gui_input.get_mouse_button(
-                GuiInput.MouseNS.LEFT
+            if (
+                self._first_person_mode
+                and self.gui_input.get_mouse_button_down(GuiInput.MouseNS.LEFT)
```
eundersander, line 547: Note change to use `get_mouse_button_down`. This is only True on the frame you press down the button. `get_mouse_button` is True every frame as long as you keep holding the button, which we don't want here.

examples/siro_sandbox/sandbox_app.py
rpartsey: Yes, this happened to me too. 

Also, if the can grasp radius is large there may be the cases when the objects can be picked up through the walls or obstacles. Unfortunately, I didn't come up with the idea how that could be addressed.

rpartsey (approved):
Tested locally, works great!

PR #1360, opened 2023-05-19
fixing pop_play_kinematic_oracle_humanoid_spot.yaml on SIRo

akshararai:
## Motivation and Context
PR to fix the humanoid_spot config file


habitat-baselines/habitat_baselines/config/experiments_hab3/pop_play_kinematic_oracle_humanoid_spot.yaml
```diff
@@ -60,6 +64,7 @@ habitat:
       - agent_0_is_holding
       - agent_0_ee_pos
       - agent_0_localization_sensor
+      - agent_0_other_agent_gps
       - agent_0_has_finished_oracle_nav
```
jimmytyyang, line 68: the conclusion is not to use the depth image of Spot?

akshararai: We will use depth image, but also the other agent gps. I think we need both for social rearrangement right now.

jimmytyyang (approved):

PR #1362, opened 2023-05-20
Si ro fp fixes 1

akshararai:
Merging FP fixes branches


akshararai (commented):
merging changes for num_objects

PR #1365, opened 2023-05-21
[Bug fix]: use default_factory in data default structured configs when creating data classes to support Python 3.11

villekuosmanen:
## Motivation and Context

I was trying to set up Habitat-lab for Python 3.11 and ran into this problem:

```
ValueError: mutable default <class 'habitat.config.default_structured_configs.IteratorOptionsConfig'> for field iterator_options is not allowed: use default_factory
```

This PR fixes all cases where a mutable `Config` class is used inside another `dataclass`'s variables by adding a `field(default_factory=...)` call to create a new instance.

Not sure if Python 3.11 is supported yet, but one day it will be and this fix will be needed at that point :)

## How Has This Been Tested

Before the fix:
- Use Python 3.11
- Run the example from README: `python examples/example.py`
- Note the ValueError from earlier.

After the fix, the example runs without errors.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/config/default_structured_configs.py
+++ b/habitat-lab/habitat/config/default_structured_configs.py
@@ -110,7 +110,7 @@ class EnvironmentConfig(HabitatBaseConfig):
     """
     max_episode_steps: int = 1000
     max_episode_seconds: int = 10000000
-    iterator_options: IteratorOptionsConfig = IteratorOptionsConfig()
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

Bluefin-Tuna:
Is there a reason why this change wasn't merged?

PR #1367, opened 2023-05-22
add a flag to overfit

xavierpuigf:
## Motivation and Context

Adds an overfitting flag. Simply add at the end of your run:

```+habitat.simulator.overfit=True```

It will use always the same scene and the same starting position.

## Types of changes

## Checklist

### CODE REVIEW

File: habitat-lab/habitat/tasks/rearrange/rearrange_task.py [hunk 1]
```diff
@@ -187,9 +187,16 @@ def reset(self, episode: Episode, fetch_observations: bool = True):
             self._is_episode_active = True
 
             if self._should_place_articulated_agent:
+                
+                habitat_config = self._sim.habitat_config
```
ASzot, line 191: Don't reference Hydra config in the reset. Instead store as a property in the __init__.

ASzot (approved):

PR #1371, opened 2023-05-23
[Bug Fix] change the snap point to be attached the largest island id

jimmytyyang:
## Motivation and Context

 This PR ensures that the snap point of the object is always in the largest island. Previously it was checked by computing the largest island radius, which is not stable. 

This is the joint work with Alex!

## How Has This Been Tested

 Run the script, and it generates the episode.

## Checklist

- [] I have updated the documentation if required.

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
+++ b/habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
@@ -74,6 +74,7 @@ class ObjectSampler:
         self.sample_region_ratio = sample_region_ratio
         self.nav_to_min_distance = nav_to_min_distance
         self.set_num_samples()
+        self.largest_island_id = -1
         # More possible parameters of note:
         # - surface vs volume
         # - apply physics stabilization: none, dynamic, projection
@@ -87,6 +88,7 @@ class ObjectSampler:
         self.receptacle_candidates = None
         # number of objects in the range should be reset each time
         self.set_num_samples()
+        self.largest_island_id = -1
 
     def sample_receptacle(
         self,
@@ -240,10 +242,14 @@ class ObjectSampler:
 
         # Note: we cache the largest island ID to reject samples which are primarily accessible from disconnected navmesh regions.
         # This assumption limits sampling to the largest navigable component of any scene.
-        island_areas = list(
-            map(sim.pathfinder.island_area, range(sim.pathfinder.num_islands))
-        )
-        self.largest_island_id = island_areas.index(max(island_areas))
+        if self.largest_island_id == -1:
+            island_areas = list(
+                map(
+                    sim.pathfinder.island_area,
+                    range(sim.pathfinder.num_islands),
+                )
+            )
+            self.largest_island_id = island_areas.index(max(island_areas))
 
         while num_placement_tries < self.max_placement_attempts:
             num_placement_tries += 1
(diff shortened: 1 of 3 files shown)


habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -237,9 +237,13 @@ def sample_placement(
...
             sim.pathfinder.build_navmesh_vertices(), axis=0
         )
-        # Note: we cache the largest island to reject samples which are primarily accessible from disconnected navmesh regions. This assumption limits sampling to the largest navigable component of any scene.
-        self.largest_island_size = max(
-            [sim.pathfinder.island_radius(p) for p in navmesh_vertices]
+        # Note: we cache the largest island ID to reject samples which are primarily accessible from disconnected navmesh regions.
+        # This assumption limits sampling to the largest navigable component of any scene.
+        island_size = [
+            sim.pathfinder.island_radius(p) for p in navmesh_vertices
+        ]
```
Skylion007, line 244:
```suggestion
        island_size = list(map(sim.pathfinder.island_radius, navmesh_vertices))
```
Nit is that this might be marginally faster.

Skylion007 (approved):

aclegg3 (commented):
API suggestion. This should be faster and more accurately capture the largest island.

habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -237,13 +237,13 @@ def sample_placement(
...
-        # Note: we cache the largest island to reject samples which are primarily accessible from disconnected navmesh regions. This assumption limits sampling to the largest navigable component of any scene.
-        self.largest_island_size = max(
-            [sim.pathfinder.island_radius(p) for p in navmesh_vertices]
+
+        # Note: we cache the largest island ID to reject samples which are primarily accessible from disconnected navmesh regions.
+        # This assumption limits sampling to the largest navigable component of any scene.
+        island_areas = list(
+            map(sim.pathfinder.island_area, range(sim.pathfinder.num_islands))
         )
+        self.largest_island_id = island_areas.index(max(island_areas))
```
aclegg3, line 246: Seems wasteful to compute this every time we sample an object. Instead, we should check if `self.largest_island_id` is unset and compute it once, then cache the result until `reset` is called.

aclegg3 (commented):
Looking good. Small efficiency idea.

jimmytyyang:
@aclegg3 Do you mind if you take a one more look? Thank you! :D

aclegg3 (commented):
Looks good to me overall.

My only remaining consideration is that the largest island may not always be the correct one. For example if the scene has a large pool/yard then that may be a larger navmesh than the interior.

Possible improvements:
1. Make this configurable. Allow users to turn off this check (maybe already done?)
2. Check if the navmesh is outdoors (raycast upwards does not hit a ceiling). This won't work for all scenes, but should work for some.

We should have a way to validate this logic for any scenes we want to use. Maybe a script to highlight the largest navmesh island in a topdown view or map.

habitat-lab/habitat/datasets/rearrange/samplers/object_sampler.py
```diff
@@ -344,24 +350,23 @@ def _is_accessible(
...
         Note that this might not catch all edge cases since the distance is
         based on Euclidean distance. The nearest navigable point may be
         separated from the object by an obstacle.
         """
         if self.nav_to_min_distance == -1:
             return True
-        snapped = sim.pathfinder.snap_point(obj.translation)
-        island_radius: float = sim.pathfinder.island_radius(snapped)
+
+        snapped = sim.pathfinder.snap_point(
```
aclegg3, line 363: Minor note: this will be NaN if the snap fails. The distance checking logic will return False with comparison to NaN, so it works out.

jimmytyyang: Yes, I also noticed this. thank you, alex!

aclegg3 (approved):
LGTM

PR #1377, opened 2023-05-25
Handle episode over or task complete

rpartsey:
## Motivation and Context

Handle episode over or task complete: freeze simulation, show appropriate message about task ending/failing.

## How Has This Been Tested
Running Sandbox App locally (MacBook Pro 2021).

## Types of changes

- **Refactoring**
- **Development** 

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -176,7 +176,7 @@ class SandboxDriver(GuiAppDriver):
         """
         return not (self.env.episode_over or self._env_task_complete)
 
-    def _compute_action_and_step_env(self):
+    def _check_compute_action_and_step_env(self):
         # step env if episode is active
         # otherwise pause simulation (don't do anything)
         if self._env_episode_active():
@@ -785,7 +785,7 @@ class SandboxDriver(GuiAppDriver):
             self._viz_anim_fraction + dt * viz_anim_speed
         ) % 1.0
 
-        self._compute_action_and_step_env()
+        self._check_compute_action_and_step_env()
 
         if self.gui_input.mouse_scroll_offset != 0:
             zoom_sensitivity = 0.07
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -147,10 +147,19 @@ class SandboxDriver(GuiAppDriver):
         # self._target_obj_ids = None
         # self._goal_positions = None
 
+        self.env_reset()
+        self._prev_post_sim_update_dict = {}
+
+    def env_reset(self):
+        self._obs = self.env.reset()
+        self._metrics = self.env.get_metrics()
         self.on_environment_reset()
 
+    def env_step(self, action):
+        self._obs = self.env.step(action)
+        self._metrics = self.env.get_metrics()
+
     def on_environment_reset(self):
-        self.obs = self.env.reset()
         self.ctrl_helper.on_environment_reset()
         self._held_target_obj_idx = None
         self._num_remaining_objects = None  # resting, not at goal location yet
@@ -162,6 +171,13 @@ class SandboxDriver(GuiAppDriver):
             sim._scene_obj_ids[temp_id] for temp_id in temp_ids
         ]
         self._goal_positions = [mn.Vector3(pos) for pos in goal_positions_np]
+
+    @property
+    def task_complete(self):
+        config = self.env._config
+        success_measure_name = config.task.success_measure
+        episode_success = self._metrics[success_measure_name]
+        return config.task.end_on_success and episode_success
 
     @property
     def lookat_offset_yaw(self):
@@ -249,7 +265,7 @@ class SandboxDriver(GuiAppDriver):
             can_place_position[1] = self.get_agent_feet_height()
             self._debug_line_render.draw_circle(
                 can_place_position,
-                1,
+                self._can_grasp_place_threshold,
                 mn.Color3(255 / 255, 255 / 255, 0),
                 24,
             )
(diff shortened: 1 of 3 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -163,6 +172,13 @@ def on_environment_reset(self):
         ]
         self._goal_positions = [mn.Vector3(pos) for pos in goal_positions_np]
 
+    @property
+    def task_complete(self):
+        config = self.env._config
```
eundersander, line 177: can you cache `success_measure_name` on startup? Accessing `config` is apparently super-slow and should generally be avoided per step.

habitat-lab/habitat/gui/gui_application.py
```diff
@@ -150,7 +150,7 @@ def _post_sim_update(self, post_sim_update_dict):
         if "application_cursor" in post_sim_update_dict:
             self.cursor = post_sim_update_dict["application_cursor"]
 
-        if "application_exit" in post_sim_update_dict:
+        if post_sim_update_dict.get("should_exit", False):
```
eundersander, line 153: I prefer `application_exit`. For messages that specifically go to `GuiApplication`, I was trying to make that clear by using the `application` prefix in the name. If you feel strongly, you could rename to `application_should_exit`.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -623,56 +639,65 @@ def get_grasp_release_controls_text():
...
-            controls_str += "W, S: walk\n"
-            controls_str += "Scroll: zoom\n"
-            controls_str += get_grasp_release_controls_text()
-        else:
-            controls_str += "Left-click + drag: rotate camera\n"
-            controls_str += "A, D: turn camera\n"
-            controls_str += "W, S: pan camera\n"
-            controls_str += "O, P: raise or lower camera\n"
-            controls_str += "Scroll: zoom\n"
+        if not (self.env.episode_over or self.task_complete):
```
eundersander, line 642: I don't understand the difference between `episode_over` and `task_complete`. Are you getting at whether the task was successful versus failed?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -800,6 +835,10 @@ def depth_to_rgb(obs):
 
         self._update_help_text()
 
+        # store post_sim_update_dict for frozen sim mode
+        # when episode is over or when task is completed
+        self._prev_post_sim_update_dict = post_sim_update_dict
```
eundersander, line 840: This feels like a hack. Why is this necessary?

eundersander (commented):
Let's discuss my comments and then give me another chance to review before merging.

habitat-lab/habitat/gui/gui_application.py
eundersander: Also how is your version more Pythonic or cleaner (doing a `get` versus `in`)? It seems less Pythonic.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -709,11 +748,10 @@ def sim_update(self, dt):
...
             )
 
-        action = self.ctrl_helper.update(self.obs)
+        action = self.ctrl_helper.update(self._obs)
+        self.env_step(action)
 
-        self.obs = self.env.step(action)
-
-        post_sim_update_dict = {}
+        post_sim_update_dict = {"should_exit": should_exit}
```
eundersander, line 754: I don't see a reason to send the message at all if the value here would be False.

examples/siro_sandbox/sandbox_app.py
eundersander: At a high level, we should get away from the idea that, because the simulation is paused, we're necessarily going to render the same thing or otherwise "re-use" the `update_dict`. That is not the case. E.g. I can imagine implementing a camera that kinda circles the agent during this time (even though the sim is paused).

eundersander: One thing that probably needs to change is this logic in `replay_gui_app_renderer.py`:
```
        if len(keyframes):
            self._need_render = True
```

When the sim is paused, we may still need to re-render. Again, the example of the camera moving.

habitat-lab/habitat/gui/gui_application.py
rpartsey: > I prefer application_exit. For messages that specifically go to GuiApplication

👍 , I'll revert this key change, makes sense

> Also how is your version more Pythonic or cleaner (doing a get versus in)? It seems less Pythonic.

Earlier `application_exit` was added to post_sim_update_dict only if application exit was initiated (ESC pressed). After that,  exit app condition was evaluated by checking if `"application_exit"` key is present in `post_sim_update_dict`. This was, `application_exit` key's value was never used. In this implementation I check `application_exit` key's value that's why I used `get` method. It seems to me that it is better to have `post_sim_update_dict` with fixed keys instead of variable. For example, if we ever use `application_exit` key then app behaviour should depend on its value not key being present in the dict or not. But I don't feel strongly about this, can undo this change if you wish.

examples/siro_sandbox/sandbox_app.py
rpartsey: Yes, episode is over when any of the constraints is violated and task is complete when the `composite_success` measure is True.

examples/siro_sandbox/sandbox_app.py
rpartsey: Oh, got you! I misunderstood the task. Fixed!

examples/siro_sandbox/sandbox_app.py
```diff
@@ -147,19 +147,33 @@ def __init__(self, args, config, gui_input):
...
-    def on_environment_reset(self):
+    def _env_episode_active(self) -> bool:
+        """
+        Returns True if current episode is active:
+        1) not self.env.episode_over - none of the constraints is violated, or
+        2) not self._env_task_complete - success measure value is not True
+        """
+        return not (self.env.episode_over or self._env_task_complete)
+
+    def _compute_action_and_step_env(self):
```
eundersander, line 172: Nit: when a function like this maybe or maybe won't do something, I prefer `try_` or `check_` in the name. I.e. `_check_compute_action_and_step_env`.

eundersander (approved):
I left a comment but otherwise LGTM!

PR #1378, opened 2023-05-25
Load navmesh

jimmytyyang:
## Motivation and Context

 This is to fix the issue of navmesh was not properly loaded. Also, update Spot robot's radius and height.

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import os.path as osp
 import time
 from collections import defaultdict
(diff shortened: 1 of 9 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -408,7 +409,23 @@ def _load_navmesh(self, ep_info):
...
+            navmesh_settings.agent_radius = agent_config.radius
+            navmesh_settings.agent_height = agent_config.height
+            navmesh_settings.agent_max_climb = 0.01
+            self.recompute_navmesh(
+                self.pathfinder,
+                navmesh_settings,
+                include_static_objects=True,
+            )
+            os.makedirs(osp.dirname(navmesh_path), exist_ok=True)
+            self.pathfinder.save_nav_mesh(navmesh_path)
```
akshararai, line 428: call `self.pathfinder.load_nav_mesh(navmesh_path) ` here to load the navmesh?

akshararai (approved):
do we need to load the navmesh after saving it?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
jimmytyyang: Yes, it is.

jimmytyyang:
Thanks Akshara! The logic here is that if the scene is different from the previous episode, then we load the navmesh.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -408,7 +409,23 @@ def _load_navmesh(self, ep_info):
...
+        load_success = self.pathfinder.load_nav_mesh(navmesh_path)
+
+        # If we cannot load the navmesh, recompute the navmesh and cache it.
+        if not load_success:
+            navmesh_settings = NavMeshSettings()
+            navmesh_settings.set_defaults()
+            agent_config = self.get_agent(0).agent_config
+            navmesh_settings.agent_radius = agent_config.radius
+            navmesh_settings.agent_height = agent_config.height
+            navmesh_settings.agent_max_climb = 0.01
```
ASzot, line 421: Why not also keep this from the agent config?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -408,7 +409,23 @@ def _load_navmesh(self, ep_info):
             base_dir = osp.join(*ep_info.scene_id.split("/")[:2])
 
         navmesh_path = osp.join(base_dir, "navmeshes", scene_name + ".navmesh")
-        self.pathfinder.load_nav_mesh(navmesh_path)
+        load_success = self.pathfinder.load_nav_mesh(navmesh_path)
```
ASzot, line 412: Can you just check if the path exists instead? There might some overhead with calling this function?

ASzot (approved):

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
jimmytyyang: thanks, that makes sense!

PR #1379, opened 2023-05-25
SIRo sandbox: add episode recording

eundersander:
## Motivation and Context

When running the sandbox app, use `--save-filepath-base my_session`. When the user presses `M` to reset the env, the first episode will be saved as `my_session.0.json.gz` and `my_session.0.pkl.gz`. These files contain mostly-identical data; we save both so that developers have two choices for how to consume the data later. After pressing `M` again, the second episode will be saved as `my_session.1.json.gz`, etc. For an example of consuming this data, see `test_episode_save_files.py` .

We should also save episode gfx-replay files. This will be a separate PR.

This PR also includes some unrelated polish on help text and nav hints.

## How Has This Been Tested

Local testing on Fedora laptop

## Types of changes
PR into SIRo branch

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -332,6 +332,16 @@ class SandboxDriver(GuiAppDriver):
         object_id = self._target_obj_ids[target_obj_idx]
         return rom.get_object_by_id(object_id).translation
 
+    def get_target_object_positions(self):
+        sim = self.get_sim()
+        rom = sim.get_rigid_object_manager()
+        return np.array(
+            [
+                rom.get_object_by_id(obj_id).translation
+                for obj_id in self._target_obj_ids
+            ]
+        )
+
     def _update_grasping_and_set_act_hints(self):
         if self.is_free_camera_mode():
             return None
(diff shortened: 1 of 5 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -70,6 +77,17 @@ def get_pretty_object_name_from_handle(obj_handle_str):
...
     def __init__(self, args, config, gui_input):
+        self._dataset = config.habitat.dataset
+        self._num_recorded_episodes = 0
+
+        if args.save_filepath_base:
+            self._step_recorder = StepRecorder()
+            self._save_filepath_base = args.save_filepath_base
+            self._episode_recorder_dict = {}
+        else:
+            self._step_recorder = NullRecorder()
```
eundersander, line 88: The NullRecorder makes it easier to write recording logic in controllers.py (don't have to check for null recorder).

examples/siro_sandbox/sandbox_app.py
```diff
@@ -140,16 +160,49 @@ def __init__(self, args, config, gui_input):
...
-        self.on_environment_reset()
+    def start_episode_recorder(self):
+        assert self._save_filepath_base and self._step_recorder
+        ep_dict: Any = dict()
+        ep_dict["start_time"] = datetime.now()
+        ep_dict["dataset"] = self._dataset
+        ep_dict["episode_id"] = self.env.current_episode.episode_id
 
-    def on_environment_reset(self):
+        ep_dict["target_obj_ids"] = self._target_obj_ids
```
eundersander, line 193: This mapping will be helpful, e.g., because `grasp_mgr_snap_idx` records the sim-object id of the held object.

examples/siro_sandbox/test_episode_save_files.py
```diff
@@ -0,0 +1,111 @@
...
+        assert approx_equal(
+            pkl_step["gui_humanoid"]["cam_yaw"],
+            json_step["gui_humanoid"]["cam_yaw"],
+        )
+
+        pkl_subobj = pkl_step["action"]["action_args"]
+        json_subobj = json_step["action"]["action_args"]
+        keys = [
+            "agent_0_arm_action",
+            "agent_0_oracle_nav_with_backing_up_action",
```
eundersander, line 73: I'm a little concerned about our future ability to make sense of `agent_0_oracle_nav_with_backing_up_action`. It tends to be an index into the PDDL entity list. For example, I saw that it was `3` for navigating to target object #0. I'm not sure how we'll make sense of that 3 when consuming this data.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -681,15 +740,63 @@ def get_grasp_release_controls_text():
             text_delta_y=-50,
         )
 
-    def sim_update(self, dt):
-        # todo: pipe end_play somewhere
+    def record_task_state(self):
```
eundersander, line 743: This function records a few bits and pieces of task/sim state that might be helpful later. I'm somewhat speculating what will be needed later.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -681,15 +740,63 @@ def get_grasp_release_controls_text():
...
+
+        self._step_recorder.record(
+            "target_object_positions", self.get_target_object_positions()
+        )
+
+    def record_action(self, action):
+        action_args = action["action_args"]
+
+        keys_to_clear = [
+            "human_joints_trans",
```
eundersander, line 777: This is a pretty big array of numbers for every step and I saw it massively bloating our file sizes. I also don't think it will be useful data for recomputing metrics. Meanwhile, for visualization, we should expect to use gfx-replay (separate from this PR).

examples/siro_sandbox/sandbox_app.py
eundersander: For example, if we later wanted to compute "distance traveled by each agent", we could do it by inspecting the agent translations here.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -140,16 +160,49 @@ def __init__(self, args, config, gui_input):
...
+        json_filepath = filepath_base + ".json.gz"
+        save_as_json_gzip(self._episode_recorder_dict, json_filepath)
+
+        pkl_filepath = filepath_base + ".pkl.gz"
+        save_as_pickle_gzip(self._episode_recorder_dict, pkl_filepath)
 
-        self.on_environment_reset()
+    def start_episode_recorder(self):
+        assert self._save_filepath_base and self._step_recorder
+        ep_dict: Any = dict()
```
rpartsey, line 225: `ep_dict: Dict[str, Any] = dict()`?

examples/siro_sandbox/test_episode_save_files.py
ASzot: Yeah, I agree. We reconstruct the PDDL to get what the corresponding action is. But we do also return the action executed as a string as well to the info dictionary [here](https://github.com/facebookresearch/habitat-lab/blob/5d3127746925fc232841c2183be795b29719a631/habitat-baselines/habitat_baselines/rl/hrl/hl/neural_policy.py#L240). So this should provide us all the context we need.

examples/siro_sandbox/sandbox_app.py
ASzot: I agree this is unnecessary.

ASzot (approved):

examples/siro_sandbox/sandbox_app.py
rpartsey: It is also possible to create dict with all the values at init:
```
ep_dict = {
    "start_time": datetime.now(),
    ...
}
```

examples/siro_sandbox/sandbox_app.py
rpartsey: Let's then exclude all the keys where `human_joints_trans` in action_arg_key.

rpartsey (approved):
LGTM. Left a few minor comments.
ps
My review was in progress right before you merged.

PR #1383, opened 2023-05-26
[SIRo] - Receptacle caching

aclegg3:
## Motivation and Context

Reloading all Receptacle data at every reset for training is unnecessary drag on SPS. Instead, we should be caching and re-using Receptacle objects.

First stab: cache for all use of Receptacles, but continue to use the per-episode process to compute global AABB.

TODO: better would be to provide a shortcut or cache to directly grab the global AABB instead of re-computing every time.

Secondary changes:
- remove trimesh dependency
- refactor for Magnum API change on Habitat-sim `main`
- add `bounds` to all Receptacles.
  - NOTE 1: bounds of mesh receptacles will be very shallow, so using this for containment is unlikely to succeed. Some configured or default "depth" would be necessary to use the AABB for anything but distance/center.
  - NOTE 2: bounds of concave mesh Receptacles or those with high curvature will be inaccurate.

## How Has This Been Tested

Needs more testing.
Episode generation runs locally, but haven't tested the results.

## Types of changes

## Checklist


habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -593,6 +615,13 @@ def import_tri_mesh(mesh_file: str) -> List[mn.trade.MeshData]:
     return mesh_data
 
 
+# global cache of Receptacles to reduce load time costs
+# maps receptacle name (not unique name) to a Receptacle instance (NOTE: parent_object_handle must be updated per-user-instance)
+_rec_cache: Dict[
+    str, Union[Receptacle, AABBReceptacle, TriangleMeshReceptacle]
```
jimmytyyang, line 621: Thank you, alex! do we ensure the uniqueness of the name over all scenes and all receptacle?

jimmytyyang (commented):
Looks good, thank you Alex! just a small question.

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
aclegg3: Great question. No guarantees here, unfortunately! It *should* be unique, but likely is not across all datasets and there is not any logic to validate this.

I believe these will be unique within HSSD currently.
Could write some logic to validate the dataset. :thinking:

habitat-lab/habitat/datasets/rearrange/samplers/receptacle.py
```diff
@@ -15,7 +15,6 @@
 import corrade as cr
 import magnum as mn
 import numpy as np
-import trimesh
```
jturner65, line 17: Hooray!

PR #1385, opened 2023-05-27
SIRo - Update for Magnum breaking API change

0mdc:
Port of #1356 to the SIRo branch to handle a breaking change in the Magnum API.


ASzot (approved):
Merging now because I need this fix to run on recent habsim versions.

PR #1388, opened 2023-05-29
Speed Upgrades and FP Bug Fixes

ASzot:


Skylion007 (approved):

akshararai (approved):
looks good, I want to test that training works once, and then we can merge.

PR #1389, opened 2023-05-30
Sandbox App CLI episodes filter

rpartsey:
## Motivation and Context

Enabled a possibility to specify a subset of play episodes on the command line (`--episodes-filter` CLI argument).

Episodes filter string should be in the form '0:10 12 14:20:2', where single integer number ('12' in this case) represents an episode id, colon separated integers ('0:10' and '14:20:2') represent start:stop:step ids range.

## How Has This Been Tested
Local testing on MacBook Pro 2021.

## Types of changes

- PR into SIRo branch

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -20,7 +20,7 @@ sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)
 import argparse
 from datetime import datetime
 from functools import wraps
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Set, Tuple
 
 import magnum as mn
 import numpy as np
@@ -84,7 +84,10 @@ class SandboxState(Enum):
 @requires_habitat_sim_with_bullet
 class SandboxDriver(GuiAppDriver):
     def __init__(self, args, config, gui_input):
-        self._dataset = config.habitat.dataset
+        self._dataset_config = config.habitat.dataset
+        self._play_episodes_filter_str = args.episodes_filter
+        self._end_on_success = config.habitat.task.end_on_success
+        self._success_measure_name = config.habitat.task.success_measure
         self._num_recorded_episodes = 0
 
         if args.save_filepath_base:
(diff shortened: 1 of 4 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -170,12 +172,57 @@ def __init__(self, args, config, gui_input):
...
+    def _make_dataset(self, config):
+        from habitat.datasets import make_dataset
+
+        dataset_config = config.habitat.dataset
+        dataset = make_dataset(
+            id_dataset=dataset_config.type, config=dataset_config
+        )
+
+        if self._play_episodes_filter_str is not None:
+            _max_num_digits: int = len(str(len(dataset.episodes)))
```
eundersander, line 188: nit: don't need underscore for local var

examples/siro_sandbox/sandbox_app.py
```diff
@@ -821,6 +875,14 @@ def _update_help_text(self):
                     text_delta_x=-150,
                     text_delta_y=-50,
                 )
+
+            progress_str = f"{self._num_iter_episodes - (self._num_episodes_done + 1)} episodes remaining"
```
eundersander, line 879: We have some help text in `_get_status_text` that instructs the user to press M at the end of the episode. Can you update this, as well? Help them understand when they're done and can just quit the application.

eundersander (approved):
Left some minor comments but otherwise LGTM!

examples/siro_sandbox/sandbox_app.py
rpartsey: I decided to remove `Press M to try again on the next episode.` from the status string since when task is complete or something went wrong, control string changes and contains only "ESC" and (when next episode exists) "M". This way our control and status strings are decoupled nicely and don'd contain duplicate information.

examples/siro_sandbox/sandbox_app.py
rpartsey: Done!

PR #1390, opened 2023-05-31
Data-collection Session Episode Gfx-replays

rpartsey:
## Motivation and Context

Enhanced data-collection session episode gfx-replays. See the [task description](https://docs.google.com/document/d/1KUYO1jJwsTdT8yAJEmCrdpzrgFCu2yhtZf9lWxdasNE/edit#bookmark=id.56xr73iwt93e) for more details.

## How Has This Been Tested
Running Sandbox App locally (MacBook Pro 2021).

**Todo**:  modify [this](https://github.com/facebookresearch/habitat-sim/blob/main/examples/tutorials/nb_python/replay_tutorial.py#L312) tutorial to consume the gfx-replay file output by Sandbox tool and make sure everything is played as expected.  

## Types of changes

- RP into SIRo branch

## Checklist


examples/siro_sandbox/sandbox_app.py
```diff
@@ -739,23 +742,26 @@ def _free_camera_lookat_control(self):
...
         # self._recording_keyframes format:
-        #     List['{"keyframe": {...}', '{"keyframe": {...}', ...]
+        #     List['{"keyframe":{...}', '{"keyframe":{...}',...]
         # Output format:
-        #     '{"keyframes": [{...}, {...}, ...]}'
-        json_keyframes = "".join(
-            keyframe[12:-1] + ","
-            for keyframe in self._recording_keyframes[:-1]
+        #     '{"keyframes":[{...},{...},...]}'
+        json_keyframes = ",".join(
```
rpartsey, line 753: This should be equivalent to old code.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -739,23 +742,26 @@ def _free_camera_lookat_control(self):
         )
 
     def _save_recorded_keyframes_to_file(self):
+        if not self._recording_keyframes:
+            return
+
         # Consolidate recorded keyframes into a single json string
         # self._recording_keyframes format:
-        #     List['{"keyframe": {...}', '{"keyframe": {...}', ...]
+        #     List['{"keyframe":{...}', '{"keyframe":{...}',...]
```
rpartsey, line 750: Removed whitespaces as there were no whitespaces in the keyframe strings.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -1032,12 +1038,9 @@ def sim_update(self, dt):
         post_sim_update_dict: Dict[str, Any] = {}
 
         if self.gui_input.get_key_down(GuiInput.KeyNS.ESC):
+            self._check_save_episode_data()
```
rpartsey, line 1041: Save recorded data not only at episode reset (pressed M) but also if App exit event was issued or all episodes were done and have to exit (pressed ESC).

eundersander: I think we're asking for trouble writing code this way. Let's have a single code-path for "end of episode". I think this could be `_check_reset_environment`. Call `_check_save_episode_data` in there. And call `_check_reset_environment` here when the user presses ESC. And we could consider renaming `_check_reset_environment` to `_end_episode_and_try_reset` or similar.

eundersander: Nice catch, btw! I guess the original code failed to save the last episode.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -739,23 +749,26 @@ def _free_camera_lookat_control(self):
...
 
         # Save keyframes to file
-        with open(self._gfx_replay_save_path, "w") as json_file:
-            json_file.write(json_content)
+        filepath_base = self._find_episode_save_filepath_base()
+        filepath = filepath_base + ".gfx_replay.json.gz"
+        save_as_gzip(json_content.encode("utf-8"), filepath)
+
+        # reset recording_keyframes
+        self._recording_keyframes.clear()
```
eundersander, line 771: Gfx-replay keyframes are incremental. Each frame just describes the changes to the visual state since the last keyframe. I.e. Keyframe #2 only makes sense when it immediately follows keyframe #1. We can't arbitrarily clear/discard old ones and then expect a sequence of new ones to "work". We have to be very careful to only do this clear at about the same time that we reset the underlying sim.

I think the solution here is to move this clear to wherever we reset the sim.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -739,23 +749,26 @@ def _free_camera_lookat_control(self):
         )
 
     def _save_recorded_keyframes_to_file(self):
+        if not self._save_filepath_base or not self._recording_keyframes:
```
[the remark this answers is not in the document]
eundersander: Alternately, you could just avoid calling this if `self._save_filepath_base` is False.

I do think the check for `not self._recording_keyframes` is more like basic error-checking (don't want to save an empty file), so no reason to call the function `check_` just because you're doing that check.

examples/siro_sandbox/sandbox_app.py
rpartsey: :white_check_mark:

examples/siro_sandbox/sandbox_app.py
```diff
@@ -314,9 +314,19 @@ def _reset_environment(self):
             else None
         )
 
-        if self._save_filepath_base:
+        self._check_save_episode_data()
+
+    def _check_save_episode_data(self):
```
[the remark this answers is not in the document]
rpartsey: Yes, agree. :white_check_mark:

examples/siro_sandbox/sandbox_app.py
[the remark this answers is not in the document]
rpartsey: :white_check_mark:

examples/siro_sandbox/sandbox_app.py
```diff
@@ -1274,28 +1284,28 @@ def parse_debug_third_person(args, framebuffer_size):
...
+        help="Save the episode data to file. Use --save-filepath-base to specify the filepath base.",
     )
     parser.add_argument(
         "--save-filepath-base",
         default=None,
         type=str,
-        help="filepath base used for saving various session data files. Include a full path including basename, but not an extension.",
+        help="Filepath base used for saving various session data files. Include a full path including basename, but not an extension.",
     )
     args = parser.parse_args()
```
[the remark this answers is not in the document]
rpartsey: :white_check_mark:

examples/siro_sandbox/sandbox_app.py
rpartsey: :white_check_mark:

eundersander: Update here based on more investigation from Ruslan and our Slack chat: it does not work to simply clear old keyframes, even when the sim is reset. If we try this, Ruslan observed that the second gfx-replay file is missing vital `loads`.

eundersander (approved):
LGTM!

PR #1406, opened 2023-06-07
SIRo sandbox: one-line fix to make neural policies work

eundersander:
## Motivation and Context

One-line fix to make neural policies work with sandbox app. Update readme to talk about snapshots.

## How Has This Been Tested

## Types of changes
PR into SIRo branch

## Checklist

### DIFF
--- a/examples/siro_sandbox/README.md
+++ b/examples/siro_sandbox/README.md
@@ -6,54 +6,15 @@
 This is a 3D interactive GUI app for testing various pieces of SIRo, e.g. rearrangement episode datasets, Fetch and Spot robots, humanoids (controllers, animation, skinning), trained agent policies, batch rendering and other visualization.
 
 # Known Issues
-* June 5: The tool is not very stable in the `SIRo` branch due to rapid iteration in various parts of the codebase. See recently-tested snapshot below including specific commits that have been tested to work together.
+* The tool is not very stable in the `SIRo` branch due to rapid iteration in various parts of the codebase. See Snapshots section below for best results.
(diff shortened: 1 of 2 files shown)


rpartsey (approved):
I can successfully run commands from the [SIRo Sandbox Snapshots Google Doc](https://docs.google.com/document/d/1cvKuXXE2cKchi-C_O7GGVFZ5x0QU7J9gHTIETzpVKJU/edit#) on my MacBook 2021 machine. 

Checkpoint loading seems to be working as expected.

I wanted to make sure that checkpoint evaluation is also working as expected (not only loading). First check I did is comparison trained checkpoint vs randomly initialized agent 
(`habitat_baselines.eval.should_load_ckpt=True` vs `habitat_baselines.eval.should_load_ckpt=False`) in free camera mode (with `--gui-controlled-agent-index` omitted). And I do see a difference, meaning that trained checkpoint is behaving differently from (better that) random agent. 

But how can we be sure that this behaviour is in fact expected behaviour? To test this I was comparing  Sandbox App free camera mode vs videos recorded by [offline eval runs](https://github.com/facebookresearch/habitat-lab/blob/SIRo/SIRO_README.md#demo-fetch-human-fixed-planner). And I do see the difference in agents behaviour. For example, humanoid is always stuck and not or barely moving once it grasped or approached the object. Also Spot is never able to grasp and carry an object. I see 3/10 successful videos from  offline eval runs but none of Sandbox App free camera mode is successful. I'm a bit concerned about this.

PR #1423, opened 2023-06-16
Fix snap point

jimmytyyang:
## Motivation and Context

 fix snap points

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
@@ -441,6 +441,12 @@ class RearrangeSim(HabitatSim):
             self.pathfinder.island_radius(p) for p in self._navmesh_vertices
         ]
         self._max_island_size = max(self._island_sizes)
+        largest_size_vertex = self._navmesh_vertices[
+            np.argmax(self._island_sizes)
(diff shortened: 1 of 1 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -492,31 +495,22 @@ def safe_snap_point(self, pos: np.ndarray) -> np.ndarray:
...
             )
-            island_radius = self.pathfinder.island_radius(new_pos)
-
-        if np.isnan(new_pos[0]) or island_radius != self._max_island_size:
-            # This is a last resort, take a navmesh vertex that is closest
-            use_verts = [
-                x
-                for s, x in zip(self._island_sizes, self._navmesh_vertices)
-                if s == self._max_island_size
+            largest_island_id = island_areas.index(max(island_areas))
```
ASzot, line 505: We already store `self._max_island_size`. Can you instead use that instead of having to create a new cache?

ASzot:
There is no need for `self.scene_id_2_island_id` since we already compute this information in `_load_navmesh`. But other than that, LGTM.

akshararai (approved):
lgtm, thanks for fixing Jimmy!

PR #1432, opened 2023-06-20
fix snap point NaN issue

jimmytyyang:


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -502,31 +502,17 @@ def safe_snap_point(self, pos: np.ndarray) -> np.ndarray:
...
-        new_pos = self.pathfinder.snap_point(pos)
-        island_radius = self.pathfinder.island_radius(new_pos)
-
-        if np.isnan(new_pos[0]) or island_radius != self._max_island_size:
-            # The point is not valid or not in a different island. Find a
-            # different point nearby that is on a different island and is
-            # valid.
+        new_pos = self.pathfinder.snap_point(pos, self._largest_island_idx)
+        regen_i = 0
+        while np.isnan(new_pos[0]) and regen_i < 10:
```
ASzot, line 507: Is this expected to commonly fail? If not, could you add some warning log statement on failure so we can know when it's happening?

xavierpuigf: Added an assert

PR #1331, opened 2023-05-10
General Hab2 Bug Fixes

ASzot:
## Motivation and Context

Bug fixes for Habitat 2.0 tasks:
- Spawning the agent on the largest navigation island.
- Bug fix for checking `on_top` predicate.
- Fixed bug with grasping manager not releasing grasp constraints.
- Refactored `hierarchical_policy`. 
- Compatibility with `rnn_build_seq_info=None` in distributed updater.

## How Has This Been Tested

End-to-end training.

## Types of changes

- **\[Refactoring\]** Large changes to the code that improve its functionality or performance
- **\[Bug Fix\]** (non-breaking change which fixes an issue)
- **\[Development\]** A pull request that add new features to the [habitat-lab](/habitat-lab) task and environment codebase. Development Pull Requests must be small (less that 500 lines of code change), have unit testing, very extensive documentation and examples. These are typically new tasks, environments, sensors, etc... The review process for these Pull Request is longer because these changes will be maintained by our core team of developers, so make sure your changes are easy to understand!

PR #1436, opened 2023-06-25
[Development] Batch renderer PPO training integration

0mdc:
## Motivation and Context

This changeset adds supports for using the PPO trainer with the batch renderer.

Example run command:
```
habitat-baselines/habitat_baselines/run.py --config-name=rearrange/rl_skill.yaml habitat_baselines.trainer_name="ppo" habitat.simulator.renderer.enable_batch_renderer=True habitat.simulator.habitat_sim_v0.enable_gfx_replay_save=True habitat.simulator.create_renderer=False habitat.simulator.concur_render=False habitat.simulator.renderer.composite_files=[path/to/composite_file.gltf]
```

Documentation to train with the batch renderer: https://github.com/facebookresearch/habitat-lab/pull/1443

## How Has This Been Tested

Tested locally against `rl_skill.yaml` and with extended `test_trainers` test.

## Checklist

### DIFF
--- a/habitat-baselines/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat-baselines/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -268,6 +268,7 @@ class PPOTrainer(BaseRLTrainer):
         self._ppo_cfg = self.config.habitat_baselines.rl.ppo
 
         observations = self.envs.reset()
+        observations = self.envs.post_step(observations)
         batch = batch_obs(observations, device=self.device)
         batch = apply_obs_transforms_batch(batch, self.obs_transforms)  # type: ignore
 
@@ -405,6 +406,7 @@ class PPOTrainer(BaseRLTrainer):
             ]
 
         with g_timer.avg_time("trainer.update_stats"):
+            observations = self.envs.post_step(observations)
             batch = batch_obs(observations, device=self.device)
             batch = apply_obs_transforms_batch(batch, self.obs_transforms)  # type: ignore
 
@@ -848,6 +850,7 @@ class PPOTrainer(BaseRLTrainer):
             self._agent.load_state_dict(ckpt_dict)
 
         observations = self.envs.reset()
+        observations = self.envs.post_step(observations)
         batch = batch_obs(observations, device=self.device)
         batch = apply_obs_transforms_batch(batch, self.obs_transforms)  # type: ignore
 
@@ -963,6 +966,8 @@ class PPOTrainer(BaseRLTrainer):
             )
             for i in range(len(policy_infos)):
                 infos[i].update(policy_infos[i])
+            
+            observations = self.envs.post_step(observations)
             batch = batch_obs(  # type: ignore
                 observations,
                 device=self.device,
--- a/habitat-baselines/habitat_baselines/rl/ppo/ppo_trainer.py
+++ b/habitat-baselines/habitat_baselines/rl/ppo/ppo_trainer.py
@@ -966,7 +966,7 @@ class PPOTrainer(BaseRLTrainer):
             )
             for i in range(len(policy_infos)):
                 infos[i].update(policy_infos[i])
-            
+
             observations = self.envs.post_step(observations)
             batch = batch_obs(  # type: ignore
                 observations,
--- a/habitat-lab/habitat/core/batch_rendering/env_batch_renderer.py
+++ b/habitat-lab/habitat/core/batch_rendering/env_batch_renderer.py
@@ -119,6 +119,12 @@ class EnvBatchRenderer:
         # Pre-load graphics assets using composite GLTF files.
         EnvBatchRenderer._load_composite_files(config, self._replay_renderer)
 
+    def close(self) -> None:
+        r"""
+        Release the resources and graphics context.
+        """
+        self._replay_renderer = None
+
     def post_step(self, observations: List[OrderedDict]) -> List[OrderedDict]:
         r"""
         Renders observations for all environments by consuming keyframe observations.
--- a/habitat-lab/habitat/core/batch_rendering/env_batch_renderer.py
+++ b/habitat-lab/habitat/core/batch_rendering/env_batch_renderer.py
@@ -72,7 +72,9 @@ class EnvBatchRenderer:
         assert (
             not config.habitat.simulator.create_renderer
         ), "Batch renderer requires create_renderer to be disabled in config."
-
+        assert (
+            not config.habitat.simulator.debug_render
+        ), "Batch renderer requires debug_render to be disabled in config."
         logger.warn(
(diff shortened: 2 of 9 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -842,14 +850,19 @@ def step(self, action: Union[str, int]) -> Observations:
...
             self.gfx_replay_manager.save_keyframe()
 
         if self._needs_markers:
             self._update_markers()
 
         # TODO: Make debug cameras more flexible
-        if "third_rgb" in obs and self._debug_render:
+        if (
+            "third_rgb" in obs
+            and self._debug_render
```
ASzot, line 863: Can you change this to just set `self._debug_render` to False when `self._batch_render` is enabled? This could catch more debug render situations, rather than just checking this `if` statement.

habitat-baselines/habitat_baselines/rl/ppo/ppo_trainer.py
```diff
@@ -268,6 +268,7 @@ def _init_train(self, resume_state=None):
         self._ppo_cfg = self.config.habitat_baselines.rl.ppo
 
         observations = self.envs.reset()
+        observations = self.envs.post_step(observations)
```
ASzot, line 271: Can you raise an exception in the VER trainer to explain that batched rendering doesn't work with VER for now?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
0mdc: I instead added an assert to `EnvBatchRenderer` to explicit about this.

habitat-lab/habitat/core/vector_env.py
```diff
@@ -480,6 +480,9 @@ def close(self) -> None:
 
         self._is_closed = True
 
+        # Force the batch renderer to release its context.
+        self._batch_renderer = None
```
0mdc, line 484: The tests were leaking the graphics context because `VectorEnv` would not clear the batch renderer shared pointer.

eundersander: This looks odd. There is no guarantee of Python actually deleting this object when the references go to zero, right? How about an explicit `close()` on the batch renderer?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -821,7 +822,14 @@ def step(self, action: Union[str, int]) -> Observations:
...
-        if self._concur_render:
+        if self._batch_render:
+            for _ in range(self.ac_freq_ratio):
+                self.internal_step(-1, update_articulated_agent=False)
+
+            self._prev_sim_obs = self.get_sensor_observations()
+            self.add_keyframe_to_observations(self._prev_sim_obs)
+            obs = self._prev_sim_obs
+        elif self._concur_render:
             self._prev_sim_obs = self.start_async_render()
```
eundersander, line 833: I don't understand this line (and I know this is just existing code). As far as I can see, `start_async_render` doesn't return anything. So is this code setting `_prev_sim_obs` to None? And the third case (not batch rendering and not concurrent rendering) doesn't update `_prev_sim_obs` at all? And so it's not clear to me what your code should do. Can you ask @ASzot what we should be doing with `_prev_sim_obs` in here? He recently made [this change](https://github.com/facebookresearch/habitat-lab/pull/1331/files#diff-2e30591431a2b62e1c995e1d0af87b17231e46f20c4fb892ad4ee2a0bc8e4aebL761).

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -842,7 +850,8 @@ def step(self, action: Union[str, int]) -> Observations:
             )
             self.add_perf_timing("get_sensor_observations", t_start)
 
-        if self._enable_gfx_replay_save:
+        # TODO: Support recording while batch rendering
+        if self._enable_gfx_replay_save and not self._batch_render:
```
eundersander, line 854: We definitely need to clean this up in July. Super-confusing that when both of these are set to true, we don't save any keyframes. You should document this in your markdown doc.

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
```diff
@@ -210,9 +210,14 @@ def reset(self, episode: Episode, fetch_observations: bool = True):
...
-        obs = self._sim._sensor_suite.get_observations(
-            self._sim.get_sensor_observations()
-        )
+        obs = self._sim.get_sensor_observations()
+
+        if not self._sim.sim_config.enable_batch_renderer:
+            # Post-process visual sensor observations
+            obs = self._sim._sensor_suite.get_observations(obs)
+        else:
+            # Keyframes are added so that the simulator state can be reconstituted later.
```
eundersander, line 219: How about a comment about how, for `enable_batch_renderer`, the equivalent post-processing will happen later in `whatever_function_name`.

eundersander (commented):
Good progress. I'd like to take one more look before merging so let me know!

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
0mdc: Right, this is odd. The only place where `_prev_sim_obs` is used is to return `previous_step_collided()` (within the parent object `HabitatSim`).

Let me see what's the status on this.

0mdc: @ASzot Mentioned that it’s not used in any of the Habitat 2.0 tasks. The RearrangeSim is only for Habitat 2.0 tasks, so it doesn’t impact that Habitat 1.0 tasks that use this.

So, we shouldn’t be using the _prev_sim_obs here.

Good catch.

eundersander (commented):
Still concerned about freeing the replay renderer, see  comment below.

habitat-lab/habitat/core/batch_rendering/env_batch_renderer.py
```diff
@@ -119,6 +119,12 @@ def __init__(self, config: DictConfig, num_envs: int) -> None:
         # Pre-load graphics assets using composite GLTF files.
         EnvBatchRenderer._load_composite_files(config, self._replay_renderer)
 
+    def close(self) -> None:
+        r"""
+        Release the resources and graphics context.
+        """
+        self._replay_renderer = None
```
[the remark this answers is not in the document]
0mdc: Fixed! An explicit `replay_renderer.close()` function was added. Depends on this sim branch: https://github.com/facebookresearch/habitat-sim/pull/2145

eundersander (approved):
LGTM!

PR #1479, opened 2023-07-31
SIRo HITL tool policy eval fix and refactor

rpartsey:
## Motivation and Context

This is first step towards unifying `PPOTrainer`'s `_eval_checkpoint` and HITL tool baselines evaluation. In scope of this PR, I replace `BaselinesController`'s `_actor_critic` initialization code with `SingleAgentAccessMgr` and `MultiAgentAccessMgr` and use the same evaluation functions calls as in the `_eval_checkpoint` (also using the same naming). 

Link to the human-spot-GTCoord.pth checkpoint: https://gatech.enterprise.slack.com/files/U013PLZ8ZTL/F05KBL5FJ4V/human-spot-gtcoord.pth.

**Sandbox App launch commands**
Controlling humanoid:
```
python examples/siro_sandbox/sandbox_app.py \
--never-end \
--gui-controlled-agent-index 1 \
--cfg experiments_hab3/pop_play_kinematic_oracle_humanoid_spot_fp.yaml \
--cfg-opts \
habitat_baselines.num_environments=1 \
habitat_baselines.eval_ckpt_path_dir=\'data/siro_checkpoints/human-spot-GTCoord.pth\' \
habitat_baselines.evaluate=True \
habitat_baselines.eval.should_load_ckpt=True \
habitat_baselines.rl.agent.num_pool_agents_per_type=\[1,1\] \
habitat.task.actions.agent_0_oracle_nav_with_backing_up_action.longitudinal_lin_speed=10.0 \
habitat.task.actions.agent_1_oracle_nav_with_backing_up_action.lin_speed=10.0 habitat_baselines.verbose=True \
~habitat.task.measurements.agent_blame_measure
```

https://github.com/facebookresearch/habitat-lab/assets/37709869/fdf8f0a9-8fe1-4802-83bb-a102327d9ebd

Free camera mode:
```
python examples/siro_sandbox/sandbox_app.py \
--never-end \
--cfg experiments_hab3/pop_play_kinematic_oracle_humanoid_spot_fp.yaml \
--cfg-opts \
habitat_baselines.num_environments=1 \
habitat_baselines.eval_ckpt_path_dir=\'data/siro_checkpoints/human-spot-GTCoord.pth\' \
habitat_baselines.evaluate=True \
habitat_baselines.eval.should_load_ckpt=True \
habitat_baselines.rl.agent.num_pool_agents_per_type=\[1,1\] \
habitat.task.actions.agent_0_oracle_nav_with_backing_up_action.longitudinal_lin_speed=10.0 \
habitat.task.actions.agent_1_oracle_nav_with_backing_up_action.lin_speed=10.0 habitat_baselines.verbose=True
```

https://github.com/facebookresearch/habitat-lab/assets/37709869/a527e1a3-f982-4f26-9c2d-4ab06216883a

PPOTrainer._eval_checkpoint:

https://github.com/facebookresearch/habitat-lab/assets/37709869/4acc5954-e760-4470-8356-c2b0c05e90cd

## Types of changes

- **Refactoring** 

## Checklist


examples/siro_sandbox/sandbox_app.py
```diff
@@ -1424,6 +1432,17 @@ def _parse_debug_third_person(args, framebuffer_size):
             )
             gui_controlled_agent_config.sim_sensors.clear()
 
+            for sensor_name in [
```
eundersander, line 1435: What motivated this change? This looks like a great bugfix! Is this actually turning off some unnecessary rendering that we have been doing in the HITL tool?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -214,19 +228,23 @@ def get_play_episodes_ids(play_episodes_filter_str):
...
 
     def _env_step(self, action):
-        self._obs = self.env.step(action)
-        self._metrics = self.env.get_metrics()
+        (
+            self._obs,
+            reward,
+            done,
+            self._metrics,
+        ) = self.gym_habitat_env.direct_hab_step(action)
```
eundersander, line 236: Why can't you use the already-public `step`? Why do you have to also expose `direct_hab_step`?

eundersander (commented):
Looking good. I left some small comments. I'd like to take another look before you merge.

examples/siro_sandbox/sandbox_app.py
rpartsey: @ASzot I'd appreciate your thought here. 

We may have one or two controllers at SandboxApp runtime:
**Single agent case**
1. SingleAgentBaselineController - one policy-controlled agent
2. GuiController - one GUI-controlled agent

**Multi agent case**
1. MultiAgentBaselineController - multiple policy-controlled agents
2. [GuiController, SingleAgentBaselineController] - one GUI-controlled agent + one policy-controlled agent
3. [GuiController, MultiAgentBaselineController] - one GUI-controlled agent + multiple policy-controlled agents (**not supported yet**)

In **Multi agent case**, for example, option 2: observation and action spaces are split between two controllers and both GuiController and SingleAgentBaselineController return an action. Currently, both of these actions are already in the form {"action": ..., "action_args": ...} i.e. for GuiController the action is constructed in that form and for SingleAgentBaselineController `continuous_vector_action_to_hab_dict` is called after `agent.actor_critic.act()`.

After that, these actions are concatenated: 
```python
  all_names = []
  all_args = {}
  for controller in self.controllers:
      ctrl_action = controller.act(obs, self._env)
      all_names.extend(ctrl_action["action"])
      all_args.update(ctrl_action["action_args"])
  action = {"action": tuple(all_names), "action_args": all_args}
```

Then I want to pass aggregated action to the environment (HabGymWrapper). But it's step method doesn't accept actions in the form {"action": ..., "action_args": ...}. To sidestep this I exposed `direct_hab_step` and call it directly.

How would you suggest to combine controllers actions and pass to the environment? Would it be better to have controllers returning actions as `np.array`, then merging them and passing to the env.step (not having controllers returning {"action": ..., "action_args": ...})?

examples/siro_sandbox/sandbox_app.py
rpartsey: This is motivated by the fact that policy evaluation code expects to be operating on the the `GymHabitatEnv` (`habitat.Env` wrapped in a bunch of wrappers that modify/transform/filter observations returned by the env and also actions before they are passed to the env directly).

You may see that now I define two env references:
```python
self.gym_habitat_env: "GymHabitatEnv" = (
    habitat.gym.make_gym_from_config(config=config, dataset=dataset)
)
self.habitat_env: habitat.Env = (
    self.gym_habitat_env.unwrapped.habitat_env
)
```
GUI and app related core operates directly on `self.habitat_env` and baselines (agents) on self.gym_habitat_env.

examples/siro_sandbox/controllers/controller_abc.py
```diff
@@ -0,0 +1,220 @@
...
+        self._load_agent_state_dict(checkpoint)
+
+    def _batch_and_apply_transforms(self, obs):
+        batch = batch_obs(obs, device=self.device)
+        batch = apply_obs_transforms_batch(batch, self._obs_transforms)
+
+        return batch
+
+    def on_environment_reset(self):
+        self._test_recurrent_hidden_states = torch.zeros(
```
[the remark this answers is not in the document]
rpartsey: > It would be cleaner to declare this member var in __init__, even just setting it to None.
done

examples/siro_sandbox/sandbox_app.py
ASzot: I suggest having the controllers always returning actions as `np.array` and calling `.step` instead of `.direct_hab_step`. The Gym API is the interface for interacting with the Habitat environment. You should refrain from using the underlying Habitat internals such as the `{"action": ..., "action_args": ...}` action format.

xavierpuigf: Is there a reason for this, @ASzot ? I find it hard to get the right index in the np.array containing all the actions, and dict seems more intuitive. Is there a check that the gymwrapper is doing that we should account for?

In the previous version of the sandbox, we were calling the `habitat.core.env.Env` in the sandbox. Why not keeping that one instead of the `GymWrapper`?

examples/siro_sandbox/controllers/controller_abc.py
```diff
@@ -215,12 +214,10 @@ def act(self, obs, env):
         else:
             step_data = [a.item() for a in action_data.env_actions.cpu()]
 
-        action = gym_wrapper.continuous_vector_action_to_hab_dict(
-            self._env_spec.orig_action_space,
-            self._env_spec.action_space,
-            step_data[0],
-        )
+        action = step_data[0]
```
eundersander, line 217: How about assert `len(action_data.env_actions) == 1` and then directly grab the first item? If we know there's only one environment, let's assert that early. Let's not construct unnecessary one-element lists.

eundersander (approved):
I requested a few minor changes. Aside from that, LGTM. If my changes aren't doable or don't make sense, let's discuss.

PR #1480, opened 2023-08-01
lint

xavierpuigf:
## Motivation and Context

Correct relative transform human. This fix makes the relative transform of the human match that of the robot. Before the human's base_transformation was shifted. Now it works as it should: local Z coordinate of the human points upwards. This is needed so that pddl_Sim_state works.

## How Has This Been Tested

Sandbox app and test episodes

## Types of changes

## Checklist


akshararai (approved):
looks great, thanks for fixing!

PR #1486, opened 2023-08-07
[SIRo] ZSC with Learned Agents

ASzot:


akshararai (approved):
looks good to me! I will test it locally quickly to make sure it works with some checkpoints, but overall makes sense. Thanks for doing this!

xavierpuigf:
LGTM for learned agents. For planning-based agents, note that plan_idx corresponds to the following:

0,1,2,3: doing no goal, goal 1, goal 2, both goals.
-1,-2,-3,-4: correspond to always the same goal (rand [1]), always a single goal (rand [1,2]),  one or two goals (rand [1,2,3]), any possibility (rand [0,1,2,3]).

In the past we were always evaluating with the -X options. I would keep with that.

PR #1508, opened 2023-08-17
[hab3_merge] Add Skills for the robot to move forward/backward and turn

xavierpuigf:
## Motivation and Context

Add skills to allow low level movement for the robot. This is needed so that the robot can make way to the human when about to collide.

## How Has This Been Tested

Tested with hl_fixed policy, verifying that the agent would turn when the skill is called.

## Types of changes

## Checklist


ASzot (changes requested):
This looks good, but if we are merging into main, we should have a test for this new skill.

xavierpuigf:
Thanks @ASzot ! I added pytests for the new skills

ASzot (approved):

PR #1511, opened 2023-08-18
[SIRo] Place Block

ASzot:
Disables picking an object after is placed.

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
@@ -48,6 +48,7 @@ class RearrangeGraspManager:
         self.ee_index = ee_index
 
         self._kinematic_mode = self._sim.habitat_config.kinematic_mode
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -307,6 +314,11 @@ def snap_to_obj(
         :param force: Will kinematically snap the object to the robot's end-effector, even if
             the object is already in the grasped state.
         """
+
+        if snap_obj_id in self._did_place:
```
akshararai, line 318: this should use the `self._block_double_pick` flag too right?

akshararai (approved):
small bug I think where you don't use the self._block_double_pick flag when blocking

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
ASzot: Uh oh you're right. I just fixed.

PR #1529, opened 2023-08-23
Fix solo GUI-controlled humanoid mode

rpartsey:
## Motivation and Context

This PR removes oracle related sim sensors, lab sensors and task measurements.  Currently, it is done via code, but in the future iterations we may want to consider having app config files that could override base config (remove update actions/sensors/measure etc). 

Launch command:
```
"program": "examples/siro_sandbox/sandbox_app.py",
"env": {
    "HABITAT_SIM_LOG": "warning",
    "MAGNUM_LOG": "warning"
},
"args": [
    "--disable-inverse-kinematics",
    "--gui-controlled-agent-index", "0",
    "--never-end",
    "--app-state", "rearrange",  // fetch  rearrange socialnav
    "--cfg", "experiments_hab3/single_agent_pddl_planner_kinematic_oracle_humanoid.yaml",
    "--cfg-opts", 
    "habitat_baselines.evaluate=True", 
    "habitat_baselines.num_environments=1",          
]
```

## Types of changes

- **Bug Fix** 

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -719,23 +719,6 @@ if __name__ == "__main__":
                 )
                 exit()
 
-            # avoid camera sensors for GUI-controlled agents
-            gui_controlled_agent_config = get_agent_config(
-                sim_config, agent_id=args.gui_controlled_agent_index
-            )
-            gui_controlled_agent_config.sim_sensors.clear()
-
-            for sensor_name in [
-                "head_depth",
-                "head_rgb",
-                "has_finished_oracle_nav",
-            ]:
-                agent_sensor_name = (
-                    f"agent_{args.gui_controlled_agent_index}_{sensor_name}"
-                )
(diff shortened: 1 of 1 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -750,21 +733,60 @@ def _parse_debug_third_person(args, framebuffer_size):
                 )
                 exit()
 
+            # avoid camera sensors for GUI-controlled agents
```
eundersander, line 736: One day soon, we should organize all this config-customization logic and other logic that runs at startup before `SandboxDriver.__init__`. E.g. put it into a class with readable helper functions. For now, this is fine.

eundersander (approved):
LGTM. Can you update the snapshots doc with a snapshot for the solo-humanoid rearrange use case? This can be before or after the PR merges (I realize we don't have a final commit hash until the PR merges).

PR #1546, opened 2023-08-29
HITL fetch state remote mode switch

rpartsey:
## Motivation and Context

Adaptation of the `app_state_fetch.py` to support GUI-controlled humanoid following arbitrary pose (position and orientation change, for example, from the VR headset).

`gui_navigation_helper.py` is refactored to support calculating walk direction to arbitrary position and once that position is reached walk direction is set to rotation vector. Walk direction is represented as a tuple of (walk_dir, distance_multiplier). If distance_multiplier = 1 agent moves in the walk_dir by |walk_dir|, if distance_multiplier is 0 agent rotates on the spot in the walk_dir.

 `gui_controller.py` is not changed since it was designed to follow the walk_dir specified as an actuation hint. 
 
 In theory, switching to VR headset pose following should be enabled by adding `--remote-gui-mode` argument and providing `self._sandbox_service.remote_gui_input`.

## How Has This Been Tested
Launching app locally on MacBook Pro 2021.

## Types of changes

- **Development:** PR into SIRo_hitl

## Checklist

### DIFF
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -108,7 +108,7 @@ class AppStateFetch(AppState):
         walk_dir = None
         if not self._first_person_mode:
             candidate_walk_dir = (
-                self._nav_helper.viz_and_get_humanoid_walk_dir()
+                self._nav_helper.viz_and_get_humanoid_walk_dir_from_ray_cast()
             )
             if self._sandbox_service.gui_input.get_mouse_button(
                 GuiInput.MouseNS.RIGHT
@@ -172,6 +172,28 @@ class AppStateFetch(AppState):
                 mn.Color3(255 / 255, 255 / 255, 0),
                 24,
             )
+
+    def _viz_agent_go_to_box(self):
+        color = mn.Color3(255 / 255, 0 / 255, 0 / 255)  # red
+
+        agent_pos, _ = self._get_agent_pose()
+
+        box_half_size = 0.5
+        box_offset = mn.Vector3(
+            box_half_size, 2 * box_half_size, box_half_size
+        )
+        self._sandbox_service.line_render.draw_box(
+            agent_pos - box_offset,
+            agent_pos + box_offset,
+            color,
+        )
+        circle_radius = 0.1
+        agent_pos[1] = self._get_agent_feet_height()
+        self._sandbox_service.line_render.draw_circle(
+            agent_pos,
+            circle_radius,
+            color,
+        )
 
     def get_gui_controlled_agent_index(self):
         return self._gui_agent_ctrl._agent_idx
@@ -250,11 +272,15 @@ class AppStateFetch(AppState):
                 text_delta_y=-50,
             )
 
-    def _get_camera_lookat_pos(self):
+    def _get_agent_pose(self):
         agent_root = get_agent_art_obj_transform(
             self.get_sim(), self.get_gui_controlled_agent_index()
         )
-        lookat = agent_root.translation + mn.Vector3(0, 1, 0)
+        return agent_root.translation, agent_root.rotation
+
+    def _get_camera_lookat_pos(self):
+        agent_pos, _ = self._get_agent_pose()
+        lookat = agent_pos + mn.Vector3(0, 1, 0)
         return lookat
 
     def sim_update(self, dt, post_sim_update_dict):
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -106,17 +106,33 @@ class AppStateFetch(AppState):
                         ]
 
         walk_dir = None
+        distance_multiplier = 0.0
         if not self._first_person_mode:
-            candidate_walk_dir = (
-                self._nav_helper.viz_and_get_humanoid_walk_dir_from_ray_cast()
-            )
+            if self._sandbox_service.args.remote_gui_mode:
+                (
+                    candidate_walk_dir,
+                    candidate_distance_multiplier,
+                ) = self._nav_helper._get_humanoid_walk_dir_from_remote_gui_input(
+                    visualize_path=False
+                )
+            else:
+                (
+                    candidate_walk_dir,
+                    candidate_distance_multiplier,
+                ) = self._nav_helper._get_humanoid_walk_dir_from_ray_cast(
+                    visualize_path=True
+                )
+
             if self._sandbox_service.gui_input.get_mouse_button(
                 GuiInput.MouseNS.RIGHT
             ):
                 walk_dir = candidate_walk_dir
+                distance_multiplier = candidate_distance_multiplier
 
+        assert isinstance(self._gui_agent_ctrl, GuiHumanoidController)
         self._gui_agent_ctrl.set_act_hints(
             walk_dir,
+            distance_multiplier,
             grasp_object_id,
             drop_pos,
             self._camera_helper.lookat_offset_yaw,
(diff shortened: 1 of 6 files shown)


examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -100,11 +100,86 @@ def viz_and_get_humanoid_walk_dir(self):
...
+
+        if not ray or ray.direction.y >= 0 or ray.origin.y <= floor_y:
+            return None
+
+        dist_to_floor_y = (ray.origin.y - floor_y) / -ray.direction.y
+        target_on_floor = ray.origin + ray.direction * dist_to_floor_y
+
+        return target_on_floor
+
+    def _compute_turn_dir(self, target_rot_quat):
```
rpartsey, line 175: I'm not sure if this is a correct way to convert quaternion to direction vector.

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -100,11 +100,86 @@ def viz_and_get_humanoid_walk_dir(self):
...
+    ):
+        walk_dir = None
+        distance_multiplier = 0.0
+        geodesic_dist_threshold = 0.05
+
+        found_path, path = self._get_humanoid_walk_path_to(target_pos)
+        if (
+            found_path
+            and len(path.points) >= 2
+            and path.geodesic_distance >= geodesic_dist_threshold
```
rpartsey, line 149: `HumanoidRearrangeController` doesn't move the agent exactly `|walk_dir|` in the `walk_dir` direction. It either overshoots or undershoots and we usually end up in the situations when agent is moving forward and backward in the proximity if the target position. Similar happens to rotations: if agent rotates on the spot it can move from side to side. That's why I use `geodesic_dist_threshold`.

examples/siro_sandbox/gui_navigation_helper.py
eundersander: If you have a Magnum Quaternion, you can use this: https://doc.magnum.graphics/python/magnum/Quaternion/#transform_vector

Generally, we should try to use Magnum vector/matrix/quaternion classes instead of numpy for 3D operations.

Another issue here is that the input rotation may not generally be around the up axis (e.g. a VR headset looking slightly up or down, or tilted slightly), in which case your output vector won't be in the ground plane. A possible fix here is to set heading_vector.y = 0 and then re-normalize the vector.

eundersander, line 175: A better name is probably "_compute_facing_dir" or "_compute_forward_dir".

examples/siro_sandbox/controllers/gui_controller.py
```diff
@@ -245,8 +247,16 @@ def get_random_joint_action(self):
...
+    def set_act_hints(
+        self,
+        walk_dir,
+        distance_multiplier,
+        grasp_obj_idx,
+        do_drop,
+        cam_yaw=None,
+    ):
         self._hint_walk_dir = walk_dir
+        self._hint_distance_multiplier = distance_multiplier
```
eundersander, line 259: Where is the logic that uses distance_multiplier?

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -250,11 +266,15 @@ def _update_help_text(self):
...
+    def _get_agent_pose(self):
         agent_root = get_agent_art_obj_transform(
             self.get_sim(), self.get_gui_controlled_agent_index()
         )
-        lookat = agent_root.translation + mn.Vector3(0, 1, 0)
+        return agent_root.translation, agent_root.rotation
+
+    def _get_camera_lookat_pos(self):
+        agent_pos, _ = self._get_agent_pose()
+        lookat = agent_pos + mn.Vector3(0, 1, 0)
```
eundersander, line 277: I know your PR didn't introduce this code, but it would be nice to replace this 1 magic number with a named constant like "pelvis_height" or "lookat_y_offset".

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -100,11 +100,86 @@ def viz_and_get_humanoid_walk_dir(self):
             path_points, path_endpoint_radius, path_color
         )
 
-        if len(path.points) >= 2:
-            walk_dir = mn.Vector3(path.points[1]) - mn.Vector3(path.points[0])
-            return walk_dir
+    def _get_humanoid_walk_dir_from_remote_gui_input(
```
eundersander, line 103: These are public methods so they shouldn't use leading underscore, right?

eundersander, line 103: I find the name "get_humanoid_walk_dir..." to be confusing given you are actually returning walk_dir plus another thing (distance_multiplier). How about "get_humanoid_walk_hints..." or similar?

eundersander (commented):
I left some comments.

examples/siro_sandbox/controllers/gui_controller.py
rpartsey: Oh, yes. I was experimenting with random quaternions and unintentionally rolled back the code with distance_multiplier.

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -100,11 +95,85 @@ def viz_and_get_humanoid_walk_dir(self):
...
+
+        if not ray or ray.direction.y >= 0 or ray.origin.y <= floor_y:
+            return None
+
+        dist_to_floor_y = (ray.origin.y - floor_y) / -ray.direction.y
+        target_on_floor = ray.origin + ray.direction * dist_to_floor_y
+
+        return target_on_floor
+
+    def _compute_forward_dir(self, target_rot_quat):
```
rpartsey, line 170: Used [this code](https://github.com/facebookresearch/habitat-lab/blob/929c0efea33a34a786e75430735c3da9f7bff233/habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py#L129) as an example.

examples/siro_sandbox/gui_navigation_helper.py
rpartsey: addressed

rpartsey: addressed

examples/siro_sandbox/app_states/app_state_fetch.py
rpartsey: addressed

examples/siro_sandbox/gui_navigation_helper.py
rpartsey: addressed

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -112,14 +112,14 @@ def _update_grasping_and_set_act_hints(self):
...
-                ) = self._nav_helper._get_humanoid_walk_dir_from_remote_gui_input(
+                ) = self._nav_helper.get_humanoid_walk_hints_from_remote_gui_input(
                     visualize_path=False
                 )
             else:
                 (
                     candidate_walk_dir,
                     candidate_distance_multiplier,
-                ) = self._nav_helper._get_humanoid_walk_dir_from_ray_cast(
+                ) = self._nav_helper.get_humanoid_walk_dir_from_ray_cast(
```
eundersander, line 122: why still "walk_dir"? Make sure to test these codepaths given we don't have CI hooked up yet!

rpartsey: done

eundersander (approved):
LGTM!

PR #1562, opened 2023-09-05
[hab3_merge] -  navmesh related refactor from SIRo to existing files

aclegg3:
## Motivation and Context

This PR contains remaining NavMesh related changes from SIRo to files which already exist in `main`.

## How Has This Been Tested

CI, no additional testing was done

## Types of changes

## Checklist


habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -457,24 +457,29 @@ def get_robot_spawns(
...
+        # Place within `distance_threshold` of the object.
+        start_position = sim.pathfinder.get_random_navigable_point_near(
+            target_position,
+            distance_threshold,
+            island_index=sim.largest_island_idx,
+        )
+        # get_random_navigable_point_near() can return NaNs for start_position.
+        # We want to make sure that the generated start_position is valid
+        if np.isnan(start_position).any():
+            continue
```
aclegg3, line 474: NOTE: this function is significantly refactored from SIRo. Seems like that code wasn't doing what it claimed to do.
Here we actually want to sample new points from the correct navmesh until we find a pose which is valid (not NaN, within distance threshold, non-colliding).

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -443,23 +448,42 @@ def _load_navmesh(self, ep_info):
...
-            self.pathfinder.build_navmesh_vertices(), axis=0
-        )
-        self._island_sizes = [
-            self.pathfinder.island_radius(p) for p in self._navmesh_vertices
-        ]
-        self._max_island_size = max(self._island_sizes)
-        self._largest_island_idx = self.pathfinder.get_island(
-            self._navmesh_vertices[np.argmax(self._island_sizes)]
+        if osp.exists(navmesh_path):
+            self.pathfinder.load_nav_mesh(navmesh_path)
```
vincentpierre, line 453: Is it guaranteed that the navmesh that is loaded here has the correct `agent_config.max_climb` and `agent_config.max_slope`? It seems `base_dir` is a scene id, so it could be shared between runs (of different configs) no ?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -443,23 +448,42 @@ def _load_navmesh(self, ep_info):
...
+            os.makedirs(osp.dirname(navmesh_path), exist_ok=True)
+            self.pathfinder.save_nav_mesh(navmesh_path)
+
+        # NOTE: allowing indoor islands only
+        self._largest_island_idx = get_largest_island_index(
+            self.pathfinder, self, allow_outdoor=False
         )
 
     @property
     def largest_island_idx(self) -> int:
```
vincentpierre, line 484: If it is not too much work, I would rename `largest_island_idx` to `largest_indoors_island_idx`.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -511,37 +535,37 @@ def _set_ao_states_from_ep(self, ep_info: RearrangeEpisode) -> None:
...
+        distances to it.
         """
-        new_pos = self.pathfinder.snap_point(pos)
-        island_radius = self.pathfinder.island_radius(new_pos)
+        new_pos = self.pathfinder.snap_point(pos, self._largest_island_idx)
+
+        max_iter = 10
+        offset_distance = 1.5
+        distance_per_iter = 0.5
+        num_sample_points = 1000
```
vincentpierre, line 554: That is a lot of parameters here. I would create a private const/static struct at the top of the script so it is easily accessible, or capitalize these so we know they are hard coded unchangeable magic numbers. But this is a question of **preference**. Feel free to ignore.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -511,37 +535,37 @@ def _set_ao_states_from_ep(self, ep_info: RearrangeEpisode) -> None:
...
 
-        if np.isnan(new_pos[0]) or island_radius != self._max_island_size:
-            # The point is not valid or not in a different island. Find a
-            # different point nearby that is on a different island and is
-            # valid.
+        regen_i = 0
+        while np.isnan(new_pos[0]) and regen_i < 10:
```
vincentpierre, line 557: Should this 10 also be a constant like the ones above ?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -511,37 +535,37 @@ def _set_ao_states_from_ep(self, ep_info: RearrangeEpisode) -> None:
...
-            ]
-            distances = np.linalg.norm(
-                np.array(pos).reshape(1, 3) - use_verts, axis=-1
-            )
-            closest_idx = np.argmin(distances)
-            new_pos = self._navmesh_vertices[closest_idx]
+
+        assert not np.isnan(
+            new_pos[0]
+        ), f"The snap position is NaN. scene_id: {self.ep_info.scene_id}, new position: {new_pos}, original position: {pos}"
```
vincentpierre, line 568: What is the recommendation when this happens ? 
Should the user change some configuration or modify a scene?

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -484,19 +489,6 @@ def get_robot_spawns(
         if physics_stability_steps == 0:
```
vincentpierre, line 489: Does this need to `sim.set_state(state)` before return ?

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -516,11 +508,10 @@ def get_robot_spawns(
             if did_collide:
                 break
 
+        sim.set_state(state)
```
vincentpierre, line 511: Is there a cost to calling `sim.set_state(state)` ? If that is the case, sim.set_state should be called only before returning and not inside this for loop. I think this way it will be called every iteration instead of just once.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
aclegg3: True, it is not guaranteed that the loaded navmesh parameters are correct. However, the user should be able to provide any navmesh they want without risking automated re-compute if a parameter is different. 
The automated re-compute is a backup in case the provided path does not exist. The warning should let the user know this is happening.

habitat-lab/habitat/tasks/rearrange/utils.py
aclegg3: This was done in the previous implementation, but I think we can remove that entirely. We can handle the robot state explicitly without saving/setting all the object states.

habitat-lab/habitat/tasks/rearrange/utils.py
aclegg3: This was used to reset the robot position before re-sampling I think. Probably not necessary.

aclegg3:
@vincentpierre Thanks for probing about the `get_robot_spawns` logic. I took a closer look and re-wrote that function. I think it makes more sense now.

xavierpuigf (approved):
LGTM!

rpartsey (approved):
LGTM

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
```diff
@@ -82,6 +82,9 @@ def __init__(
         self._episode_id: str = ""
         self._cur_episode_step = 0
         self._should_place_articulated_agent = should_place_articulated_agent
+        # support hab2 benchmark fixed states
+        if self._sim.habitat_config.agents["main_agent"].is_set_start_state:
+            self._should_place_articulated_agent = False
```
aclegg3, line 87: Note: this has probably been broken for awhile. Without this, all RearrangeTasks were overriding the Hab2 benchmark fixed states, invalidating the benchmark.

aclegg3: @ASzot @xavierpuigf check me here on the expected logic.
When this start state is provided in config the `ArticulatedAgentManager.post_obj_load_reconfigure` processes the configured values, so we should respect that and turn off forced re-placement of the agent.

aclegg3:
Note: The Hab2 interact benchmark is failing now after the fix. This is an existing bug in Habitat-lab which was exposed by the fix in [21d62b8199e43bccfacc773f4141de737a265a9d](https://github.com/facebookresearch/habitat-lab/pull/1562/commits/21d62b8199e43bccfacc773f4141de737a265a9d). It should be addressed externally to this PR.

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
ASzot: Yes agreed.

PR #1564, opened 2023-09-06
Fix episodes done count

rpartsey:
## Motivation and Context

Fixes episodes done count.

## How Has This Been Tested
Local testing MacBook Pro 2021.

## Types of changes

- **Refactoring** 
- **Bug Fix**

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -150,6 +150,9 @@ class SandboxDriver(GuiAppDriver):
         self._viz_anim_fraction = 0.0
         self._pending_cursor_style = None
 
+        self._num_iter_episodes: int = len(self.habitat_env.episode_iterator.episodes)  # type: ignore
+        self._num_episodes_done: int = 0
+
         def local_end_episode(do_reset=False):
             self._end_episode(do_reset)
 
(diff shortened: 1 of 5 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -261,8 +265,14 @@ def _env_step(self, action):
             self._metrics,
         ) = self.gym_habitat_env.step(action)
 
+    def get_num_episodes_done(self):
```
eundersander, line 268: I don't think these need to be public. And perhaps we don't need these getters at all? I.e. for constructing SandboxService, just do `lambda: self._num_episodes_done`?

examples/siro_sandbox/sandbox_service.py
```diff
@@ -88,3 +94,15 @@ def end_episode(self):
     @property
     def set_cursor_style(self):
         return self._set_cursor_style
+
```
eundersander, line 97: This is so verbose... three new members just to make this episode info available... I'm worried SandboxService is going to become very bloated. I guess this is fine for now. In the future, maybe we need to find a more elegant and scalable way to expose things from SandboxDriver to AppStates.

eundersander (approved):
A few comments, aside from those this looks good.

examples/siro_sandbox/sandbox_service.py
eundersander: If you're feeling ambitious: I suppose an EpisodeHelper class, that SandboxDriver owns, and then SandboxService exposes this object. So we only have one new member in SandboxService instead of three. Maybe that is cleaner.

rpartsey: done

PR #1566, opened 2023-09-08
[bugfix]: remove joint motors when converting objects to MotionType.KINEMATIC

aclegg3:
## Motivation and Context

JointMotors created for ArticulatedObjects stick around in the Bullet backend when the AO is removed from the simulation world (set to KINEMATIC). These motors undergo costly constrain row computation every physics step but do nothing. Instead, they should be removed for kinematic mode.

NOTE: This assumes we won't switch back to dynamic mode later (JointMotors would need to be re-created)

## How Has This Been Tested

Locally via benchmark.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
+++ b/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
@@ -11,7 +11,7 @@ import numpy as np
 from habitat.articulated_agents.articulated_agent_interface import (
     ArticulatedAgentInterface,
 )
-from habitat_sim.physics import JointMotorSettings
+from habitat_sim.physics import JointMotorSettings, MotionType
 from habitat_sim.simulator import Simulator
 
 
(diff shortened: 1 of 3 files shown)


habitat-lab/habitat/articulated_agents/articulated_agent_base.py
```diff
@@ -224,8 +224,10 @@ def leg_joint_pos(self, ctrl: List[float]):
 
             joint_positions = self.sim_obj.joint_positions
 
+            mt = self.sim_obj.motion_type
             for i, jidx in enumerate(self.params.leg_joints):
-                self._set_motor_pos(jidx, ctrl[i])
+                if mt == MotionType.DYNAMIC:
+                    self._set_motor_pos(jidx, ctrl[i])
```
ASzot, line 230: To confirm, the agent joint position will still update in Kinematic mode? Do we also need to have similar logic for the other articulated objects (like cabinets and fridges)? Or are the motors only instantiated on the robots?

aclegg3: > To confirm, the agent joint position will still update in Kinematic mode? 
Yes, the line outside the loop:
```
self.sim_obj.joint_positions = joint_positions
```
Does the kinematic position setting. The conditional line here sets the motor targets. These motors have already been removed for kinematic mode, so it would error out.

> Do we also need to have similar logic for the other articulated objects (like cabinets and fridges)?
The changes in rearrange_sim remove all joint motors for all articulated object instances in kinematic mode. Assuming the scene is already loaded and all the furniture is instantiated, that should remove all joint motors. If that assumption isn't valid we should follow-up.

ASzot (approved):

PR #1591, opened 2023-09-29
Unoccluded navmesh snap

aclegg3:
## Motivation and Context

**TL;DR:** This PR introduces a rejection sampling based approach to generate unoccluded snap points.

### Context: 
Snapping a target point to the navmesh is used in many ways throughout the rearrangement task including:
- as an accessibility check when generating object placements
- to generate nav location targets from which and agent should pick an object
- to check whether an object can be picked from an agent's position

However, standard navmesh snapping can result in a snapped point on the wrong side of obstacles (e.g. on the other side of a wall in another room) especially when snapping clutter object locations which sit off the navmesh on furniture. For example, a bowl on the back of a counter may actually be closer to the navmesh in the room sharing a wall with the counter than the kitchen floor. These error manifest in infeasible object placements and pick/place artifacts which hinder the realism of the task.

### New feature:

Adds the `unoccluded_navmesh_snap` and `snap_point_is_occluded` utility functions.
The new approach is slower than standard snapping, but still fast enough to be used within the generator loop.  

- `snap_point_is_occluded` - uses raycasting to check whether a provided point (assumed to be on the floor) has an unoccluded line to the target point. Uses a configurable height and search granularity to match agent constraints and compute limitations.
- `unoccluded_navmesh_snap` - uses `get_random_navigable_point_near` sim API call and rejection sampling with `snap_point_is_occluded` to find an approximation to the closest unoccluded snap point for a given target position. Relies on [sim PR 2221](https://github.com/facebookresearch/habitat-sim/pull/2221) for accuracy. See test section below for examples of the variance using this approximation.

### NOTE:

TODO: This PR currently includes changes in the generator to produce better episodes, but does not refactor rearrange_task or rearrange_sim to use the new functions for snapping.

## How Has This Been Tested

Local episode generation and interactive testing application.

Examples of different results returned by `unoccluded_navmesh_snap` to demonstrate approximation variance. This variance is decreased by increasing max samples and decreasing minimum distance between samples (at the cost of compute):
unoccluded_snap_1

unoccluded_snap_2

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
+++ b/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
@@ -44,12 +44,17 @@ def snap_point_is_occluded(
         if (
             raycast_results.has_hits()
             and raycast_results.hits[0].ray_distance < 1
-            and target_object_id is not None
         ):
-            if raycast_results.hits[0].object_id == target_object_id:
+            if (
+                target_object_id is not None
+                and raycast_results.hits[0].object_id == target_object_id
+            ):
+                # we hit an allowed object (i.e., the target object), so not occluded
                 return False
+            # the ray hit a not-allowed object and is occluded
             continue
         else:
+            # ray hit nothing, so not occluded
             return False
     return True
 
--- a/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
+++ b/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
@@ -10,6 +10,137 @@ from habitat.tasks.utils import get_angle
 from habitat_sim.physics import VelocityControl
 
 
+def snap_point_is_occluded(
+    target: mn.Vector3,
+    snap_point: mn.Vector3,
+    height: float,
+    sim: habitat_sim.Simulator,
+    granularity: float = 0.2,
+    target_object_id: Optional[int] = None,
+) -> bool:
+    """
+    Uses raycasting to check whether a target is occluded given a navmesh snap point.
+
+    :property target: The 3D position whic should be unoccluded from the snap point.
+    :property snap_point: The navmesh snap point under consideration.
+    :property height: The height of the agent. Given navmesh snap point is grounded, the maximum height from which a visibility check should indicate non-occlusion. First check starts from this height.
+    :property sim: The Simulator instance.
+    :property granularity: The distance between raycast samples. Finer granularity is more accurate, but more expensive.
+    :property target_object_id: An optional object id which should be ignored in occlusion check.
+
+    NOTE: If agent's eye height is known and only that height should be considered, provide eye height and granulatiry > height for fastest check.
+
+    :return: whether or not the target is considered occluded from the snap_point.
+    """
+    # start from the top, assuming the agent's eyes are not at the bottom.
+    cur_height = height
+    while cur_height > 0:
+        ray = habitat_sim.geo.Ray()
+        ray.origin = snap_point + mn.Vector3(0, cur_height, 0)
+        cur_height -= granularity
+        ray.direction = target - ray.origin
+        raycast_results = sim.cast_ray(ray)
+        # distance of 1 is the displacement between the two points
+        if (
+            raycast_results.has_hits()
+            and raycast_results.hits[0].ray_distance < 1
+            and target_object_id is not None
+        ):
+            if raycast_results.hits[0].object_id == target_object_id:
+                return False
+            continue
+        else:
+            return False
+    return True
+
+
+def unoccluded_navmesh_snap(
+    pos: mn.Vector3,
+    height: float,
+    pathfinder: habitat_sim.nav.PathFinder,
+    sim: habitat_sim.Simulator,
+    target_object_id: Optional[int] = None,
+    island_id: int = -1,
+    search_offset: float = 1.5,
+    test_batch_size: int = 20,
+    max_samples: int = 200,
+    min_sample_dist: float = 0.5,
+) -> Optional[mn.Vector3]:
+    """
+    Snap a point to the navmesh considering point visibilty via raycasting.
+
+    :property pos: The 3D position to snap.
+    :property height: The height of the agent. Given navmesh snap point is grounded, the maximum height from which a visibility check should indicate non-occlusion. First check starts from this height.
+    :property pathfinder: The PathFinder defining the NavMesh to use.
+    :property sim: The Simulator instance.
+    :property target_object_id: An optional object_id which should be ignored in the occlusion check. For example, when pos is an object's COM, that object should not occlude the point.
+    :property island_id: Optionally restrict the search to a single navmesh island. Default -1 is the full navmesh.
+    :property search_offset: The additional radius to search for navmesh points around the target position. Added to the minimum distance from pos to navmesh.
+    :property test_batch_size: The number of sample navmesh points to consider when testing for occlusion.
+    :property max_samples: The maximum number of attempts to sample navmesh points for the test batch.
+    :property min_sample_dist: The minimum allowed L2 distance between samples in the test batch.
+
+    NOTE: this function is based on smapling and does not guarantee the closest point.
+
+    :return: An approximation of the closest unoccluded snap point to pos or None if an unoccluded point could not be found.
+    """
+
+    # first try the closest snap point
+    snap_point = pathfinder.snap_point(pos, island_id)
+    is_occluded = snap_point_is_occluded(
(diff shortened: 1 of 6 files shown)


habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+    snap_point: mn.Vector3,
+    height: float,
+    sim: habitat_sim.Simulator,
+    granularity: float = 0.2,
+    target_object_id: Optional[int] = None,
+) -> bool:
+    """
+    Uses raycasting to check whether a target is occluded given a navmesh snap point.
+
+    :property target: The 3D position whic should be unoccluded from the snap point.
```
jimmytyyang, line 24: typo: which

jimmytyyang: And I imagine this will be gripper's EE location?

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+    :property snap_point: The navmesh snap point under consideration.
```
jimmytyyang, line 25: And this will be the robot's current location? `base_pos`

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+
+    NOTE: If agent's eye height is known and only that height should be considered, provide eye height and granulatiry > height for fastest check.
+
+    :return: whether or not the target is considered occluded from the snap_point.
+    """
+    # start from the top, assuming the agent's eyes are not at the bottom.
+    cur_height = height
+    while cur_height > 0:
+        ray = habitat_sim.geo.Ray()
+        ray.origin = snap_point + mn.Vector3(0, cur_height, 0)
```
jimmytyyang, line 39: I think this boils down to where the camera's location is. So Spot, we will need to offset the `base_pos` to the front and then set the height to be ee's height?

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+        cur_height -= granularity
+        ray.direction = target - ray.origin
+        raycast_results = sim.cast_ray(ray)
+        # distance of 1 is the displacement between the two points
+        if (
+            raycast_results.has_hits()
+            and raycast_results.hits[0].ray_distance < 1
```
jimmytyyang, line 46: index zero means this is the first hit?

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+            and target_object_id is not None
+        ):
+            if raycast_results.hits[0].object_id == target_object_id:
```
jimmytyyang, line 49: Will this be a little bit too hard? Since that if the object is placed in the location that there is a cover on top of it (e.g., shelf), then the robot needs to lower the arm until it sees it? Not sure in this case, how should we treat that?

jimmytyyang: Or something like `norm(raycast_results.hits[0].translation - target_object_translaton) <= threshold`, then we set `true`?

jimmytyyang (commented):
This is such an awesome design!!! Thank you so much for this, and the code makes sense to me. I just have a few design question and how to use the function. Thank you so much, Alex!

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: yes, hits are sorted by distance

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: target is the location you want visibility to (e.g. the target object's location).

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: This is a navmesh snapped point. So, yes it could be the robot's base position.

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: Yeah, exactly. Height should be the maximum height where the robot hand/camera can have valid visibility to the object. Max EE camera height may be too much (e.g. the robot can't grasp an object it can only see with it's arm stretched all the way up).

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: In general this function checks all the heights from configured max to floor, so it gives you a single binary for whether there exists a visible height or not. 
If you want to check a specific height, then pass that one and set the granularity >= to height. You could then get a binary flag for whether the current state allow visibility.

jimmytyyang (approved):
LGTM! thank you so much for working on this. This looks nice!

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,137 @@
...
+                return False
+            continue
+        else:
```
jimmytyyang, line 52: Hi Alex,

I tried this function and discovered some design questions. I will talk about this during our meeting! Thanks a lot!

jimmytyyang (changes requested):

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
jimmytyyang: proposed a minor change, we can discuss more during the meeting. thanks Alex!

https://github.com/facebookresearch/habitat-lab/pull/1599

aclegg3: Merged our changes into this branch.

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
```diff
@@ -10,6 +10,142 @@
...
+
+    :return: whether or not the target is considered occluded from the snap_point.
+    """
+    # start from the top, assuming the agent's eyes are not at the bottom.
+    cur_height = height
+    while cur_height > 0:
+        ray = habitat_sim.geo.Ray()
+        ray.origin = snap_point + mn.Vector3(0, cur_height, 0)
+        cur_height -= granularity
+        ray.direction = target - ray.origin
```
jimmytyyang, line 41: Hi Alex,

After a test, I found that the ray pointing to target (i.e., target object) from a `snap_point` location usually has 0.98 distance value, which is close to one. I think this is because that the target object is usually on the ground/floor, and the first thing it hits could be something else. So, I propose to change this to `ray.direction =  ray.origin - target`. Theoretically, these two are the same thing if there is nothing in between, but this offers much predicted results since we are casting a ray from the target object to the ceiling

jimmytyyang: Some visual about this proposed change. Red circle is the resulting snap point found
Screenshot 2023-10-04 at 9 25 01 AM
ookresearch/habitat-lab/assets/55121504/44ddb51d-2597-4c3a-8bf3-c97e9d3a6396">
esearch/habitat-lab/assets/55121504/9e8b4f40-5d4a-48f9-a7cd-329970d539d7">

jimmytyyang (commented):
Thanks for the discussion, Alex! Please let me know what you think. After this fix, I do not think we need to have a threshold thing anymore. Happy to chat :)

habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
aclegg3: Agreed that the ray direction does not matter too much theoretically.

However, there is risk with starting the ray inside the target. We expect the snap_point is in open space (because snap was possible), but we don't know about target. Starting a raycast inside a convex collision shape will miss that shape. This seems like it could introduce errors and I'm not sure what the gain here would be.

jimmytyyang (approved):
LGTM! thanks alex!!!

PR #1593, opened 2023-09-30
Enable HBAO in the sandbox tool.

0mdc:
## Motivation and Context

This enables HBAO in the sandbox tool.
Depends on: https://github.com/facebookresearch/habitat-sim/pull/2223

HITL_HBAO

## How Has This Been Tested

Tested locally.

### DIFF
--- a/habitat-lab/habitat/gui/replay_gui_app_renderer.py
+++ b/habitat-lab/habitat/gui/replay_gui_app_renderer.py
@@ -59,6 +59,8 @@ class ReplayGuiAppRenderer(GuiAppRenderer):
         # TODO:  Reenable frustum culling when replay renderer issues are solved.
         if hasattr(cfg, "enable_frustum_culling"):
             cfg.enable_frustum_culling = False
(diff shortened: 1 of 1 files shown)


rpartsey (approved):

xavierpuigf (approved):
Looks good here! Do you have suggestions for how to do this generally for habitat-lab? Should we have a flag to turn on or off hbao? Why do we care about giving this option? Are things slower?

jimmytyyang (approved):
LGTM!

0mdc:
@xavierpuigf 

>Do you have suggestions for how to do this generally for habitat-lab?

`SimulatorConfiguration` now has a `enable_hbao` flag. I believe that the correct approach would be to add a `enable_hbao` configuration flag in structured configs, then pass it to `habitat-sim` in `habitat_simulator.py`'s `create_sim_config`.

>Why do we care about giving this option? Are things slower?

HBAO is great for creating visuals for humans, but it's unlikely to help trained agents converge faster. It does have a performance impact, albeit not very large. I would not enable this feature when training.

PR #1599, opened 2023-10-03
Handle the case if there is no object_id

jimmytyyang:


aclegg3:
Closing because we merged this fix into original PR branch.

PR #1602, opened 2023-10-05
Add begging state near human

jimmytyyang:


jimmytyyang:
Hi @eundersander , I took at this line 

```
                    is_accessible = (
                        float(
                            np.linalg.norm(
                                np.array((target_trans - obj_trans))[[0, 2]]
                            )
                        )
                        < IS_ACCESSIBLE_THRESHOLD
                    )
```
and I was trying to use `length()` in magnum as you suggested to compute the norm, but I need to do something like target_trans[1] = 0 and obj_trans[1] = 0 to project the vector. I am not sure about a better way to do that. Thank you!

jimmytyyang:
Thanks, I will merge this PR by Monday since I want to confirm with Alex about the navmesh height stuff.

PR #1606, opened 2023-10-10
Current episode reset

rpartsey:
## Motivation and Context

Adds current episode reset functionality ([Asana task](https://app.asana.com/0/1205684818278379/1205617988863636)). 
- M to go to next episode (new episode with new scene), with cycling
- X to restart (same exact episode). Note, that at reset agents positions are set randomly (this is Habitat-lab reset works).

## How Has This Been Tested
Launching SandboxApp locally on MacBook Pro 2021.

https://github.com/facebookresearch/habitat-lab/assets/37709869/77ef1939-dcc2-4423-8eb9-722b5b93a246

First part of the video - current episode reset (X), second part of the video - new episode with new scene reset (M).

## Types of changes

- **Development:** PR into  `eundersander/SIRo_hitl_Sep5_demo_refactored`

## Checklist

### DIFF
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -92,7 +92,9 @@ class AppStateFetch(AppState):
         return self._is_remote_active_toggle
 
     def on_environment_reset(self, episode_recorder_dict):
-        self._current_episode = self._sandbox_service.env.current_episode
+        self._current_episode = (
+            self._sandbox_service.episode_helper.current_episode
+        )
         self._held_target_obj_idx = None
 
         sim = self.get_sim()
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -17,11 +17,7 @@ from gui_avatar_switch_helper import GuiAvatarSwitchHelper
 from gui_navigation_helper import GuiNavigationHelper
 from gui_pick_helper import GuiPickHelper
 from gui_throw_helper import GuiThrowHelper
-from hablab_utils import (
-    get_agent_art_obj_transform,
-    get_articulated_agent,
-    get_grasped_objects_idxs,
-)
+from hablab_utils import get_agent_art_obj_transform, get_grasped_objects_idxs
 
 from habitat.gui.gui_input import GuiInput
 from habitat.gui.text_drawer import TextOnScreenAlignment
@@ -54,7 +50,6 @@ class AppStateFetch(AppState):
         self._cam_transform = None
 
         self._current_episode = None
-        self._init_agent_states = None
         self._held_target_obj_idx = None
         self._held_hand_idx = None  # currently only used with remote_gui_input
         self._recent_reach_pos = None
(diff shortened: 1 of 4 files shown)


examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -742,13 +763,43 @@ def _viz_fetcher(self, post_sim_update_dict):
...
+
+        # reset current episode
+        # env's current_episode setter will set internal attribute
+        # _episode_from_iter_on_reset to False as a result reset
+        # will be called without the episode changing
+        self._sandbox_service.env.current_episode = self._current_episode
+        self._sandbox_service.end_episode(do_reset=True)
+
+        # restore initial agent states
+        sim = self.get_sim()
```
eundersander, line 779: I don't want a custom reset codepath in the HITL codebase. It would be tedious to maintain. We should find a way to re-use habitat-lab episode-resetting (which should correctly initialize articulated objects and everything else in the scene).

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -742,13 +763,43 @@ def _viz_fetcher(self, post_sim_update_dict):
...
+        for agent_idx in range(self.get_num_agents()):
+            articulated_agent = get_articulated_agent(sim, agent_idx)
+            prev_pos, prev_rot = prev_init_agent_states[agent_idx]
+            articulated_agent.base_pos = prev_pos
+            articulated_agent.base_rot = float(prev_rot)
+        self._init_agent_states = prev_init_agent_states
+
+    def _set_next_episode(self):
```
eundersander, line 787: I don't understand this helper.

eundersander (commented):
I'd like to find a cleaner way to do this. See comment below.

examples/siro_sandbox/app_states/app_state_fetch.py
rpartsey: The idea is instead of having the code written under 
```
if self._sandbox_service.gui_input.get_key_down(GuiInput.KeyNS.<Any key>):
    # code
``` 
wrap the code in a method and give it a meaningful name. 

For example, having  `self._set_next_episode()` is (IMO) more meaningful than `self._sandbox_service.end_episode(do_reset=True)`. 

But I can roll this back.

eundersander: Ok. I was just thrown by the name. So it means "end the episode and start the next episode". I guess I would call it `_end_episode_and_start_next`.

examples/siro_sandbox/episode_helper.py
```diff
@@ -2,12 +2,17 @@ class EpisodeHelper:
...
+        self._num_iter_episodes: int = len(habitat_env.episode_iterator.episodes)  # type: ignore
         self._num_episodes_done: int = 0
 
     @property
     def num_iter_episodes(self):
-        return self._num_iter_episodes
+        return (
+            self._num_iter_episodes
+            if not self._episode_iterator_cycle
+            else float("inf")
```
eundersander, line 14: -1 would make more sense than a float here.

examples/siro_sandbox/configs/fetch.yaml
```diff
@@ -112,6 +112,10 @@ habitat:
 
   environment:
     max_episode_steps: 1000
+    iterator_options:
```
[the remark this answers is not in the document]
eundersander: This is fine. I think Mikael or I will be looking to set up dedicated hotkeys to change episodes within a scene or skip to an episode for a different scenes, but I still think we can merge this PR pretty much as-is (aside from my other nit comments in this review).

eundersander (approved):
A few small requests but otherwise LGTM!

eundersander:
I'm currently cherry-picking these commits into a new PR to offer scene selection in the tool, so let's not merge this right now.

examples/siro_sandbox/episode_helper.py
rpartsey: In other app states we display num episodes left help text. It outputs num_iter_episodes - num_episodes_done. If we set -1 then we will be getting negative number,  with `float("inf")` inf will be displayed. IMHO it is more meaningful.

PR #1609, opened 2023-10-10
Eundersander/si ro hitl sep5 demo refactored humanoid multidir

xavierpuigf:
## Motivation and Context
Adds fixed to the humanoid walking and reaching motion, reducing jitter and making it able to walk backwards.

## How Has This Been Tested

In unity and desktop tool

## Types of changes

## Checklist


examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -247,21 +258,29 @@ def _update_grasping_and_set_act_hints_remote(self):
...
         (
             walk_dir,
             distance_multiplier,
+            forward_dir,
         ) = self._nav_helper.get_humanoid_walk_hints_from_remote_gui_input(
             visualize_path=False
         )
 
         # If the humanoid is near enough to the VR user that it doesn't need to walk
         # (distance_multiplier == 0.0), then try to reach to held object.
```
eundersander, line 273: This comment looks outdated.

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -167,13 +174,28 @@ def _get_humanoid_walk_hints(
                 if visualize_path:
                     self._viz_path(path)
             else:
-                walk_dir = None
+                # TODO: potentially check for previous rotation
```
eundersander, line 177: I don't understand this code now. The "else" code is almost identical to the "if" code (setting distance_multiplier to 1.0), with the only difference being that we don't call _viz_path for the else code. Can you add some comments explaining what you're trying to do here?

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -167,13 +174,28 @@ def _get_humanoid_walk_hints(
...
+        if (
+            found_path
+            and len(path.points) >= 2
+            and path.geodesic_distance >= dist_to_always_move_forward
+        ):
+            target_rot_quat = None
+
+        if target_rot_quat is not None:
+            # Get the forward direction
+            forward_gaze = target_rot_quat.transform_vector(
```
eundersander, line 194: This looks almost identical to the code in `_compute_forward_dir`, and I wonder if the result `forward_gaze` is identical to `forward_dir` elsewhere in this function. Can you refactor to avoid code duplication?

examples/siro_sandbox/gui_navigation_helper.py
```diff
@@ -167,13 +174,28 @@ def _get_humanoid_walk_hints(
...
+                mn.Vector3(0, 0, 1)
+            )
+            forward_gaze.y = 0
+        return walk_dir, distance_multiplier, forward_gaze
```
eundersander, line 198: Why do we need `walk_dir` *and* `forward_gaze`? Aren't these basically the same thing? They are a vector in the ground plane indicating the direction the agent is facing, right?

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -162,7 +162,10 @@ def calculate_turn_pose(self, target_position: mn.Vector3):
         self.calculate_walk_pose(target_position, distance_multiplier=0)
 
     def calculate_walk_pose(
-        self, target_position: mn.Vector3, distance_multiplier=1.0
+        self,
+        target_position: mn.Vector3,
+        distance_multiplier=1.0,
+        target_dir=None,
```
eundersander, line 168: Can you update the function comment explaining what `target_dir` does? It seems nearly identical to `target_position` in terms of how its used inside this function.

eundersander (commented):
I left some comments. Let me know when I should take another look!

examples/siro_sandbox/gui_navigation_helper.py
xavierpuigf: Not exactly. When we move backwards, the forward gaze and the walk_dir will be opposite of each other. The forward gaze is where we should be looking whereas forward dir is where we should be walking towards. I can add a comment

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
xavierpuigf: Sounds good! See my comment above. I will update the function with a comment too

examples/siro_sandbox/gui_navigation_helper.py
xavierpuigf: That's right. There is not reason for this if / else. Will clean up. The reason I am changing this is that now we are allowing backward paths

examples/siro_sandbox/app_states/app_state_fetch.py
xavierpuigf: Addressed

eundersander (approved):
LGTM!

PR #1610, opened 2023-10-10
[Hab3] Planner Agent

ASzot:
## Motivation and Context

Depends on [this PR](https://github.com/facebookresearch/habitat-lab/pull/1524).

Adds the `Plan-Pop` multi-agent baseline and associated config to run this baseline. Also adds info to the README about how to run the multi-agent experiments.

## How Has This Been Tested

Training runs. 

## Types of changes


habitat-baselines/habitat_baselines/config/multi_agent/plan_pop.yaml
```diff
@@ -0,0 +1,139 @@
+# @package _global_
```
ASzot, line 1: Ideally this could derive from the `pop_play` config, but I couldn't get Hydra to correctly override the `/habitat_baselines/rl/policy@habitat_baselines.rl.policy.agent_1: hl_neural_multi_agent` entry. If I used the `override` keyword, running training would fail due to Hydra error. If I don't use the `override` keyword, the Hydra tests fail.

habitat-baselines/habitat_baselines/config/multi_agent/pop_play.yaml
```diff
@@ -18,23 +18,14 @@ defaults:
   - _self_
 
 
-habitat:
```
xavierpuigf, line 21: We set joint_noise to 0 in our experiments. Are we setting this to 0 elsewhere?

habitat-baselines/habitat_baselines/rl/hrl/hl/planner_policy.py
```diff
@@ -0,0 +1,316 @@
...
+        for path in paths:
+            all_ac_seqs.append([node.action for node in path])
+        # Sort by the length of the action sequence
+        full_plans = sorted(all_ac_seqs, key=len)
+        # Each full plan will be a permutation of the other full plans.
+        plans = full_plans[0]
+        return plans
+
+    def _replan(self, pred_vals, plan_idx: int):
+        if self._select_random_goal:
```
xavierpuigf, line 227: Do we specify somewhere what select_random_Goal means? It seems like we should set it to True in all our exps?

habitat-baselines/habitat_baselines/rl/hrl/hl/planner_policy.py
```diff
@@ -0,0 +1,316 @@
...
+            # We select a plan at random
+            index_plan = plan_idx
```
akshararai, line 229: how is this a random plan? Seems like the input plan index right?

akshararai (approved):
very minor comments, looks good overall.

habitat-baselines/habitat_baselines/rl/hrl/hl/planner_policy.py
ASzot: Now I updated `hl_planner.yaml` to detail this. Also I think it should be False in evaluation to evaluate against specific plans?

habitat-baselines/habitat_baselines/rl/hrl/hl/planner_policy.py
ASzot: The plan index is randomly generated. I renamed this variable to be clearer.

PR #1621, opened 2023-10-13
Add server->client messaging to the sandbox app + object highlights.

0mdc:
## Motivation and Context

This changeset adds the capability to send custom messages from the server to the client.

This works by extending the `gfx-replay` keyframe with a custom `message` dict. This dict can be mutated via the `MessagingService` class in the sandbox app.

A first message is implemented, which adds object highlights. Here's the feature shown from the Unity client:
image
Video:
https://github.com/facebookresearch/habitat-lab/assets/110583667/ce35f6e3-ab47-49a2-ac85-1b746b610fc8

## How Has This Been Tested

Tested locally.

### DIFF
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -514,16 +514,12 @@ class AppStateFetch(AppState):
             )
 
     def _get_target_object_position(self, target_obj_idx):
-        sim = self.get_sim()
-        rom = sim.get_rigid_object_manager()
-        object_id = self._target_obj_ids[target_obj_idx]
-        return rom.get_object_by_id(object_id).translation
+        return self._get_target_rigid_object(target_obj_idx).translation
 
     def _get_target_object_bounding_box(self, target_obj_idx) -> mn.Range3D:
-        sim = self.get_sim()
-        rom = sim.get_rigid_object_manager()
-        object_id = self._target_obj_ids[target_obj_idx]
-        return rom.get_object_by_id(object_id).collision_shape_aabb
+        return self._get_target_rigid_object(
+            target_obj_idx
+        ).collision_shape_aabb
 
     def _get_target_object_positions(self):
         sim = self.get_sim()
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -557,14 +557,10 @@ class AppStateFetch(AppState):
             )
             # color = mn.Color4(color.r, color.g, color.b, 1.0 - self._sandbox_service.get_anim_fraction())
 
-        num_segments = 24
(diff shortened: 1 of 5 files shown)


examples/siro_sandbox/server/messaging_service.py
```diff
@@ -0,0 +1,42 @@
...
+        assert pos
+        assert len(pos) == 3
+
+        if "highlights" not in self._message:
+            self._message["highlights"] = []
+        self._message["highlights"].append(
+            {"t": [pos[0], pos[1], pos[2]], "r": radius}
+        )
+
+    def add_message_to_keyframe(self, keyframe_obj) -> None:
```
eundersander, line 36: It's probably cleaner if this class doesn't know about keyframes. Instead, SandboxDriver decides how to append the message dict to the keyframe.

examples/siro_sandbox/server/messaging_service.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Meta Platforms, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict, List
+
+
+class MessagingService:
```
eundersander, line 10: I wish we had a more meaningful name. Maybe ClientMessagingService? To convey that this is particularly for sending messages to the remote client. And maybe ClientMessageManager? I chose "SandboxService" to perhaps convey that it doesn't particularly own or manage the things it offers, whereas clearly this class owns and manages these messages.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -528,6 +534,7 @@ def sim_update(self, dt):
                 obj = json.loads(keyframe_json)
                 assert "keyframe" in obj
                 keyframe_obj = obj["keyframe"]
+                self._messaging_service.add_message_to_keyframe(keyframe_obj)
```
eundersander, line 537: I don't like that our "keyframe" term is now ambiguous (it means a gfx-replay keyframe OR it means that PLUS messages). But we can probably refactor this later.

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -519,6 +519,12 @@ def _get_target_object_position(self, target_obj_idx):
         object_id = self._target_obj_ids[target_obj_idx]
         return rom.get_object_by_id(object_id).translation
 
+    def _get_target_object_bounding_box(self, target_obj_idx) -> mn.Range3D:
```
eundersander, line 522: I suppose this and `_get_target_object_position` could both use `_get_target_rigid_object` to avoid code duplication.

eundersander (approved):
If you are okay doing my suggested renames, then otherwise this LGTM. Let me know if you want to discuss more.

PR #1623, opened 2023-10-13
Sync client player on episode change.

0mdc:
## Motivation and Context

This sends the position of the humanoid to the client to that it can reposition the VR controller.
This avoid spawning the VR controller outside of the scene, or forcing the humanoid to catch back the controller at the start of an episode.

## How Has This Been Tested

Tested locally.

### DIFF
--- a/examples/siro_sandbox/server/client_message_manager.py
+++ b/examples/siro_sandbox/server/client_message_manager.py
@@ -38,3 +38,10 @@ class ClientMessageManager:
         self._message["highlights"].append(
             {"t": [pos[0], pos[1], pos[2]], "r": radius}
         )
+
+    def change_humanoid_position(self, pos: List[float]) -> None:
+        r"""
+        Change the position of the humanoid.
+        Used to synchronize the humanoid position in the client when changing scene.
+        """
+        self._message["humanoidPosition"] = [pos[0], pos[1], pos[2]]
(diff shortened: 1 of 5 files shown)


examples/siro_sandbox/server/remote_gui_input.py
```diff
@@ -260,3 +260,6 @@ def update(self):
 
     def on_frame_end(self):
         self._gui_input.on_frame_end()
+
+    def on_reset(self):
```
eundersander, line 264: `on_reset` is an odd name. I wouldn't expect a low-level input class like this to know about RL concepts like episode reset. How about just `clear_history`.

examples/siro_sandbox/server/client_message_manager.py
```diff
@@ -38,3 +38,10 @@ def add_highlight(self, pos: List[float], radius: float) -> None:
         self._message["highlights"].append(
             {"t": [pos[0], pos[1], pos[2]], "r": radius}
         )
+
+    def change_humanoid_position(self, pos: List[float]) -> None:
+        r"""
+        Change the position of the humanoid.
+        Used to synchronize the humanoid position in the client when changing scene.
+        """
+        self._message["humanoidPosition"] = [pos[0], pos[1], pos[2]]
```
eundersander, line 47: How about a better name like "forceAvatarBasePosition" (player? user?). "Humanoid" is overly specific, and I want "base" because you are specifying a location for the base (the feet). (We can imagine in the future, for multi-story houses, the y coord will be more significant.)

eundersander (approved):
A few naming requests, otherwise LGTM!

0mdc:
I'll merge this and polish if time permits. Two issues remain:
* It would be great to teleport the avatar when connection is established between the client and server.
* The HITL humanoid navigates to the old avatar position while the scene is loading on the client.

PR #1624, opened 2023-10-15
The robot stops to move toward human if the human is near the target AND the robot is too closed to human + agents starting locations

jimmytyyang:


jimmytyyang:
Need to test this PR on the demo scenes

jimmytyyang:
text
Screenshot 2023-10-15 at 9 16 02 PM

jimmytyyang:
scene key 1
Screenshot 2023-10-16 at 1 01 29 PM

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -99,7 +128,37 @@ def __init__(self, sandbox_service, gui_agent_ctrl, robot_agent_ctrl):
...
+    def _make_robot_near_human(self, human_pos):
+        """Constrain the robot location to be near human pos"""
+        robot_pos = self.get_sim().pathfinder.get_random_navigable_point_near(
+            human_pos,
+            ROBOT_SPAWN_NEAR_HUMAN_DIS,
+            island_index=self.get_sim()._largest_island_idx,
+        )
+
+        # If the robot_pos is not NaN
+        if not np.isnan(robot_pos[0]):
```
eundersander, line 147: Should we assert here? Don't we want to know if this fails?

examples/siro_sandbox/sandbox_app.py
```diff
@@ -363,7 +363,10 @@ def _reset_environment(self):
 
         # Reset all the app states
         for app_state in self._app_states:
-            app_state.on_environment_reset(self._episode_recorder_dict)
+            app_state.on_environment_reset(
+                self._episode_recorder_dict,
+                self._episode_helper.current_episode.scene_id,
```
eundersander, line 368: This shouldn't be necessary to add this argument. From AppStateFetch, have you tried sandbox_server.episode_helper.current_episode?

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -365,20 +367,31 @@ def _set_height(self):
         # Update the base
         self.get_articulated_agent().sim_obj.transformation = target_trans
 
-    def _start_fetch(self, human_pos, obj_pos):
-        """Start to fetch the object condition"""
+    def _human_unblock_robot(self, human_pos, robot_pos, obj_pos):
```
eundersander, line 370: Can you explain this function more? Like, what does it mean if it returns True? Can you name it something descriptive like "_is_robot_blocked_by_human"?

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -365,20 +367,31 @@ def _set_height(self):
...
-            np.linalg.norm(self.rigid_obj_interest.linear_velocity)
-            < START_FETCH_OBJ_VEL_THRESHOLD
-            and float(np.linalg.norm(np.array((human_pos - obj_pos))[[0, 2]]))
+            float(np.linalg.norm(np.array((human_pos - obj_pos))[[0, 2]]))
             > START_FETCH_OBJ_DIS_THRESHOLD
+            or float(np.linalg.norm(np.array((human_pos - robot_pos))[[0, 2]]))
+            > START_FETCH_ROBOT_DIS_THRESHOLD
+        )
+
+    def _start_fetch(self, human_pos, robot_pos, obj_pos):
```
eundersander, line 379: Can you explain this more? It looks like a boolean function that is checking something, whereas "start_fetch" sounds like you would call it to start fetching. Can you name it "_is_ok_to_start_fetch" or similar?

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -508,6 +521,13 @@ def act(self, obs, env):
                         self.current_state = FetchState.PICK
                     self._init_policy_input()
 
+            # Check if the human blocks the robot when robot is near the target
+            self.human_block_robot_when_searching = (
```
eundersander, line 525: I can't find where  you're using `self._human_block_robot_when_searching`. Is this dead code?

eundersander (commented):
I left some comments!

examples/siro_sandbox/app_states/app_state_fetch.py
jimmytyyang: `assert` here is risky since the whole app crashes. The logic here is that we did our best to find a good robot pos.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: maybe searching for `self.human_block_robot_when_searching` ?

PR #1626, opened 2023-10-16
Addressing font scaling issue

rpartsey:
## Motivation and Context

Addressing font scaling issue as Mosra suggested [here](https://cvmlp.slack.com/archives/CFN5TAUSD/p1697449064836769?thread_ts=1697296903.351319&cid=CFN5TAUSD). Corresponds to [this](https://app.asana.com/0/home/1205684818278375/1205683322171498) Asana task.

## How Has This Been Tested
Local testing on MacBook Pro 2021 with second monitor connected. See photos below:
photo_2023-10-16 14 54 51 

photo_2023-10-16 14 54 47

## Types of changes

- **Bug Fix**

## Checklist

### DIFF
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -791,7 +791,7 @@ if __name__ == "__main__":
         )
 
     # TextDrawer kwargs
-    display_font_size = 20
+    display_font_size = 40
     font_size_multiplier = (
         mn.Vector2(gui_app_wrapper.framebuffer_size)
         * gui_app_wrapper.dpi_scaling
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -793,9 +793,9 @@ if __name__ == "__main__":
     # TextDrawer kwargs
     display_font_size = 20
     font_size_multiplier = (
-        gui_app_wrapper.framebuffer_size
+        mn.Vector2(gui_app_wrapper.framebuffer_size)
         * gui_app_wrapper.dpi_scaling
-        / gui_app_wrapper.window_size
+        / mn.Vector2(gui_app_wrapper.window_size)
     ).max()
     text_drawer_kwargs = {
         "display_font_size": int(display_font_size * font_size_multiplier),
--- a/examples/siro_sandbox/sandbox_app.py
+++ b/examples/siro_sandbox/sandbox_app.py
@@ -790,15 +790,23 @@ if __name__ == "__main__":
             ),
         )
 
+    # TextDrawer kwargs
+    display_font_size = 20
+    font_size_multiplier = (
+        gui_app_wrapper.framebuffer_size
+        * gui_app_wrapper.dpi_scaling
+        / gui_app_wrapper.window_size
+    ).max()
+    text_drawer_kwargs = {
+        "display_font_size": int(display_font_size * font_size_multiplier),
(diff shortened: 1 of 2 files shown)


examples/siro_sandbox/sandbox_app.py
```diff
@@ -790,15 +790,23 @@ def _parse_debug_third_person(args, framebuffer_size):
             ),
         )
 
+    # TextDrawer kwargs
+    display_font_size = 20
+    font_size_multiplier = (
+        gui_app_wrapper.framebuffer_size
+        * gui_app_wrapper.dpi_scaling
+        / gui_app_wrapper.window_size
```
mosra, line 798:
```suggestion
        mn.Vector2(gui_app_wrapper.framebuffer_size)
        * gui_app_wrapper.dpi_scaling
        / mn.Vector2(gui_app_wrapper.window_size)
```

Your code will work correctly only with an integer ratio of the framebuffer and window sizes, casting to a floating-point type makes it work with 1.5x scale and such too (I have a 1.6875x scale on my external display, for example).

rpartsey: Thank you @mosra!

mosra:
Oh, sorry, one more thing I forgot -- for crisp rendering you'll probably want to keep the display size and the size at the font is opened & rasterized in sync. I.e., this line should take `display_font_size` too, instead of having 13 hardcoded: https://github.com/facebookresearch/habitat-lab/blob/7abbc2a1ee4ed01666ffb75760610baaddd79280/examples/siro_sandbox/utils/gui/text_drawer.py#L64

An even crisper rendering can be achieved if the loaded font size is twice as much as the display size, so `2*display_font_size`. 

You might (or might not) run into an issue where the larger font doesn't fit into the GlyphCache after that anymore, to fix that bump its size from 256 to 512, for example.

rpartsey:
> bump its size from 256 to 512, for example.

Thank you for the suggestions. Looks really nicer. In my case I had to increase GlyphCache size to 6 * 256 after setting load_font_size = 2 * display_font_size to make text appear on the screen (using our default font size 40).

mosra:
Yeah the efficiency of the glyph cache packing that's there isn't great, sorry about that :) 

Coincidentally I'm currently working on the text renderer, and there's [a new atlas packer with a much better packing efficiency](https://doc.magnum.graphics/magnum/classMagnum_1_1TextureTools_1_1AtlasLandfill.html). I'll be opening a PR updating Magnum hopefully sometime this week, then you'll be able to reduce that size back to something reasonable again.

habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -62,7 +63,7 @@ def __init__(
         self._max_display_text_chars = max_display_text_chars
 
         # Glyphs we need to render everything
-        self._glyph_cache = text.GlyphCache(mn.Vector2i(256))
+        self._glyph_cache = text.GlyphCache(mn.Vector2i(6 * 256))
```
eundersander, line 66: How about a comment on how we're choosing this number?

mosra: Hmm, and with https://github.com/facebookresearch/habitat-sim/pull/2243 now in, which includes the above-mentioned better atlas packer, it probably doesn't have to be that large anymore.

rpartsey: Just tested, still has to be at least 6 * 256, otherwise text don't appear ob the screen.

PR #1631, opened 2023-10-18
Better init location

jimmytyyang:


examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -227,8 +232,13 @@ def on_environment_reset(self, episode_recorder_dict):
             self.get_sim().get_agent_data(
                 self.get_gui_controlled_agent_index()
             ).articulated_agent.base_pos = human_pos
-            # Make robot look at human
-            self._make_robot_face_human(human_pos, robot_pos)
+        # Make agents look at each other
+        self._make_agent_face_target(
+            human_pos, 1 - self.get_gui_controlled_agent_index()
```
eundersander, line 237: self.state_machine_agent_ctrl._agent_idx

examples/siro_sandbox/sandbox_app.py
```diff
@@ -367,6 +365,9 @@ def _reset_environment(self):
                 self._episode_recorder_dict,
             )
 
+        # Call resetting the controller after the app reset since the human action requires to get a human base pos
+        self.ctrl_helper.on_environment_reset()
```
eundersander, line 369: Can you update this comment: "hack: we have to reset controllers after AppState reset in case AppState reset overrides the start pose of agents"

In general, this is a bad hack because it's not obvious why this should need to run after AppState reset. We can aim to fix this hack in the future.

eundersander (approved):
Please address my comments, otherwise LGTM!

PR #1654, opened 2023-10-27
Unit test for HITL tool and/or SandboxDriver

rpartsey:
## Motivation and Context

Initial version of the SandboxDriver unit test.

Known issues:
- sometimes sim generated keyframe has slightly different agents positions even though we force them to be fixed by adding "overfit" config argument (for the context if "overfit" is in config we fix numpy and sim.pathfinder seeds). probably because float numbers representation? This causes test to fail sometimes.

## How Has This Been Tested
Running `pytest examples/siro_sandbox/test_sandbox_app.py ` locally (requires hab3_bench_assets assets).

## Types of changes

- **Development**  Adding test.

## Checklist


examples/siro_sandbox/test_sandbox_app.py
```diff
@@ -0,0 +1,79 @@
...
+from habitat_baselines.config.default import get_config as get_baselines_config
+
+
+@pytest.mark.skipif(
+    not os.path.exists("data/hab3_bench_assets"),
+    reason="This test requires hab3_bench_assets assets.",
+)
+@patch("utils.gui.replay_gui_app_renderer.ReplayRenderer", spec=True)
+@patch("utils.gui.replay_gui_app_renderer.ImageFramebufferDrawer", spec=True)
+@patch("utils.gui.replay_gui_app_renderer.TextDrawer", spec=True)
```
rpartsey, line 24: Patching these thing until https://app.asana.com/0/home/1205684818278375/1205279290945037 is done.

examples/siro_sandbox/test_sandbox_app.py
```diff
@@ -0,0 +1,79 @@
...
+    get_sandbox_app_args_parser,
+    override_config,
+)
+from utils.gui.gui_input import GuiInput
+from utils.gui.replay_gui_app_renderer import ReplayGuiAppRenderer
+
+from habitat_baselines.config.default import get_config as get_baselines_config
+
+
+@pytest.mark.skipif(
```
eundersander, line 18: I'm curious if we can run this as part of CI. I started a slack thread [here](https://cvmlp.slack.com/archives/C0460NTKM4G/p1698436827946659).

examples/siro_sandbox/test_sandbox_app.py
```diff
@@ -0,0 +1,79 @@
...
+        text_drawer=app_renderer._text_drawer,
+    )
+
+    sim_dt = 1 / default_args.target_sps
+    post_sim_update_dict = sandbox_driver.sim_update(dt=sim_dt)
+
+    with open("examples/siro_sandbox/test_sandbox_app_keyframe.txt", "r") as f:
+        expected_keyframe = f.read()
+
+    assert post_sim_update_dict["keyframes"][0] == expected_keyframe
```
eundersander, line 79: Can you provide instructions here (as a code comment) for how to update `examples/siro_sandbox/test_sandbox_app_keyframe.txt` if/when this test breaks due to desired changes to the app behavior?

eundersander, line 79: In addition to this equality check on the entire keyframe contents (which I like), can you also sanity-check the contents of the keyframe? There should be at least one keyframe, one load, one creation, and one stateUpdate, and the comment for that set of checks should be "the first keyframe should create and pose at least one render asset instance".

eundersander (commented):
Good progress! See comments below.

examples/siro_sandbox/test_sandbox_app.py
```diff
@@ -0,0 +1,79 @@
...
+        window_size=mn.Vector2i(default_args.width, default_args.height)
+    )
+
+    gui_input = GuiInput()
+
+    sandbox_driver = SandboxDriver(
+        args=default_args,
+        config=default_config,
+        gui_input=gui_input,
+        line_render=app_renderer._replay_renderer.debug_line_render(0),
```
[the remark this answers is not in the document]
rpartsey: Agree, thanks for the suggestion. Advantage of `DummyClass` over the patching is that this way testing SandboxDriver doesn't require importing and instantiating `ReplayRenderer`.

examples/siro_sandbox/test_sandbox_app.py
rpartsey: As Andrew replied the Slack tread `this assets are actually already being downloaded in the multi-agent trainer CI`.

examples/siro_sandbox/test_sandbox_app.py
rpartsey: Done

examples/siro_sandbox/test_sandbox_app.py
eundersander: Cool. Can you link to a successful CI run of this file here as a reply? It's not clear to me at a glance if this file is getting included in the tests as part of CI `install_and_test_ubuntu`.

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
```diff
@@ -243,6 +248,10 @@ def reset(self, episode: Episode, fetch_observations: bool = True):
             self._is_episode_active = True
 
             if self._should_place_articulated_agent:
+                if self._fixed_agent_starting_pose:
```
eundersander, line 251: Are you happy with this implementation? I see two old comments that suggest this wasn't quite working correctly yet:

From your PR description:
>slightly different agents positions even though we force them to be fixed by adding "fixed_agent_starting_pose" config argument (for the context if "fixed_agent_starting_pose" is in config we fix numpy and sim.pathfinder seeds). probably because float numbers representation? This causes test to fail sometimes.

From [this Slack thread](https://cvmlp.slack.com/archives/C0460NTKM4G/p1698675347202439?thread_ts=1698438249.886569&cid=C0460NTKM4G):
>So, I would expect that setting env.seed(my_seed) would be enough to have deterministic behaviour, but it is not, agent positions are set to different positions every time. Is it working as expected [@Andrew Szot](https://cvmlp.slack.com/team/U011205HNNT) [@Alex Clegg](https://cvmlp.slack.com/team/UKZ187YA3)?

eundersander (commented):
Added some more comments.

PR #1659, opened 2023-10-31
Decrease text.GlyphCache size

rpartsey:
## Motivation and Context

Decrease text.GlyphCache size as suggested by mosra [here](https://github.com/facebookresearch/habitat-lab/pull/1626#discussion_r1373379384).

## Types of changes

- **Refactoring**

## Checklist

### DIFF
--- a/habitat-lab/habitat/gui/text_drawer.py
+++ b/habitat-lab/habitat/gui/text_drawer.py
@@ -63,8 +63,9 @@ class TextDrawer:
         self._max_display_text_chars = max_display_text_chars
 
         # Glyphs we need to render everything
-        # Using 3 * 256 as a size of the GlyphCache, to fit larger font size
-        self._glyph_cache = text.GlyphCache(mn.Vector2i(3 * 256))
+        # Using(1024, 768) as a size of the GlyphCache, to fit larger font size
+        # Ideal size for the GPU is of power-of-two in at least one dimension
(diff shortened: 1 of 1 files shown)


habitat-lab/habitat/gui/text_drawer.py
```diff
@@ -63,8 +63,8 @@ def __init__(
         self._max_display_text_chars = max_display_text_chars
 
         # Glyphs we need to render everything
-        # Using 6 * 256 as a size of the GlyphCache, to fit larger font size
-        self._glyph_cache = text.GlyphCache(mn.Vector2i(6 * 256))
+        # Using 3 * 256 as a size of the GlyphCache, to fit larger font size
+        self._glyph_cache = text.GlyphCache(mn.Vector2i(3 * 256))
```
mosra, line 67: Thanks!

Ideal for the GPU would be to stay with power-of-two sizes in at least one dimension. I think this could still fit as well, it has roughly the same area:

```suggestion
        self._glyph_cache = text.GlyphCache((1024, 512))
```

mosra (approved):

habitat-lab/habitat/gui/text_drawer.py
rpartsey: Unfortunately, that isn't enough, using  (1024, 768)  instead. Still better that 6 * 256.

Thank you!

PR #1665, opened 2023-11-01
Oracle social nav for HITL study

jimmytyyang:
- Introduce a new state called `FOLLOW_ORACLE` that calls oracle nav to follow human
- For consistency, we use the same key "O" to activate the oracle social nav as the same hot key we used for search_oracle and bring_oracle
- Same CLI, nothing changed
- See the video (holding the key "O"): https://drive.google.com/file/d/1kBktMm6c2X8qzCsX9dbRjCdQ54dXAs4B/view?usp=sharing


eundersander (approved):
LGTM! FYI, related to social nav, we might want some path visualization for this, including differentiating learned social-nav from oracle point-nav. But that doesn't need to be in this PR.

jimmytyyang:
Thanks Eric! will have one more PR for adding visual :)

PR #1668, opened 2023-11-03
SocialNavStudy App State

rpartsey:
## Motivation and Context

Adds AppStateSocialNavStudy app state class that is copy of AppStateRearrange but uses FetchBaselinesController to control robot policy. We use only  FOLLOW and FOLLOW_ORACLE FetchStates:
- FOLLOW (with --enable-hybrid-social-nav flag disabled) for SocialNav policy
- FOLLOW_ORACLE for oracle policy

## How Has This Been Tested

Running sandbox app locally on MacBook Pro 2021.

Tested with:
- habitat-lab (this branch; 2493925)
- habitat-sim (5808983)
- data/fpss (01830358a)
- data/datasets/floorplanner (00c8e60)

**Launch command**
Terminal
```bash
HABITAT_SIM_LOG=warning:physics,metadata=quiet MAGNUM_LOG=warning \
python examples/siro_sandbox/sandbox_app.py \
--app-state socialnav_study \
--display-font-size 14 \
--never-end \
--gui-controlled-agent-index 1 \
--ang-speed 15 \
--first-person-mode \
--hide-humanoid-in-gui \
--save-episode-record \
--save-filepath-base my_session \
--cfg examples/siro_sandbox/configs/socialnav_study.yaml \
--cfg-opts \
habitat.dataset.data_path=data/datasets/floorplanner/rearrange/scratch/test/largetest_10s_500epi_2obj.json.gz \
~habitat.task.measurements.agent_blame_measure
```

VS Code
```
{   
    "type": "python",
    "request": "launch",
    "program": "examples/siro_sandbox/sandbox_app.py",
    "env": {
        "HABITAT_SIM_LOG": "warning",
        "MAGNUM_LOG": "warning"
    },
    "args": [
        "--app-state", "socialnav_study",  
        "--display-font-size", "14",
        "--disable-inverse-kinematics",
        "--never-end",
        "--gui-controlled-agent-index", "1",
        "--ang-speed", "15",
        "--first-person-mode",
        "--hide-humanoid-in-gui",
        // "--oracle-follow-human",
        "--save-episode-record",
        // "--enable-gfx-replay-save",
        "--save-filepath-base", "my_session",
        "--cfg", "examples/siro_sandbox/configs/socialnav_study.yaml",
        "--cfg-opts",
"habitat.dataset.data_path=data/datasets/floorplanner/rearrange/scratch/test/largetest_10s_500epi_2obj.json.gz",
        "~habitat.task.measurements.agent_blame_measure",
    ],
}
```

By default, on app startup Spot is controlled by learned SocialNav policy.   If you want Spot to be controlled by oracle from app start use "--oracle-follow-human" flag. Also, you may press "O" key to switch between learned SocialNav policy and oracle nav.

Add "--enable-gfx-replay-save" if you want to save gfx replay keyframes (we discussed we don't need this, so I didn't include in the launch command).

## Types of changes

- **Development:** PD into `eundersander/SIRo_hitl_Sep5_demo_refactored` (demo branch)

## Checklist


examples/siro_sandbox/configs/social_nav_study.yaml
```diff
@@ -0,0 +1,4 @@
+# @package _global_
+
+defaults:
+  - fetch@_here_
```
eundersander, line 4: I don't think we should inherit from fetch config yaml. The underlying principle here is that someone should be free to tweak the fetch app state without breaking/affecting other app states. How about some kind of base sandbox config yaml that both fetch and social_nav_study inherit from?

examples/siro_sandbox/controllers/controller_helper.py
```diff
@@ -101,24 +101,36 @@ def __init__(
                 )
             self.controllers.append(gui_agent_controller)
             if is_multi_agent:
-                if args.app_state == "fetch":
+                if (
```
eundersander, line 104: Instead of checking args.app_state, let's have the calling code do this logic. And then let's add some parameters to `ControllerHelper.__init__`: 1. an integer for policy_controlled_agent_index (this lets us get rid of the hacky code below for computing this index), and 2. a bool use_fetch_baselines_controller.

Background: my philosophy here is that helper classes should know as little as possible about the calling code (this makes them more reusable). Here, a tension is that it's so convenient to just pass args to this helper class rather than forward all the individual items from args that this class needs (like args.lin_speed), with the downside that access to args lets this class know a lot (basically too much) about the calling context (like the existence of app states and which app state it is being called from).

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -81,6 +81,7 @@ def __init__(
         gym_env,
         habitat_env,
         hybrid_social_nav,
+        oracle_follow,
```
eundersander, line 84: I always like boolean parameter names to sound boolean. How about `force_oracle_follow` or `use_oracle_follow`.

examples/siro_sandbox/episode_helper.py
```diff
@@ -2,9 +2,18 @@ class EpisodeHelper:
     def __init__(self, habitat_env):
         self._habitat_env = habitat_env
 
-        self._num_iter_episodes = float("inf") if habitat_env.episode_iterator.cycle else len(self._habitat_env.episode_iterator.episodes)  # type: ignore
+        self._episode_iterator_cycle = habitat_env.episode_iterator.cycle
+        self._num_iter_episodes: int = len(habitat_env.episode_iterator.episodes)  # type: ignore
         self._num_episodes_done: int = 0
 
+    @property
```
eundersander, line 9: I prefer the old code. The new code looks like needless complexity. The old code has the same functionality. The new code has the confusing disparity `self._num_iter_episodes != self.num_iter_episodes` in the cycle case, which may eventually cause a bug when someone misunderstands.

examples/siro_sandbox/sandbox_app.py
```diff
@@ -604,6 +605,12 @@ def _parse_debug_third_person(args, framebuffer_size):
         type=int,
         help="Vertical resolution of the window.",
     )
+    parser.add_argument(
```
eundersander, line 608: Nice!

examples/siro_sandbox/sandbox_app.py
```diff
@@ -625,6 +632,11 @@ def _parse_debug_third_person(args, framebuffer_size):
         action="store_true",
         help="If specified, we use a hybrid social nav design that consists of point nav and social nav. Otherwise, we only use social nav policy purely.",
     )
+    parser.add_argument(
+        "--oracle-follow-human",
+        action="store_true",
+        help="If specified, we use the oracle to follow human.",
```
eundersander, line 638: Maybe `If true, use an oracle pointnav policy to follow the human. If false, use a learned social-nav policy.`?

examples/siro_sandbox/app_states/app_state_socialnav_study.py
```diff
@@ -0,0 +1,459 @@
...
+        self._camera_helper.update(self._get_camera_lookat_pos(), dt=0)
+
+        if episode_recorder_dict:
+            episode_recorder_dict["target_obj_ids"] = self._target_obj_ids
+            episode_recorder_dict["goal_positions"] = self._goal_positions
+
+    def get_sim(self):
+        return self._sandbox_service.sim
+
+    def _update_grasping_and_set_act_hints(self):
```
eundersander, line 84: If you're feeling motivated, I can imagine a `GuiRearrangeHelper` class that encapsulates this grasping logic and visualization and gets shared between AppStateRearrange and AppStateSocialNavStudy.

However I'm also happy to defer this to a future PR. I'm curious to see if the specifics here end up diverging a bit between the two app states over time.

eundersander (commented):
Good progress! Let me know when I should take another look.

examples/siro_sandbox/sandbox_app.py
rpartsey: Done

examples/siro_sandbox/episode_helper.py
rpartsey: Agree, reverted

examples/siro_sandbox/controllers/fetch_baselines_controller.py
rpartsey: Done

examples/siro_sandbox/controllers/controller_helper.py
rpartsey: Added use_fetch_baselines_controller argument.

Regarding policy_controlled_agent_index, agents and agent controllers are not the same things. We don't know agent controller index in advance in the calling code since agent controllers are instantiated by ControllerHelper it its __init__. 

I've added comments with explanations. Let me know if it looks good to you.

examples/siro_sandbox/configs/social_nav_study.yaml
rpartsey: Yes, good catch!  Well, in this case the most secure decision is to have two independent config copies. 

In fact, both fetch and social_nav_study configs are not app state config but rather baselines configs overrides (I'd say finetuned versions of rearrange baselines config). So, how about having two independent copies? This is similar to what we do with app sates classes: copy instead of overrides.

examples/siro_sandbox/app_states/app_state_socialnav_study.py
rpartsey: Will work on this in separate PR.

examples/siro_sandbox/configs/social_nav_study.yaml
eundersander: >are not app state config but rather baselines configs overrides
Agree! And we might say each is a baseline config override intended to be used with a particular app state (hence their names).

eundersander: I do still think a base sandbox config would make sense. I.e. a baseline config override intended to be used by all sandbox app use cases. And then fetch.yaml and social_nav_study.yaml can do a few additional overrides of that.

examples/siro_sandbox/configs/social_nav_study.yaml
```diff
@@ -0,0 +1,291 @@
+# @package _global_
+
+defaults:
+  - /benchmark/rearrange: multi_agent_tidy_house_fp_spot_humanoid
```
eundersander, line 4: I suspect you are copy-pasting a ton of configs between this and fetch.yaml. I still suggest a base sandbox config yaml for stuff that is fundamentally required for the sandbox app (and not specific to an app state). This is conceptually the same as how we have lots of shared code in sandbox_driver.py that is shared between app states (because that code is fundamental to the sandbox app, regardless of app state).

examples/siro_sandbox/controllers/controller_helper.py
```diff
@@ -101,39 +102,59 @@ def __init__(
...
+        # if there is no gui controlled agent the policy driven agent controller
+        # will always be one and only one in the controllers list
+        # either SingleAgentBaselinesController or MultiAgentBaselinesController
+        # depending on the number of agents
         if self._gui_controlled_agent_index is None:
             return self.controllers[0]
-
-        # This is pretty hacky
-        return self.controllers[1]
+        # if there is gui controlled agent (its controller is always the first one in the list)
```
eundersander, line 150: These comments are helpful, thanks!

eundersander (commented):
I'd still like to see a base sandbox config yaml as described below. Feel free to reach out on Slack or reply here if you disagree!

eundersander (approved):
I spotted a Hydra warning when testing this (below), otherwise LGTM!

PR #1674, opened 2023-11-07
Send signal when episode changes so that the client can react.

0mdc:
## Motivation and Context

This sends a signal to the client when the episode has changed so that it can react accordingly.
Used to implement an effect in the Unity client here: https://github.com/eundersander/siro_hitl_unity_client/pull/16

### DIFF
--- a/examples/siro_sandbox/server/client_message_manager.py
+++ b/examples/siro_sandbox/server/client_message_manager.py
@@ -46,8 +46,8 @@ class ClientMessageManager:
         """
         self._message["teleportAvatarBasePosition"] = [pos[0], pos[1], pos[2]]
 
-    def signal_episode_change(self) -> None:
+    def signal_scene_change(self) -> None:
         r"""
-        Signals the client that the episode is being changed during this frame.
+        Signals the client that the scene is being changed during this frame.
         """
-        self._message["episodeChanged"] = True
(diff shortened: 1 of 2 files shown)


examples/siro_sandbox/server/client_message_manager.py
```diff
@@ -45,3 +45,9 @@ def change_humanoid_position(self, pos: List[float]) -> None:
         Used to synchronize the humanoid position in the client when changing scene.
         """
         self._message["teleportAvatarBasePosition"] = [pos[0], pos[1], pos[2]]
+
+    def signal_episode_change(self) -> None:
+        r"""
+        Signals the client that the episode is being changed during this frame.
+        """
+        self._message["episodeChanged"] = True
```
eundersander, line 53: Same naming comment as in the other PR. I feel like the Unity codebase shouldn't need to have RL or ML-specific concepts like "episode".

eundersander (approved):
Aside from minor comment, LGTM!

examples/siro_sandbox/server/client_message_manager.py
eundersander: Related: let's document this function to tell folks what we're really telling the client with this message: that we're requesting a transition effect to help convey the scene-change to the user.

PR #1675, opened 2023-11-08
Interesting cases for the demo + Terminate FOLLOW state

jimmytyyang:
Two things in this PR
- Add the termination condition for the FOLLOW state. It terminates if the robot is close to a human and the human does not move. See video: https://drive.google.com/file/d/1R9TBeNQrnyCfp03MQwn_9ULtDLbP1KFU/view?usp=sharing, and link to Asana task: https://app.asana.com/0/0/1205912179057906/f
- Get a good spawning location for both the robot and the human

### DIFF
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -50,26 +50,40 @@ ROBOT_SPAWN_NEAR_HUMAN_DIS = 5.0
 # Randomly reset the location of the robot or human
 RANDOM_AGENT_LOCATION_RESET = False
 # Sloppy: hardcore location of the robot and human
+# Note that you do not have to provide the orientation information
+# since both agents will automatically look at each other
 FIX_AGENT_INIT_POS = {
     "data/fpss/scenes-uncluttered/102344193.scene_instance.json": (
-        [-1.8878, -5.14189],
-        [-5.84479, -4.75223],
+        # Scene 1: human stands near the bed and the robot blocked the room door.
+        # So the human needs to walk out of room and will observe how Spot avoid collision.
+        # This showcases the social nav policy
+        [0.773143, -4.09137],
+        [1.53322, -1.53417],
     ),
     "data/fpss/scenes-uncluttered/102344280.scene_instance.json": (
+        # Scene 2: human stands in the kitchen + living room area. Just showcase the scene
         [-1.4465, 1.36246],
         [-7.46346, 1.5259],
     ),
     "data/fpss/scenes-uncluttered/102817200.scene_instance.json": (
-        [-6.77723, -11.9282],
-        [-6.40974, -7.65545],
+        # Scene 3: human stands in the living room with lots of furniture (sofa, cluter, etc),
+        # and it is easier to trigger Spot begging motion.
+        # This showcases the pick policy/begging motions
+        [-3.63359, -8.33051],
+        [-1.03828, -7.71454],
     ),
     "data/fpss/scenes-uncluttered/103997424_171030444.scene_instance.json": (
-        [-11.8683, -6.15889],
-        [-17.2134, -8.89483],
+        # Scene 4: human stands near the shelf in which he/she can put the object on
+        # multiple surfaces of the shelf, and at the beginning, human is able to
+        # see robot come to find him/herself. This showcases the pick policy
+        [-14.3757, -10.0978],
+        [-18.5925, -12.9898],
     ),
     "data/fpss/scenes-uncluttered/103997541_171030615.scene_instance.json": (
-        [-10.7404, 0.634613],
-        [-11.9156, -4.56228],
+        # Scene 5: human passes a long hallway to reach the target and get
+        # the object. This showcases the social nav policy
+        [-10.84, 5.24],
+        [-10.86, 8.18],
     ),
 }
 
--- a/examples/siro_sandbox/app_states/app_state_fetch.py
+++ b/examples/siro_sandbox/app_states/app_state_fetch.py
@@ -49,7 +49,7 @@ disable_spot = False
 ROBOT_SPAWN_NEAR_HUMAN_DIS = 5.0
 # Randomly reset the location of the robot or human
 RANDOM_AGENT_LOCATION_RESET = False
-# Sloppy: hardcore location of the robot and human
+# Sloppy: hardcode location of the robot and human
 # Note that you do not have to provide the orientation information
 # since both agents will automatically look at each other
 FIX_AGENT_INIT_POS = {
--- a/examples/siro_sandbox/controllers/fetch_baselines_controller.py
+++ b/examples/siro_sandbox/controllers/fetch_baselines_controller.py
@@ -150,6 +150,11 @@ class FetchBaselinesController(SingleAgentBaselinesController):
         }
         return {"ll_policy": policy_info}
 
+    def change_state(self, next_state):
+        """A helpful function for handling state change"""
+        self.current_state = next_state
+        self._policy_info = self._get_policy_info()
+
     def cancel_fetch(self, skip_reset_arm=False):
         if self.grasped_object:
             env = self._habitat_env
@@ -516,8 +521,9 @@ class FetchBaselinesController(SingleAgentBaselinesController):
         # Start the state machine
         if self.current_state == FetchState.WAIT:
             if not self._finished_follow:
-                self.current_state = FetchState.FOLLOW
-            self._policy_info = self._get_policy_info()
+                self.change_state(FetchState.FOLLOW)
+            else:
+                self.change_state(FetchState.WAIT)
 
         elif self.current_state == FetchState.FOLLOW:
             # This is the following state, in which the robot tries to follow the human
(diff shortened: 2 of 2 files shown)


examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -68,8 +68,11 @@ class FollowStatePolicy(Enum):
 START_FETCH_ROBOT_DIS_THRESHOLD = (
     1.8  # if the human is too close to Spot, they block Spot
 )
+# The distance between the robot and the human to switch from point nav to social nav
 FOLLOW_SWITCH_GEO_DIS_FOR_POINT_SOCIAL_NAV = 2.5
 PICK_STEPS = 40
+# The distance between the robot and the human to consider termination of FOLLOW state
+ROBOT_CLOSE_TO_HUMAN_DIS = 2.0
```
eundersander, line 75: Ruta will need a way to disable this for the social-nav study, since we want "pure learned social-nav" with no extra heuristics. Can you add a command-line arg or similar?

examples/siro_sandbox/app_states/app_state_fetch.py
```diff
@@ -45,27 +45,41 @@
...
+        # Scene 4: human stands near the shelf in which he/she can put the object on
+        # multiple surfaces of the shelf, and at the beginning, human is able to
+        # see robot come to find him/herself. This showcases the pick policy
+        [-14.3757, -10.0978],
+        [-18.5925, -12.9898],
     ),
     "data/fpss/scenes-uncluttered/103997541_171030615.scene_instance.json": (
-        [-10.7404, 0.634613],
-        [-11.9156, -4.56228],
+        # Scene 5: human passes a long hallway to reach the target and get
```
eundersander, line 79: These all sound great!

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -411,10 +417,37 @@ def act(self, obs, env):
...
         robot_pos = env._sim.agents_mgr[
             self._agent_idx
         ].articulated_agent.base_transformation.translation
 
+        # Check if the human walks or rotates
+        is_human_walk = (
+            np.linalg.norm(
+                np.array((self._prev_human_pos - human_pos))[[0, 2]]
+            )
+            > 0.01
```
eundersander, line 432: Can you expose this as a global at top of file? I will need to tune it for VR, where the headset is always moving slightly.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -411,10 +417,37 @@ def act(self, obs, env):
...
+            if self._prev_human_pos is not None
+            else False
+        )
+        is_human_rotate = (
+            (
+                (np.abs(float(self._prev_human_rot - human_rot)) + np.pi)
+                % (2.0 * np.pi)
+                - np.pi
+            )
+            > 0.01
```
eundersander, line 442: Same here.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -411,10 +417,37 @@ def act(self, obs, env):
...
+        # Check if the human walks or rotates
+        is_human_walk = (
+            np.linalg.norm(
+                np.array((self._prev_human_pos - human_pos))[[0, 2]]
+            )
+            > 0.01
+            if self._prev_human_pos is not None
+            else False
+        )
+        is_human_rotate = (
```
eundersander, line 436: Do you think it's necessary to resume social nav if the human is rotating without changing translation? E.g. this would be relevant if our social-nav policy tries to face the front of the human (or if it might do this in the future).

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -411,10 +417,37 @@ def act(self, obs, env):
...
             1 - self._agent_idx
         ].articulated_agent.base_transformation.translation
+        human_rot = env._sim.agents_mgr[
+            1 - self._agent_idx
+        ].articulated_agent.base_rot
         robot_pos = env._sim.agents_mgr[
             self._agent_idx
         ].articulated_agent.base_transformation.translation
 
+        # Check if the human walks or rotates
```
eundersander, line 427: You should move all this logic into a helper function for readability.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -477,23 +512,34 @@ def act(self, obs, env):
...
+                        act_space, "agent_0_oracle_nav_action"
+                    )
+                    action_array[
+                        action_ind_nav[0] : action_ind_nav[0] + 3
+                    ] = human_pos
+                else:
+                    raise ValueError(f"Skill {type_of_skill} not recognized.")
             else:
-                raise ValueError(f"Skill {type_of_skill} not recognized.")
+                self.current_state = FetchState.WAIT
```
eundersander, line 541: Given the current complexity of your state machine, can I suggest you get more organized about how to transition states? Just as a tiny example here of the kind of bugs we want to avoid, I wonder if you should have cleared `gt_path` on this transition right here. This kind of stuff is much easier to catch if you have one `change_state` helper function that lets you carefully think about every possible state transition, e.g. always calling `self._policy_info = self._get_policy_info()` and always clearing `gt_path`.

For reference, I had started to do that with my `cancel_fetch` helper function, although that was aimed more narrowly at ending the fetch behavior (not relevant for this state transition).

eundersander: Related: it looks like `act()` is over 500 lines long now... that suggests you want to do some cleanup to make it more readable and maintainable.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
```diff
@@ -411,10 +417,37 @@ def act(self, obs, env):
...
+            1 - self._agent_idx
+        ].articulated_agent.base_rot
         robot_pos = env._sim.agents_mgr[
             self._agent_idx
         ].articulated_agent.base_transformation.translation
 
+        # Check if the human walks or rotates
+        is_human_walk = (
+            np.linalg.norm(
+                np.array((self._prev_human_pos - human_pos))[[0, 2]]
```
eundersander, line 430: It seems possible in some edge case (perhaps some future edge case) that `_prev_human_pos` will be `None` here. You should handle that gracefully; similar for `_prev_human_rot`.

eundersander (commented):
Good progress! I left some comments.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Thanks, Eric! This has been done. Please see the snapshot document, and let me know if you have any questions!

jimmytyyang: you can do  `--enable-hybrid-social-nav` in CLI to enable hybrid social nav

examples/siro_sandbox/app_states/app_state_fetch.py
jimmytyyang: Thanks, Eric! Thanks for the helpful discussion!

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Thank you, Eric! I added `DISTANCE_ROTATION_TRESHOLD_HUMAN_MOVING` for this.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Thank you, Eric! I added `DISTANCE_ROTATION_TRESHOLD_HUMAN_MOVING` for this.

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: For my observation based on few cases, it seems that it does not matter. However, it is safer to let robots move so that it can adjust its orientation. And by the way, good point! Happy to discuss more :)

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Thanks, Eric! Agree with you!

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Hi Eric, I did add `if self._prev_human_pos is not None` to check the condition. Please let me know if this is okay. Thank you!

examples/siro_sandbox/controllers/fetch_baselines_controller.py
jimmytyyang: Thanks, Eric! Working on this

jimmytyyang: Initial implementation is done

Issue #1718, opened 2023-12-17
[Social Nav] Error during evaluation ('NoneType' object is not subscriptable.)


qwerty863:
## Habitat-Lab and Habitat-Sim versions
Habitat-Lab: vx.x.x or master?
master

Habitat-Sim: vx.x.x or master?
master

Habitat is under active development, and we advise users to restrict themselves to stable releases. Are you using the latest release versions of [Habitat-Lab](https://github.com/facebookresearch/habitat-lab/releases) and [Habitat-Sim](https://github.com/facebookresearch/habitat-sim/releases)? Your question may already be addressed in the latest versions. We may also not be able to help with problems in earlier versions because they sometimes lack the more verbose logging needed for debugging.

Master branch contains 'bleeding edge' code and should be used at your own risk.

## Docs and Tutorials
Did you read the docs? https://aihabitat.org/docs/habitat-lab/
Yes

Did you check out the tutorials? https://aihabitat.org/tutorial/2020/
Yes

Perhaps your question is answered there. If not, carry on!

## ❓ Questions and Help
Hi, I'm evaluating my reproduced social navigation agent following the command given by @jimmytyyang with a different `habitat_baselines.test_episode_count` value (500->1200) since the validation scenes are 12 and each has 100 episodes.
https://github.com/facebookresearch/habitat-lab/tree/main/habitat-baselines

But I got this error ('NoneType' object is not subscriptable).
Has this error been seen before? Or am I missing something?
```
 78%|██████████████████▉               | 939/1200 [3:21:16<1:04:45, 14.89s/it][ERROR,rearrange_task] 2023-12-17 06:47:18,091 pddl_sim_state.py:217 Failed to place the robot.
[...]

PR #1744, opened 2024-01-10
[BE] Fix the env pause issue

jimmytyyang:
## Motivation and Context

 This is to fix the issue when doing an evaluation on the HRL policy, the size of internal stateful parameters does not match the number of active skills. The issue has been known in multiple places:
(1) our team's slack channel
(2) github issues: https://github.com/facebookresearch/habitat-lab/issues/1718
(3) myself 
issue_be

## How Has This Been Tested

 Loading the social nav checkpoint and run full dataset evaluation

## Types of changes

## Checklist

### DIFF
--- a/habitat-baselines/habitat_baselines/rl/ppo/habitat_evaluator.py
+++ b/habitat-baselines/habitat_baselines/rl/ppo/habitat_evaluator.py
@@ -302,6 +302,10 @@ class HabitatEvaluator(Evaluator):
                 rgb_frames,
             )
 
+            # We pause the statefull parameters in the policy
+            if any(envs_to_pause):
+                agent.actor_critic.on_envs_pause(envs_to_pause)
+
         pbar.close()
         assert (
             len(ep_eval_count) >= number_of_eval_episodes
--- a/habitat-baselines/habitat_baselines/rl/ppo/habitat_evaluator.py
+++ b/habitat-baselines/habitat_baselines/rl/ppo/habitat_evaluator.py
@@ -302,7 +302,10 @@ class HabitatEvaluator(Evaluator):
                 rgb_frames,
             )
 
-            # We pause the statefull parameters in the policy
+            # We pause the statefull parameters in the policy.
+            # We only do this if there are envs to pause to reduce the overhead.
+            # In addition, HRL policy requires the solution_actions to be non-empty, and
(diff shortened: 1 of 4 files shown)


habitat-baselines/habitat_baselines/rl/ppo/habitat_evaluator.py
```diff
@@ -302,6 +302,10 @@ def evaluate_agent(
                 rgb_frames,
             )
 
+            # We pause the statefull parameters in the policy
+            if any(envs_to_pause):
```
ASzot, line 306: Why add this? Should `on_envs_pause` should be able to handle empty lists of ends to pause?

ASzot (approved):

xavierpuigf (approved):
Same comment as Andrew, I think you dont need the if condition and can always call envs_to_pause with empty list. Unless the reason is efficiecny in not doing these checks.
Thanks for the fix!

jimmytyyang:
@xavierpuigf and @ASzot Thanks for the review! I noticed that this code (https://github.com/facebookresearch/habitat-lab/blob/379b1deb2ef177b51362d2d44eb81d2d9d6e5015/habitat-baselines/habitat_baselines/rl/hrl/hl/fixed_policy.py#L38) could prevent us from removing the condition `if any(envs_to_pause):`

Maybe we could add a check inside `def filter_envs(self, curr_envs_to_keep_active)`?

xavierpuigf:
> @xavierpuigf and @ASzot Thanks for the review! I noticed that this code (
> 
> https://github.com/facebookresearch/habitat-lab/blob/379b1deb2ef177b51362d2d44eb81d2d9d6e5015/habitat-baselines/habitat_baselines/rl/hrl/hl/fixed_policy.py#L38
> 
> ) could prevent us from removing the condition `if any(envs_to_pause):`
> Maybe we could add a check inside `def filter_envs(self, curr_envs_to_keep_active)`?

good point! yeah I would add the check in fixed_policy instead, as you suggest

jimmytyyang:
Thanks for the review! Andrew spoke with me saying that he is fine with the current implementation, and it might introduce an overhead if there are no envs to pause. So based on this, I will add a comment there to say why we want to have this condition. Thanks for the awesome review! :)

PR #1772, opened 2024-01-25
Fixed config habitat_hitl.episodes_filter='4' and episodes_filter episode order

henrysamer:
Fixed error when using config habitat_hitl.episodes_filter='4' and Fixed episodes_filter episode order not respected

**Types of changes**
config habitat_hitl.episodes_filter='4' Casting int->str

episodes_filter episode not in order To keep the sequence, the update uses a list instead of a set to preserve order, appends/extends to list instead of adding to set, and returns the final list

**Local Testing**
To test this change:

Checked out branch fix/episodes_filter
Modified config.yaml to set habitat_hitl.episodes_filter='4'
Ran training job and confirmed no errors when parsing filter
Modified config.yaml to complex filter string: "1000:2000:100 3000 4000 0:10:3"
Inspected order of episode ids before and after change and confirmed order matches filter string
Ran full pipeline with updated filter logic and confirmed no issues

### DIFF
--- a/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
+++ b/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
@@ -12,7 +12,7 @@ import abc
 import json
 from datetime import datetime
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Dict, List, Set
+from typing import TYPE_CHECKING, Any, Dict, List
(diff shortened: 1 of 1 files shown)


eundersander (approved):
Can you test this locally and describe how you tested this in the PR description? Otherwise LGTM!

PR #1778, opened 2024-01-29
[WIP] - Embodied Unoccluded Navmesh Snap 

aclegg3:
## Motivation and Context

TL;DR: replace varied articulated_agent state sampling functionality with one unified interface.

Uses the unoccluded snapping function as a base and adds embodiment collision checking options.

## How Has This Been Tested

Locally with a viewer application.

TODO: should test the downstream users to verify that new configuration and use-cases are fully covered. This is not done for the current state, so I am fairly confident the current useage is faulty already.

Example of heuristic embodiment snapping with examples/rearrange_ep_viewer.py demo app:

https://github.com/facebookresearch/habitat-lab/assets/1445143/e3259551-0547-4662-87ff-d40f145aef87

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
+++ b/habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
@@ -14,7 +14,7 @@ from habitat.tasks.rearrange.actions.actions import (
     BaseVelAction,
     HumanoidJointAction,
 )
-from habitat.tasks.rearrange.utils import place_agent_at_dist_from_pos
+from habitat.tasks.rearrange.utils import embodied_unoccluded_navmesh_snap
 from habitat.tasks.utils import get_angle
 from habitat_sim.physics import VelocityControl
 
@@ -127,14 +127,15 @@ class OracleNavAction(BaseVelAction, HumanoidJointAction):
             obj_pos = self._task.pddl_problem.sim_info.get_entity_pos(
                 nav_to_obj
             )
-            start_pos, _, _ = place_agent_at_dist_from_pos(
-                np.array(obj_pos),
-                0.0,
-                self._config.spawn_max_dist_to_obj,
-                self._sim,
-                self._config.num_spawn_attempts,
-                True,
-                self.cur_articulated_agent,
+            start_pos, _, _ = embodied_unoccluded_navmesh_snap(
+                target_position=np.array(obj_pos),
+                height=1.5,  # NOTE: this is default agent max height. This parameter is used to determine whether or not a point is occluded.
+                sim=self._sim,
+                island_id=self._sim._largest_indoor_island_idx,
+                search_offset=self._config.spawn_max_dist_to_obj,
+                max_samples=self._config.num_spawn_attempts,
+                agent_embodiement=self.cur_articulated_agent,
+                target_object_id=None,  # TODO: this must be the integer id of the target object or no unoccluded state will be found because this object will be considered occluding
             )
             if self.motion_type == "human_joints":
                 self.humanoid_controller.reset(
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
@@ -419,6 +419,7 @@ class RearrangeSim(HabitatSim):
                 island_index=self._largest_indoor_island_idx
             )
 
+            # TODO: is this correct behavior?
             start_pos = self.safe_snap_point(start_pos)
             start_rot = np.random.uniform(0, 2 * np.pi)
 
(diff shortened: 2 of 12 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -548,10 +549,11 @@ def is_point_within_bounds(self, pos):
     def safe_snap_point(self, pos: np.ndarray) -> np.ndarray:
         """
         Returns the 3D coordinates corresponding to a point belonging
-        to the biggest navmesh island in the scenee and closest to pos.
+        to the biggest navmesh island in the scene and closest to pos.
         When that point returns NaN, computes a navigable point at increasing
         distances to it.
         """
+        # TODO: is this correct behavior?
```
ASzot, line 556: What is your concern about the logic here?

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -785,3 +606,197 @@ def get_camera_lookat_relative_to_vertial_line(
     # Get angle between location and the vector
     angle = get_camera_object_angle(cam_T, vertical_dir, local_vertical_dir)
     return angle
+
+
+def embodied_unoccluded_navmesh_snap(
```
ASzot, line 611: My primary concern about this change is the behavior is fundamentally different from the previous `_get_robot_spawns`. If robot spawns failed, it would still return _some_ starting configuration for the robot. However, if this fails, it will return None, and presumably the robot will not be moved. I think it's dangerous not to have this fall back of placing the robot at least somewhere reasonable, even if it's in collision. Not moving the robot at all can cause big problems in reward calculation that are hard to catch.

ASzot (changes requested):

habitat-lab/habitat/tasks/rearrange/actions/oracle_nav_action.py
```diff
@@ -127,14 +127,15 @@ def _get_target_for_idx(self, nav_to_target_idx: int):
...
-                self.cur_articulated_agent,
+            start_pos, _, _ = embodied_unoccluded_navmesh_snap(
+                target_position=np.array(obj_pos),
+                height=1.5,  # NOTE: this is default agent max height. This parameter is used to determine whether or not a point is occluded.
+                sim=self._sim,
+                island_id=self._sim._largest_indoor_island_idx,
+                search_offset=self._config.spawn_max_dist_to_obj,
+                max_samples=self._config.num_spawn_attempts,
+                agent_embodiement=self.cur_articulated_agent,
+                target_object_id=None,  # TODO: this must be the integer id of the target object or no unoccluded state will be found because this object will be considered occluding
```
jimmytyyang, line 138: could this be a receptacle ID? A task where the robot wants to place the object on the receptacle

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -785,3 +606,197 @@ def get_camera_lookat_relative_to_vertial_line(
...
+    # distance to closest snap point is the absolute minimum
+    min_radius = (snap_point - target_position).length()
+    # expand the search radius
+    search_radius = min_radius + search_offset
+
+    # gather a test batch
+    test_batch: List[Tuple[mn.Vector3, float]] = []
+    sample_count = 0
+    while len(test_batch) < test_batch_size and sample_count < max_samples:
+        sample = pathfinder.get_random_navigable_point_near(
```
jimmytyyang, line 675: Do we know that this sample always returns a valid point? In other words, we do not want to have `NaN` cases

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -785,3 +606,197 @@ def get_camera_lookat_relative_to_vertial_line(
...
+                    # if this sample is invalid, try the next
+                    if is_collision:
+                        continue
+
+                if agent_embodiement is not None:
+                    # contact testing with collision shapes
+                    start_position = agent_embodiement.base_pos
+                    start_rotation = agent_embodiement.base_rot
+
+                    agent_embodiement.base_pos = batch_sample[0]
```
jimmytyyang, line 773: There is a case that once ` agent_embodiement.base_pos` is assigned to `nan` point, we cannot reset the agent pose by giving a valid point. do we want to add a check for not assigning nan point?

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -785,3 +606,197 @@ def get_camera_lookat_relative_to_vertial_line(
...
+                    # Only care about collisions between the robot and scene.
+                    is_feasible_state = details.robot_scene_colls == 0
+                    if not is_feasible_state:
+                        continue
+
+                # if we made it here, all tests passed and we found a valid placement state
+                return batch_sample[0], desired_angle, True
+
+    # unable to find a valid navmesh point within constraints
+    return None, None, False
```
jimmytyyang, line 802: at least we can return something to set the agent pose after best effect of searching?

PR #1791, opened 2024-02-05
HITL: Make skinning support optional.

0mdc:
## Motivation and Context

This changeset makes skinning transmission optional.
* If `self._hitl_config.networking.active_features.skinning` is disabled, rig keyframe data is removed before transmission.
* Otherwise, rig keyframe data is consolidated along with other data.

See: https://github.com/0mdc/siro_hitl_unity_client/pull/4

## How Has This Been Tested

Tested locally along with the Unity app.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/pick_throw_vr/config/pick_throw_vr.yaml
+++ b/examples/hitl/pick_throw_vr/config/pick_throw_vr.yaml
@@ -28,5 +28,7 @@ habitat_hitl:
     client_sync:
       # The client controls its own camera.
       camera_transform: False
+      # This is a first-person application. We don't need to transmit skinned mesh poses.
+      skinning: False
 
 pick_throw_vr:
--- a/habitat-hitl/habitat_hitl/_internal/networking/keyframe_utils.py
+++ b/habitat-hitl/habitat_hitl/_internal/networking/keyframe_utils.py
@@ -73,7 +73,9 @@ def update_consolidated_keyframe(
     if "rigCreations" in inc_keyframe:
         if networking_config.active_features.skinning:
             ensure_list(consolidated_keyframe, "rigCreations")
-            consolidated_keyframe["rigCreations"] += inc_keyframe["rigCreations"]
+            consolidated_keyframe["rigCreations"] += inc_keyframe[
+                "rigCreations"
+            ]
         else:
             # Remove rigCreations if skinned mesh pose transmission is disabled.
             del inc_keyframe["rigCreations"]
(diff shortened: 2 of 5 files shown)


henrysamer (approved):
Tested locally, looks good! 
Note: Might need to perform another pre-commit before before adding/committing to ensure ci test passes.

habitat-hitl/habitat_hitl/config/hitl_defaults.yaml
```diff
@@ -34,6 +34,10 @@ habitat_hitl:
       code_available: 200  # "Ok"
       code_unavailable: 503  # "Service Unavailable"
 
+    active_features:
```
eundersander, line 37: I left a similar comment in an earlier PR. "Active" means nothing and "features" is too broad. Maybe your group here is getting at what kinds of data are sent and received; maybe a group name along those lines.

examples/hitl/pick_throw_vr/config/pick_throw_vr.yaml
```diff
@@ -24,5 +24,9 @@ habitat_hitl:
   gui_controlled_agent:
     agent_index: 1
     ang_speed: 15
+  networking:
+    active_features:
+      # This is a first-person application. We don't need to transmit skinned mesh poses.
```
eundersander, line 29: Please add comments pointing people to `hide_humanoid_in_gui` and `camera.first_person_mode`.

I'm seeing that we're creating combinations of Hydra config fields that must be set jointly, otherwise they are somewhat invalid. I guess that is fine but we should at least add comments helping people connect the related config fields.

habitat-hitl/habitat_hitl/_internal/networking/keyframe_utils.py
```diff
@@ -45,25 +47,38 @@ def ensure_dict(keyframe, key):
...
-    #         found = False
-    #         for con_rig_update in consolidated_keyframe["rigUpdates"]:
-    #             if con_rig_update["id"] == id:
-    #                 con_rig_update["pose"] = pose
-    #                 found = True
-    #         if not found:
-    #             consolidated_keyframe["rigUpdates"].append(rig_update)
+    # add or update rigUpdates
+    if "rigUpdates" in inc_keyframe:
+        if networking_config.active_features.skinning:
```
eundersander, line 52: This filtering logic shouldn't be needed here. The `skinning` flag isn't going to change at runtime. Why does this code need to do the filtering if you're already going it in HitlDriver?

eundersander (approved):
I left some comments, otherwise LGTM

eundersander (commented):
For the case that skinning is disabled, are you doing anything to prevent this code from running on the client? https://github.com/0mdc/siro_hitl_unity_client/pull/4/files#diff-f071a636013728b2ff66c2068fe1e9cee3afc61b5af1534a2b6278794e8db51cR281

PR #1800, opened 2024-02-09
HITL: Limit amount of float decimals in JSON transmissions.

0mdc:
## Motivation and Context

This uses the `gfx_replay_manager` API introduced [here](https://github.com/facebookresearch/habitat-sim/pull/2319) to limit the amount of floating point decimals used in HITL JSON serialization.

## How Has This Been Tested

Tested locally. Verified that this doesn't introduce z-fighting.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
+++ b/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
@@ -205,6 +205,12 @@ class HitlDriver(AppDriver):
         assert create_app_state_lambda is not None
         self._app_state = create_app_state_lambda(self._app_service)
 
+        # Limit the number of float decimals in JSON transmissions
+        if hasattr(
+            self.get_sim().gfx_replay_manager, "set_max_decimal_places"
+        ):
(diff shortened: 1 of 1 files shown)


eundersander (approved):
Are you sure about 3 places? GPT says a quaternion error could be as high as 3.6 degrees. I have done this JSON trick previously but I used 5 digits at that time.

Aside from that question, LGTM!

0mdc:
>Are you sure about 3 places? GPT says a quaternion error could be as high as 3.6 degrees.

Great question. I've been running exclusively from this code for a while now. Axis-aligned quaternions don't show precision issues, and animations are smooth (testing with skinned mesh animations).

That aside, I believe that we don't have calculations that strictly depend on having normalized quaternions. As a rule of thumb, we should never trust serialized quaternions to be normalized.

Initially, I was mostly concern with Z-fighting as HSSD has lots of coinciding geometry (carpets, wall art, ...). This also works fine after decreasing prevision.

0mdc:
Changed to 4 decimal places.

PR #1770, opened 2024-01-19
[WIP] - Prepositional logic query utilities

aclegg3:
## Motivation and Context
This PR adds a set of simulator queries for geometric relationships between objects which may be leveraged by a high-level planner or to validate the results of environmental interactions. Also adds some debug line render utils to visualize the query results.

**Current implementation is a work in progress prototype. Expect changing API surface, edge cases, and sub-optimal performance.**

Currently supported queries:
- ~"above"/"below" - given an object, find all other objects which are below/above it via raycasting.~ merged
- ~"within" - given an object, find all other objects which contain it via raycasting.~ merged
- "ontop" - given a supporting object, find all other objects which are supported by it (i.e., they have active contact points with vertically oriented normals).
- "nearby" - in L2 and geodesic - get nearby objects
  - Application: L2 distance between mixed large and small objects is not very accurate. For example, you can touch a sofa and have distance > 1 meter to it's "position". This function attempts to heuristically estimate distance using bounding boxes to trim the L2 distance for each object.
 - "size_regularized_distance" - use object bounding boxes and relative displacement vector to get a heuristically regularized distance between objects.
 - "on_floor" - use navmesh to determine in an object is on the floor (in a valid location) or not
 - ~region queries: determine if an object is in a region~ merged

TODO:
- generalize "above" to other directional queries.
  - add relative directional queries (e.g. "right" or "left" of "objectB")

Note: requires `--bullet` install.

## How Has This Been Tested

TODO: add unit tests, documentation, etc...

For now, `examples/viewer.py` is added as an interactive sandbox/testbed and demo of the features. This viewer is a modifed version of the habitat-sim python viewer. Run it from habitat-lab root directory like:
```
python examples/viewer.py --dataset data/fpss/hssd-hab-articulated-uncluttered.scene_dataset_config.json --scene 102817140
```
or replace the dataset and scene with your own.

UI modifications (differences from default viewer):
- in "GRAB" mousemode (press 'm' to toggle):
  - SHIFT + click an object to set it to motion_type.DYNAMIC in order to move it around
  - click any object to select it, displaying the set of objects from the active query mode (text in the top-left corner shows this)
- 'k' to iterate through active query modes

Debug visualization color codes:
- yellow - the selected object
- green - ManagedRigidObjects which satisfy the query
- cyan - ArticulatedLinks which satisfy the query
- blue - ArticulatedObject parent bounding box if any link satisfies the query.

Example of "within" query:
image
Example of "above" query:
image

Example of "ontop" query:
[...]

PR #1817, opened 2024-02-22
"Above" set query util

aclegg3:
## Motivation and Context

Based on #1770.

Adds "above" preposition query based on raycasting from object keypoints.

NOTE: returns a set of object_id rather than previous prototype's ManagedObject tuples.

## How Has This Been Tested

Added pytests for keypoint casting and "above".

## Types of changes

## Checklist

### DIFF
--- a/test/test_sim_utils.py
+++ b/test/test_sim_utils.py
@@ -14,6 +14,8 @@ from habitat.sims.habitat_simulator.sim_utilities import (
     get_all_object_ids,
     get_all_objects,
     get_ao_link_id_map,
+    get_obj_from_handle,
+    get_obj_from_id,
     snap_down,
 )
 from habitat_sim import Simulator, built_with_bullet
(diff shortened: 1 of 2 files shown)


test/test_sim_utils.py
```diff
@@ -222,3 +224,67 @@ def test_object_getters():
...
+    sim_settings[
+        "scene_dataset_config_file"
+    ] = "data/replica_cad/replicaCAD.scene_dataset_config.json"
+    sim_settings["scene"] = "apt_0"
+    hab_cfg = make_cfg(sim_settings)
+    with Simulator(hab_cfg) as sim:
+        all_objects = get_all_object_ids(sim)
+
+        mixer_object = get_obj_from_handle(
+            sim, "frl_apartment_small_appliance_01_:0000"
```
jacobkrantz, line 248: Are these sim entities stable enough to be included in tests, or should we be placing new entities?

jacobkrantz (approved):
Looks good assuming tests pass and my question about test stability is a non-issue.

test/test_sim_utils.py
aclegg3: I'm assuming these are stable because they are ReplicaCAD scenes which we have no intention of modifying. This scene "apt_0" is a reconstruction of the first static "FRL apartment" arrangement from [Replica](https://github.com/facebookresearch/Replica-Dataset/tree/main). 

If anything were to change in the future it would most likely be dynamic settling which should preserve the tested relationships here.

PR #1823, opened 2024-02-28
ArticulatedLink Open/Close API

aclegg3:
## Motivation and Context

**Adds queries and setters for opening and closing ArticulatedLinks**:
- `get(set)_link_normalized_joint_position` - normalizes joint limit range to [min, max] -> [0,1] and gets or sets position within the normalized range
- `is_open/is_closed` - queries normalized joint range against a threshold value
- `open/close` - sets joint position from a normalized value (0 or 1)

**Limitation**: Assumes that joint limit min corresponds to "closed" and max corresponds to "open". Note that this is not necessarily the case for generic URDF. We must either:
1. provide a config option to invert this relationship
2. enforce this assumption in our assets (e.g. invert non-conforming URDF Joint axes) 

Side thesis: refactor `test_sim_utils` to import the full `sim_utilities` module instead of individual functions. This list was getting unnecessarily long.

## How Has This Been Tested

Added a pytest for the included functionality based on ReplicaCAD. Note this test points out the edge case above for one asset.

**TODO**: should test what happens when no joint limits were set for an object. Could be `inf` and should be handled with assertion.

## Types of changes

## Checklist


habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -825,3 +825,128 @@ def get_object_regions(
...
+    n_pos = (joint_pos - limits[0][joint_pos_ix]) / (
+        limits[1][joint_pos_ix] - limits[0][joint_pos_ix]
+    )
+    return n_pos
+
+
+def set_link_normalized_joint_position(
+    objectA: habitat_sim.physics.ManagedArticulatedObject,
+    link_ix: int,
+    normalized_pos: float,
```
ASzot, line 861: What happens if I pass some value out of range here? Will the joint state wrap around?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -825,3 +825,128 @@ def get_object_regions(
...
+    joint_pos_ix = objectA.get_link_joint_pos_offset(link_ix)
+    limits = objectA.joint_position_limits
+    joint_positions = objectA.joint_positions
+    joint_positions[joint_pos_ix] = limits[0][joint_pos_ix] + (
+        normalized_pos * (limits[1][joint_pos_ix] - limits[0][joint_pos_ix])
+    )
+    objectA.joint_positions = joint_positions
+
+
+def link_is_open(
```
ASzot, line 887: I personally don't think providing `link_is_open` and `link_is_closed` is necessary.

aclegg3: @jacobkrantz don't we need these queries for state evaluation? 
Agreed that these are very light wrappers, so they aren't strictly necessary. Really just convenience/clarity to call an obvious API.
I think these will mostly be necessary if we decide to handle axis inversion programmatically rather than enforcing that assets conform to our assumptions.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Great call-out. I should assert the range [0,1] or clamp here. 
It won't wrap around, it will extrapolate past the joint limits until a clamp is performed (e.g. if auto-clamp joint limits is set a sim step will snap it back).

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
rutadesai: This will allow us to check if a particular drawer is open in a chest of drawers for example, correct? We should keep it for now and remove later if we dont end up using it.

test/test_sim_utils.py
```diff
@@ -454,3 +449,75 @@ def test_region_containment_utils():
...
+                if obj.get_link_joint_type(link_id) in [
+                    JointType.Revolute,
+                    JointType.Prismatic,
+                ]:
+                    if (
+                        "cabinet" in obj.handle
+                        and "right_door" in obj.get_link_name(link_id)
+                    ):
+                        print(
+                            "TODO: Skipping 'cabinet's right_door' link because it does not follow conventions. Axis should be inverted."
```
rutadesai, line 493: not sure if I understand this printed warning. Is this specific to this one example?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -825,3 +825,132 @@ def get_object_regions(
...
+    We assume that joint limits define the closed state (min) and open state (max).
+
+    :param objectA: The parent ArticulatedObject of the link to check.
+    :param link_ix: The index of the link within the parent object. Not the link's object_id.
+    :param threshold: The normalized threshold ratio of joint ranges which are considered "closed". E.g. 0.1 = 10%
+
+    :return: Whether or not the link is considered "closed".
+    """
+
+    return get_link_normalized_joint_position(objectA, link_ix) <= threshold
```
rutadesai, line 926: where (which config) do we expect to store/set this threshold?

rutadesai (approved):
Minor comments -- LGTM! Thanks, Alex :)

test/test_sim_utils.py
aclegg3: Yeah, this is related to a specific asset which does not agree with our API. As mentioned in the "limitations" section of the PR description.

PR #1831, opened 2024-03-05
[HITL] Gui-controlled spot and local multi-user support

eundersander:
## Motivation and Context

* Add support for gui-controlled spot, e.g. `python examples/hitl/rearrange_v2/rearrange_v2.py +experiment=gui_controlled_humanoid_and_spot`. Spot's keyboard controls include `F/V` for forward/back instead of `W/S`; similar alternates for grasping (`N`) and opening (`X`). The camera controls are shared across all users (e.g. `A/D` to turn).
* Add support for multiple local users. See references to `habitat_hitl.gui_controlled_agents` and `user_index`.

Rearrange_v2 multi-user application logic is very crude so far:
* A lot of the hacks below stem from the fact that I have two local users but only one local display. This stuff will change when we have two remote users.
* PickHelper doesn't do proper object-highlighting for two users. Later, we'll need a separate DebugLineRenderer per remote user (and probably a corresponding separate PickHelper per remote user).
* Help text isn't customized per user. Later, we'll need separate help text per remote user.
* Beware agent movement is tied to camera yaw. Later, we'll need a separate camera per remote user.
* Spot forward/back movement is hard-coded to `F` and `V` keys in `GuiRobotController`. Similar for `GuiHumanoidController`. (This hard-coding behavior is not new to this PR.) Later, GuiControllers shouldn't have direct access to `gui_input`; they should just use things like `_hint_walk_dir`.
* `AppStateRearrangeV2._get_user_key_down` is currently how I encapsulate user-specific keys for grasping and opening. For the case of a local second user, I internally remap keys to other keys in a hard-coded way. In the long run, local multi-user is probably only for testing purposes, so this hackiness is fine. Later, `_get_user_key_down` could encapsulate a second internal `gui_input` from a remote user. Alternately, instead of the `_get_user_key_down` interface, we may want to offer multiple `gui_input` instances, one per user. (In the case of local multi-user, the second one could be built from the "real" one using similar key-remapping hackiness.)

For GUI-controlled Spot, a few points:
1. I opt to use the Habitat-lab action space already configured for the Spot via the Habitat-lab config, in particular, the BaseVelocity skill. Lots of other skills are present in its action space and not used. We can imagine a different paradigm where we directly manipulate the articulated agent via Habitat-sim API.
2. In terms of turning/steering the Spot, I stick to the same convention we use for the Humanoid: "turning" is actually just turning the camera, and then we pass cam_yaw into the controller. The controller then makes a best effort to steer the agent to match the cam yaw. For Spot, doing this accurately unfortunately requires tight coupling between the GuiRobotController class and the underlying BaseVelocity action. In particular, see how `turn_scale` is computed in `controller_helper.py`. Instead of the controller steering to match camera yaw, we can imagine a different paradigm: turn inputs are forwarded to the controller, the controller is free to steer the agent however it likes, and then the camera yaw would follow the agent yaw.

## How Has This Been Tested

Local testing on Macbook:
```
python examples/hitl/rearrange_v2/rearrange_v2.py 
python examples/hitl/rearrange_v2/rearrange_v2.py +experiment=gui_controlled_spot_only
python examples/hitl/rearrange_v2/rearrange_v2.py +experiment=gui_controlled_humanoid_and_spot
python examples/hitl/rearrange/rearrange.py
python examples/hitl/basic_viewer/basic_viewer.py
python examples/hitl/pick_throw_vr/pick_throw_vr.py
python examples/hitl/minimal/minimal.py
```

## Types of changes

- [HITL]

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/rearrange_v2.py
+++ b/examples/hitl/rearrange_v2/rearrange_v2.py
@@ -24,6 +24,7 @@ from habitat_hitl.core.text_drawer import TextOnScreenAlignment
 from habitat_hitl.environment.camera_helper import CameraHelper
 from habitat_hitl.environment.controllers.gui_controller import (
     GuiHumanoidController,
+    GuiRobotController,
 )
 from habitat_hitl.environment.gui_navigation_helper import GuiNavigationHelper
 from habitat_hitl.environment.gui_pick_helper import GuiPickHelper
@@ -73,8 +74,6 @@ class AppStateRearrangeV2(AppState):
         self._client_helper = None
         if self._app_service.hitl_config.networking.enable:
             self._client_helper = ClientHelper(self._app_service)
-
-        self._gui_agent_ctrl.line_renderer = app_service.line_render
 
         self._has_grasp_preview = False
         self._frame_counter = 0
@@ -182,6 +181,11 @@ class AppStateRearrangeV2(AppState):
     def get_sim(self):
         return self._app_service.sim
 
+    def _get_gui_agent_translation(self):
+        return get_agent_art_obj_transform(
+            self.get_sim(), self.get_gui_controlled_agent_index()
+        ).translation
+
     def _update_grasping_and_set_act_hints(self):
         drop_pos = None
         grasp_object_id = None
@@ -194,13 +198,17 @@ class AppStateRearrangeV2(AppState):
             if self._app_service.gui_input.get_key_down(GuiInput.KeyNS.SPACE):
                 if DO_HUMANOID_GRASP_OBJECTS:
                     # todo: better drop pos
-                    drop_pos = self._gui_agent_ctrl.get_base_translation()
+                    drop_pos = (
+                        self._get_gui_agent_translation()
+                    )  # self._gui_agent_ctrl.get_base_translation()
                 else:
                     # GuiPlacementHelper has already placed this object, so nothing to do here
                     pass
                 self._held_obj_id = None
         else:
-            query_pos = self._gui_agent_ctrl.get_base_translation()
+            query_pos = (
+                self._get_gui_agent_translation()
+            )  # self._gui_agent_ctrl.get_base_translation()
             obj_id = self._pick_helper.get_pick_object_near_query_position(
                 query_pos
             )
@@ -230,7 +238,9 @@ class AppStateRearrangeV2(AppState):
         #     walk_dir = candidate_walk_dir
         #     distance_multiplier = candidate_distance_multiplier
 
-        assert isinstance(self._gui_agent_ctrl, GuiHumanoidController)
+        assert isinstance(
+            self._gui_agent_ctrl, (GuiHumanoidController, GuiRobotController)
+        )
         self._gui_agent_ctrl.set_act_hints(
             walk_dir,
             distance_multiplier,
--- a/examples/hitl/rearrange_v2/rearrange_v2.py
+++ b/examples/hitl/rearrange_v2/rearrange_v2.py
@@ -26,7 +26,6 @@ from habitat_hitl.environment.controllers.gui_controller import (
     GuiHumanoidController,
     GuiRobotController,
 )
-from habitat_hitl.environment.gui_navigation_helper import GuiNavigationHelper
 from habitat_hitl.environment.gui_pick_helper import GuiPickHelper
 from habitat_hitl.environment.gui_placement_helper import GuiPlacementHelper
 from habitat_hitl.environment.hablab_utils import get_agent_art_obj_transform
@@ -43,7 +42,8 @@ class AppStateRearrangeV2(AppState):
 
     def __init__(self, app_service):
         self._app_service = app_service
-        self._gui_agent_ctrl = self._app_service.gui_agent_controller
+        self._gui_agent_controllers = self._app_service.gui_agent_controllers
+        self._num_users = len(self._gui_agent_controllers)
         self._can_grasp_place_threshold = (
             self._app_service.hitl_config.can_grasp_place_threshold
         )
@@ -53,6 +53,7 @@ class AppStateRearrangeV2(AppState):
         self._opened_ao_set: Set = set()
 
         self._cam_transform = None
+        self._camera_user_index = 0
         self._held_obj_id = None
         self._recent_reach_pos = None
         self._paused = False
@@ -63,12 +64,8 @@ class AppStateRearrangeV2(AppState):
             self._app_service.gui_input,
         )
 
-        self._nav_helper = GuiNavigationHelper(
-            self._app_service, self.get_gui_controlled_agent_index()
-        )
         self._pick_helper = GuiPickHelper(
             self._app_service,
-            self.get_gui_controlled_agent_index(),
         )
         self._placement_helper = GuiPlacementHelper(self._app_service)
         self._client_helper = None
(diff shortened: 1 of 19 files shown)


examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -388,24 +406,36 @@ def sim_update(self, dt, post_sim_update_dict):
...
             self._app_service.compute_action_and_step_env()
         else:
             # temp hack: manually add a keyframe while paused
             self.get_sim().gfx_replay_manager.save_keyframe()
 
+        # todo: visualize objects properly for each user (this requires a separate debug_line_render per user!), or find a reasonable debug line visualization that can be shared between both users every frame.
         if self._held_obj_id is None:
             self._pick_helper.viz_objects()
 
+        if (
```
eundersander, line 430: For local multi-user testing, use the `T` key to toggle the camera between users. Toggling the camera while holding an object is super-confusing so I disabled this.

habitat-hitl/habitat_hitl/_internal/config_helper.py
```diff
@@ -98,7 +99,12 @@ def update_config(
                 if measurement_name in task_config.measurements:
                     task_config.measurements.pop(measurement_name)
 
-            sim_sensor_names = ["head_depth", "head_rgb"]
+            # todo: decide whether to fix up config here versus validate config
+            sim_sensor_names = [
```
eundersander, line 103: In the long run, I think it's hacky to "fix up" yaml configs at runtime. It creates confusion, because the config used at runtime doesn't match what the user authored in the yaml. It would be cleaner to instead just assert on the requirements here (thus forcing the user to fix up their config yaml).

habitat-hitl/habitat_hitl/environment/controllers/controller_helper.py
```diff
@@ -77,52 +78,113 @@ def __init__(
...
+                            original_action_space, "_base_velocity"
+                        )
+
+                        assert len(action_space.shape) == 1
+                        num_actions = action_space.shape[0]
+
+                        articulated_agent = self._env._sim.agents_mgr[agent_index].articulated_agent  # type: ignore[attr-defined]
+
+                        # sloppy: derive turn scale. This is the change in yaw (in radians) corresponding to a base ang vel action of 1.0. See also Habitat-lab BaseVelAction.
+                        turn_scale = (
```
eundersander, line 135: I discuss this more in the PR description.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -24,27 +23,106 @@
...
+        cam_yaw=None,
+        throw_vel=None,
+        reach_pos=None,
+        hand_idx=None,
+        target_dir=None,
+    ):
+        assert (
+            throw_vel is None or do_drop is None
+        ), "You can not set throw_velocity and drop_position at the same time"
+        self._hint_walk_dir = walk_dir
```
eundersander, line 64: This is sloppy. I'm not actually using `_hint_walk_dir` anywhere (yet). Likewise for `_hint_distance_multiplier` and `_hint_target_dir`.

eundersander (commented):
Some comments.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -24,27 +23,106 @@
...
+    def __init__(
+        self,
+        agent_idx,
+        is_multi_agent,
+        gui_input,
+        articulated_agent,
+        num_actions,
+        base_vel_action_idx,
+        num_base_vel_actions,
+        turn_scale,
```
0mdc, line 35: Types would be great here, as this is a dependency injection point.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -24,27 +23,106 @@
...
-            grasp = None
+        # if arm_name in ac_spaces:
+        #     arm_action_space = ac_spaces[arm_name][arm_k]
+        #     arm_ctrlr = env.task.actions[arm_name].arm_ctrlr
+        #     arm_action = np.zeros(arm_action_space.shape[0])
+        #     grasp = 0
+        # else:
+        #     arm_ctrlr = None
+        #     arm_action = None
+        #     grasp = None
```
0mdc, line 125: Could you add a small header comment to explain why this is commented?

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -56,82 +134,112 @@ def act(self, obs, env):
         KeyNS = GuiInput.KeyNS
         gui_input = self._gui_input
 
+        # Note addition of 180 degrees due to our camera convention
+        turn_angle = self.angle_from_sim_obj_foward_dir_to_target_yaw(
+            self._articulated_agent.sim_obj,
+            self._cam_yaw + float(mn.Rad(mn.Deg(180))),
+        )
```
0mdc, line 141: >180 degrees due to our camera convention

I may be confronted to a math error due to this. Could you provide some context on why the 180 degree rotation brings us to the desired result here?

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -161,18 +269,25 @@ def act(self, obs, env):
...
+        #         {
+        #             arm_k: arm_action,
+        #             grip_k: grasp,
+        #         }
+        #     )
         if len(action_names) == 0:
             raise ValueError("No active actions for human controller.")
 
-        return {"action": action_names, "action_args": action_args}
+        assert len(base_action) == self._num_base_vel_actions
```
0mdc, line 283: Could we rename `base_action` and type it? The current name doesn't convey that it can sometimes be a collection structure. This also makes it difficult to understand why this assert would break.

habitat-hitl/habitat_hitl/environment/controllers/controller_helper.py
```diff
@@ -77,52 +78,113 @@ def __init__(
...
+            gui_controlled_agent_config
+        ) in self._hitl_config.gui_controlled_agents:
+            if gui_controlled_agent_config.agent_index == agent_index:
+                return gui_controlled_agent_config
+        return None
+
+    def get_gui_agent_controllers(self) -> List[Controller]:
+        """
+        Return list of controllers indexed by user index. Beware the difference between user index and agent index. For example, user 0 may control agent 1.
+        """
```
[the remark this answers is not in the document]
eundersander: >agent index concept should be entirely abstracted away at the HITL level
I doubt this will be feasible. We shouldn't think of the controller as completely encapsulating the agent. It is really just encapsulating how we give actions to the agent. E.g. if you want to know something particular about the agent state (e.g. arm position), you would probably still use the Habitat-lab agent API including knowing the particular agent index.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
eundersander: Added a comment in code. This is unrelated to sensor conventions and strictly due to camera_helper.py implementation details.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
eundersander: I didn't rename or type this, but I simplified the code. It is always a numpy array. The assert would break if _num_base_vel_actions (passed in to `GuiRobotController.__init__`) doesn't match the base_velocity action space we find here in `act()`.

habitat-hitl/habitat_hitl/environment/controllers/gui_controller.py
```diff
@@ -24,114 +22,192 @@
...
+        assert grasp_obj_idx is None
+        assert do_drop is None
+        assert throw_vel is None
+        assert reach_pos is None
+        assert hand_idx is None
+
+        self._cam_yaw = cam_yaw
+        self._hint_target_dir = target_dir
+
+    def angle_from_sim_obj_foward_dir_to_target_yaw(self, sim_obj, target_yaw):
```
0mdc, line 76: Typo: `foward`

0mdc (approved):
LGTM! I tested it locally, and it works as expected.

PR #1833, opened 2024-03-05
[Task] Add Habitat-LLM episodes in rearrange_v2 app

Ram81:
## Motivation and Context

- Add support to load tasks from habitat-llm episode dataset.
- Add changes to pop_play config to only spawn a humaoid in the HITL app. This requires changing task config used in `pop_play.yaml` to initialize a single agent.
- The current implementation loads the episodes from habitat-llm with the help of `RearrangeEmptyTask-v0` as we are not planning to use PDDL metrics for evaluation. Disabling PDDL metrics and actions makes it difficult to initialize the default hierarchical policy available in `pop_play.yaml` so for now I have disabled initializing the spot agent in the HITL app. @eundersander @0mdc One way to get around this is to initialize a dummy hierarchical policy which does not use PDDL at all. This would be a temporary fix until we figure out how to integrate the habitat-llm policy into the HITL app. Let me know if this config looks good or we should switch to a dummy hierarchicaly policy

## How Has This Been Tested

Local testing on ubuntu machine. To test the PR you'll need the following dataset 
[habitat-llm.zip](https://github.com/facebookresearch/habitat-lab/files/14497256/habitat-llm.zip) and the object assets required can be downloaded using these commands

```
cd  data
git clone https://huggingface.co/datasets/ai-habitat/OVMM_objects objects --recursive
cd objects
git checkout 3893a735352b92d46505f35d759553f5fc82a39b
```

To test the app use:

```
 python examples/hitl/rearrange_v2/rearrange_v2.py habitat_hitl.networking.enable=True
```

## Types of changes

[HITL]

## Checklist


examples/hitl/rearrange_v2/config/task.yaml
```diff
@@ -0,0 +1,87 @@
+# @package _global_
```
eundersander, line 1: Similar requests for this config:
* add a comment explaining what Habitat-lab config this was derived from.
* name is something meaningful. "task" is vague. Maybe this is a specific task, e.g. instruction-following or social rearrange?
* strip it down to the essential config fields. E.g. you name a `dataset.data_path` here that I don't think actually gets used.

eundersander (commented):
Some comments below. A few more requests:
* test with GUI-controlled spot-only (single-user use case) and GUI-controlled spot+humanoid (multi-user use case), and update the PR description with the terminal command for these
* Also update the rearrange_v2 readme with the terminal commands for these two use cases.
* Make sure CI pre-commit and lint pass. You should have our precommit hooks installed locally which makes this really easy.

examples/hitl/rearrange_v2/README.md
```diff
@@ -16,6 +16,20 @@ mv fphab fpss
 
 Run `rearrange_v2` from the Habitat-lab root directory. It will expect `data/` for Habitat-lab data, and it will also look for `examples/hitl/rearrange_v2/app_data/demo.json.gz` (included alongside source files in our git repo).
 
+
+To test the Habitat-LLM episodes in `rearrange_v2` you'll need to download and unzip the following [episode dataset](https://drive.google.com/file/d/1zFCBiWE_XFY0Ry9CZOV_NF_rfxBw1y-F/view?usp=sharing) in Habitat-Lab root directory. In addition, you'll need YCB, GSO, AI2THOR, and ABO object assets. To download these assets use the following commands:
```
eundersander, line 20: This should part of the Installation section, not the Data directory section.

examples/hitl/rearrange_v2/README.md
```diff
@@ -33,6 +47,11 @@ Headless server:
 python examples/hitl/rearrange_v2/rearrange_v2.py +experiment=headless_server
 ```
 
+Testing Habitat-LLM episodes using a headed server:
```
eundersander, line 50: This shouldn't require `habitat_hitl.networking.enable=True` (this is a recent fix).

Also, can you mention: this is testing Habitat-LLM episodes with only a user-controlled humanoid (no Spot).

examples/hitl/rearrange_v2/README.md
```diff
@@ -33,6 +47,11 @@ Headless server:
...
+```bash
+python examples/hitl/rearrange_v2/rearrange_v2.py --config-name rearrange_v2_humanoid_only habitat_hitl.networking.enable=True
+```
+
```
eundersander, line 54: Can you add instructions for testing Habitat-llm episodes with a user-controlled humanoid and user-controlled spot? @0mdc will need this.

eundersander (commented):
See comments below. The main thing I'm still looking for is the command to test the habitat-llm episodes with a user-controlled humanoid and spot (Mikael will need this).

examples/hitl/rearrange_v2/README.md
Ram81: Added command to test it with user-controlled humanoid. For humanoid+spot I'll add it with habitat-llm policy changes

examples/hitl/rearrange_v2/README.md
Ram81: Done

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -316,6 +328,8 @@ def _update_help_text(self):
             self._app_service.text_drawer.add_text(
                 status_str,
                 TextOnScreenAlignment.TOP_CENTER,
+                text_delta_x=-280,
```
akshararai, line 331: lets make these configurable. I imagine the number is different for different displays?

akshararai (approved):
minor comments but looks good, thanks!

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: I wouldn't stress about it too much right now.

This `text_drawer` class has pretty rudimentary text layout formatting. For instance, we cannot do things like text alignment or precise positioning.

We're looking to improve formatting here to avoid this kind of fiddling.

PR #1835, opened 2024-03-06
"Ontop" utility

aclegg3:
## Motivation and Context

This PR adds a utility for querying which objects are "ontop" of a subject object. Initial prototype in #1770. 

Here, "ontop" means the objects have contacting collision points with a close to vertical contact normal. The epsilon threshold for the dot product is a controllable variable, defaulting to 0.75.

The function inputs can be `ManagedObject`s or `object_id` integers. If the `object_id` provided refers to an `ArticulatedLink`, _only that link_ is checked. Conversely, if a ManagedArticulatedObject is provided, _all links_ are checked.

## How Has This Been Tested

A pytest is added using all input variant in a ReplicaCAD scene.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
+++ b/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
@@ -759,8 +759,8 @@ def ontop(
     objectA: Union[
         habitat_sim.physics.ManagedRigidObject,
         habitat_sim.physics.ManagedArticulatedObject,
+        int,
     ],
-    ao_link_map: Optional[Dict[int, int]] = None,
     do_collision_detection: bool = True,
     vertical_normal_error_threshold: float = 0.75,
 ) -> List[int]:
@@ -770,16 +770,22 @@ def ontop(
     This function uses collision points to determine which objects are resting on or contacting the surface of objectA.
 
     :param sim: The Simulator instance.
-    :param objectA: The ManagedRigidObject for which to query the 'ontop' set.
-    :param ao_link_map: A pre-computed map from link object ids to their parent ArticulatedObject's object id.
+    :param objectA: The ManagedRigidObject or object id for which to query the 'ontop' set.
     :param do_collision_detection: If True, a fresh discrete collision detection is run before the contact point query. Pass False to skip if a recent sim step or pre-process has run a collision detection pass on the current state.
     :param vertical_normal_error_threshold: The allowed error in normal alignment for a contact point to be considered "vertical" for this check. Functionally, if dot(contact normal, Y) <= threshold, the contact is ignored.
 
     :return: a list of integer object_ids for the set of objects "ontop" of objectA.
     """
 
-    if ao_link_map is None:
-        ao_link_map = get_ao_link_id_map(sim)
+    link_id = None
+    if isinstance(objectA, int):
+        subject_object = get_obj_from_id(sim, objectA)
+        if subject_object is None:
+            raise AssertionError(f"The passed object_id {objectA} is invalid.")
+        if subject_object.object_id != objectA:
+            # objectA is a link
+            link_id = subject_object.link_object_ids[objectA]
+        objectA = subject_object
 
     if do_collision_detection:
         sim.perform_discrete_collision_detection()
@@ -790,9 +796,13 @@ def ontop(
     for cp in sim.get_physics_contact_points():
         contacting_obj_id = None
         obj_is_b = False
-        if cp.object_id_a == objectA.object_id:
+        if cp.object_id_a == objectA.object_id and (
+            link_id is None or link_id == cp.link_id_a
+        ):
             contacting_obj_id = cp.object_id_b
-        if cp.object_id_b == objectA.object_id:
+        elif cp.object_id_b == objectA.object_id and (
+            link_id is None or link_id == cp.link_id_b
+        ):
             contacting_obj_id = cp.object_id_a
             obj_is_b = True
         if contacting_obj_id is not None:
@@ -806,6 +816,8 @@ def ontop(
                 > vertical_normal_error_threshold
             ):
                 ontop_object_ids.append(contacting_obj_id)
+
+    ontop_object_ids = list(set(ontop_object_ids))
 
     return ontop_object_ids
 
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
     return containment_ids
 
 
+def ontop(
+    sim: habitat_sim.Simulator,
+    objectA: Union[
```
0mdc, line 759: Nit: I would use python case here (`object_a`).

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
...
+        obj_is_b = False
+        if cp.object_id_a == objectA.object_id and (
+            link_id is None or link_id == cp.link_id_a
+        ):
+            contacting_obj_id = cp.object_id_b
+        elif cp.object_id_b == objectA.object_id and (
+            link_id is None or link_id == cp.link_id_b
+        ):
+            contacting_obj_id = cp.object_id_a
+            obj_is_b = True
```
0mdc, line 807: I'm probably misunderstanding what's going on here.

Wouldn't we want to exclude the object itself from the set of objects on top of it? Would a `continue` be better suited here here?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
...
+        if contacting_obj_id is not None:
+            contact_dir_me = (
```
0mdc, line 809: Nit: `contact_normal` would probably be clearer here.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
...
 
 
+def ontop(
+    sim: habitat_sim.Simulator,
+    objectA: Union[
+        habitat_sim.physics.ManagedRigidObject,
+        habitat_sim.physics.ManagedArticulatedObject,
+        int,
+    ],
+    do_collision_detection: bool = True,
```
0mdc, line 764: Beware that this default value may cause performance issues.

Folks may start using this function liberally to query the scene.

0mdc (approved):
Left some comments, but otherwise LGTM!

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
...
+    Get a list of all object ids or objects that are "ontop" of a particular objectA.
+    Concretely, 'ontop' is defined as: contact points between objectA and objectB have vertical normals "upward" relative to objectA.
+    This function uses collision points to determine which objects are resting on or contacting the surface of objectA.
+
+    :param sim: The Simulator instance.
+    :param objectA: The ManagedRigidObject or object id for which to query the 'ontop' set.
+    :param do_collision_detection: If True, a fresh discrete collision detection is run before the contact point query. Pass False to skip if a recent sim step or pre-process has run a collision detection pass on the current state.
+    :param vertical_normal_error_threshold: The allowed error in normal alignment for a contact point to be considered "vertical" for this check. Functionally, if dot(contact normal, Y) <= threshold, the contact is ignored.
+
+    :return: a list of integer object_ids for the set of objects "ontop" of objectA.
```
0mdc, line 777: Will users ever need to check whether a specific `object_b` is on top of `object_a`, and get a simple bool as a result? A variant function may be cheaper if they need to do this comparison often.

0mdc (commented):
Some other thoughts.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -754,6 +754,74 @@ def within(
...
+        subject_object = get_obj_from_id(sim, objectA)
+        if subject_object is None:
+            raise AssertionError(f"The passed object_id {objectA} is invalid.")
+        if subject_object.object_id != objectA:
+            # objectA is a link
+            link_id = subject_object.link_object_ids[objectA]
+        objectA = subject_object
+
+    if do_collision_detection:
+        sim.perform_discrete_collision_detection()
```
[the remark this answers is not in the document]
aclegg3: Is this really so expensive to warrant the effort and complexity? I haven't profiled, but it's just iterating over the internal structures and collecting the already computed data into a new structure to return. :thinking:

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: It could be a bit cheaper, but you'd still need to look through every contact point, so I don't think it would be much faster. I guess the advantage would the option for early abort.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Good point, we can remove the default here.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Yeah, possibly misunderstanding. For clarity, we don't know if our object is the first or second in the contact pair. This statement just figures out which one it is if either and prepares to invert the normal if necessary.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: I'll leave this to a future PR. Early abort and a check against provided object_id should be easy enough to add in a small refactor if we want it.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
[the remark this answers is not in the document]
aclegg3: If this does end up being too costly we'd likely add the cache in sim, so I'll defer this to a future PR.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
0mdc: Sounds good.
My worry is that users would also have to find the desired object within the results - the search itself could be called in a nested loop, making this overall process inefficient.
I agree with you - let's see how this ends up being used.

PR #1838, opened 2024-03-07
HITL - Add GUI rendering wrapper.

0mdc:
## Motivation and Context

This adds a wrapper for `DebugLineRender`, which also supports networking via `ClientMessageManager`.
* If `DebugLineRender` is `None` (headless mode), server rendering is skipped.
* If `ClientMessageManager` is `None` (local mode), remote rendering is skipped.

Rendering for multiple clients would be supported by either adding a "destination" parameter to `GuiDrawer` functions (relayed to `ClientMessageManager`), or carrying an array of `GuiDrawers` and `ClientMessageManagers`. I think that the former would be simpler in the long run.

## How Has This Been Tested

Tested in all apps along with the Unity mouse/keyboard client.

## Types of changes

- **\[Refactoring\]**
- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/basic_viewer/basic_viewer.py
+++ b/examples/hitl/basic_viewer/basic_viewer.py
@@ -69,7 +69,7 @@ class AppStateBasicViewer(AppState):
 
         # draw lookat point
         radius = 0.15
-        self._app_service.line_render.draw_circle(
+        self._app_service.gui_drawer.draw_circle(
             self._get_camera_lookat_pos(),
             radius,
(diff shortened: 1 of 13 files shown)


aclegg3 (approved):
Looks good.
I was thinking about ways to reduce the API forwarding/mirroring.
Talking to @0mdc we concluded that this intermediate class will have expanded function variable set and a branched implementation for client-side interface, so this mirror is a lesser of two evils for now.

PR #1841, opened 2024-03-09
"on_floor" query util

aclegg3:
## Motivation and Context

This PR adds the `on_floor` utility function for heuristically checking whether an object is "on the floor".

Context: we want to know if objects are on the navigable area of the floor for rearrangement. This function checks that the heuristic point-to-surface distance between the object and its nearest navmesh snap point is within an error threshold.

Supporting utilities:
- `get_bb_for_object_id` - a generic utility for getting the local bounding box and transform of any object or link from its object_id integer.
- `get_obj_size_along` - a function to approximate the object's size in a particular global direction from its bounding box size. Uses aabb scale to transform the vector, essentially finding a point on the best-fit ellipsoid within the aabb.
-  `size_regularized_distance` - a function to get estimated surface-to-surface distance between two objects or links using the `get_obj_size_along` function to truncate the center-to-center distance between the objects.

Also includes a few minor refactors to other utilities.

## How Has This Been Tested

Added a set of pytests.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
+++ b/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
@@ -197,7 +197,7 @@ def get_obj_size_along(
     ao_aabbs: Dict[int, mn.Range3D] = None,
 ) -> Tuple[float, mn.Vector3]:
     """
-    Uses object bounding box as a heuristic to estimate object size in a particular global direction.
+    Uses object bounding box ellipsoid scale as a heuristic to estimate object size in a particular global direction.
 
     :param sim: The Simulator instance.
     :param object_id: The integer id of the object or link.
@@ -205,7 +205,7 @@ def get_obj_size_along(
     :param ao_link_map: A pre-computed map from link object ids to their parent ArticulatedObject's object id.
     :param ao_aabbs: A pre-computed map from ArticulatedObject object_ids to their local bounding boxes. If not provided, recomputed as necessary.
 
-    :return: distance along the specified direction and center of bounding box from which distance was estimated.
+    :return: distance along the specified direction and global center of bounding box from which distance was estimated.
     """
 
     obj_bb, transform = get_bb_for_object_id(
@@ -213,7 +213,7 @@ def get_obj_size_along(
     )
     center = transform.transform_point(obj_bb.center())
     local_scale = mn.Matrix4.scaling(obj_bb.size() / 2.0)
-    local_vec = transform.inverted().transform_vector(global_vec)
+    local_vec = transform.inverted().transform_vector(global_vec).normalized()
     local_vec_size = local_scale.transform_vector(local_vec).length()
     return local_vec_size, center
 
@@ -962,6 +962,58 @@ def ontop(
     ontop_object_ids = list(set(ontop_object_ids))
 
     return ontop_object_ids
+
+
+def on_floor(
+    sim: habitat_sim.Simulator,
+    object_a: habitat_sim.physics.ManagedRigidObject,
+    distance_threshold: float = 0.04,
+    alt_pathfinder: habitat_sim.nav.PathFinder = None,
+    island_index: int = -1,
+    ao_link_map: Dict[int, int] = None,
+    ao_aabbs: Dict[int, mn.Range3D] = None,
+) -> bool:
+    """
+    Gets whether or not the object is on the "floor" using the navmesh as an abstraction.
+    NOTE: alt_pathfinder option can be used to provide an alternative navmesh sized for objects. This would allow objects to be, for example, under tables or in corners and still be considered on the navmesh.
+
+    :param sim: The Simulator instance.
+    :param object_a: The object instance.
+    :param distance_threshold: Maximum allow-able displacement between current object position and navmesh snapped position.
+    :param alt_pathfinder:Optionally provide an alternative PathFinder specifically configured for this check. Defaults to sim.pathfinder.
+    :param island_index: Optionally limit allowed navmesh to a specific island. Default (-1) is full navmesh.
+    :param ao_link_map: A pre-computed map from link object ids to their parent ArticulatedObject's object id.
+    :param ao_aabbs: A pre-computed map from ArticulatedObject object_ids to their local bounding boxes. If not provided, recomputed as necessary.
+
+    :return: Whether or not the object is considered "on the floor" given the configuration.
+    """
+
+    assert isinstance(
+        object_a, habitat_sim.physics.ManagedRigidObject
+    ), "Object must be ManagedRigidObject, not implemented for ArticulatedObjects or links."
+
+    if alt_pathfinder is None:
+        alt_pathfinder = sim.pathfinder
+
+    assert alt_pathfinder.is_loaded
+
+    # use the object's heuristic size to estimate distance from the object center to the navmesh in order to regularize the navigability constraint for larger objects
+    obj_size, center = get_obj_size_along(
+        sim,
+        object_a.object_id,
+        mn.Vector3(0.0, -1.0, 0.0),
+        ao_link_map=ao_link_map,
+        ao_aabbs=ao_aabbs,
+    )
+
+    obj_snap = alt_pathfinder.snap_point(center, island_index=island_index)
+
+    # include navmesh cell height error in the distance threshold.
+    navmesh_cell_height = alt_pathfinder.nav_mesh_settings.cell_height
+    snap_disp = obj_snap - center
+    snap_dist = snap_disp.length() - obj_size - (navmesh_cell_height / 2.0)
+
+    return snap_dist <= distance_threshold
 
 
 def object_in_region(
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -824,6 +964,58 @@ def ontop(
...
+    # use the object's heuristic size to estimate distance from the object center to the navmesh in order to regularize the navigability constraint for larger objects
+    obj_size, center = get_obj_size_along(
+        sim,
+        object_a.object_id,
+        mn.Vector3(0.0, -1.0, 0.0),
+        ao_link_map=ao_link_map,
+        ao_aabbs=ao_aabbs,
+    )
+
+    obj_snap = alt_pathfinder.snap_point(center, island_index=island_index)
```
jacobkrantz, line 1009: on_floor follows a very different logic than ontop. Why do we use a relaxed center point test instead of island-object contact points?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -139,6 +139,140 @@ def get_ao_global_bb(
...
+    return local_vec_size, center
+
+
+def size_regularized_distance(
+    sim: habitat_sim.Simulator,
+    object_id_a: int,
+    object_id_b: int,
+    ao_link_map: Dict[int, int] = None,
+    ao_aabbs: Dict[int, mn.Range3D] = None,
+) -> float:
```
jacobkrantz, line 227: This function isn't used in on_floor. Do we have a use case for it? Why introduce it with on_floor?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -139,6 +139,140 @@ def get_ao_global_bb(
...
+        return (
+            sim.get_active_scene_graph().get_root_node().cumulative_bb,
+            mn.Matrix4.identity_init(),
+        )
+
+    obj = get_obj_from_id(sim, obj_id, ao_link_map)
+
+    assert (
+        obj is not None
+    ), f"object id {obj_id} is not found, this is unexpected. Invalid/stale object id?"
```
jacobkrantz, line 171: Should `raise` an error instead of calling `assert`. Assert gets skipped outside of debug mode.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -824,6 +964,58 @@ def ontop(
...
+    :param island_index: Optionally limit allowed navmesh to a specific island. Default (-1) is full navmesh.
+    :param ao_link_map: A pre-computed map from link object ids to their parent ArticulatedObject's object id.
+    :param ao_aabbs: A pre-computed map from ArticulatedObject object_ids to their local bounding boxes. If not provided, recomputed as necessary.
+
+    :return: Whether or not the object is considered "on the floor" given the configuration.
+    """
+
+    assert isinstance(
+        object_a, habitat_sim.physics.ManagedRigidObject
+    ), "Object must be ManagedRigidObject, not implemented for ArticulatedObjects or links."
```
jacobkrantz, line 993: Just checking -- do we have use cases for AO on_floor?

jacobkrantz (commented):
Thanks, Alex! Left some comments/questions.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Yeah, we could wait to merge this with "nearby". That is were it really belongs.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Primarily because we don't have a unique object id for the floor vs. the wall or other architectural feature.
We could check contact points, but we'd need some logic to disambiguate those cases.
This also allows the check to be restricted to a specified island so it can semantically double as "on reachable floor".

I'm open to thoughts if you think this doesn't match what we really want it to.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: I don't think so because we don't support AO "pick-able" objects presently. I currently assume all AOs are furniture or robot/human.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
jacobkrantz: I was just wanting to understand. This makes sense, thanks.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -139,6 +139,140 @@ def get_ao_global_bb(
...
+    ao_aabbs: Dict[int, mn.Range3D] = None,
+) -> Tuple[mn.Range3D, mn.Matrix4]:
+    """
+    Wrapper to get a bb and global transform directly from an object id.
+    Handles RigidObject and ArticulatedLink ids.
+
+    :param sim: The Simulator instance.
+    :param obj_id: The integer id of the object or link.
+    :param ao_link_map: A pre-computed map from link object ids to their parent ArticulatedObject's object id.
+    :param ao_aabbs: A pre-computed map from ArticulatedObject object_ids to their local bounding boxes. If not provided, recomputed as necessary.
```
akshararai, line 155: for my understanding, this is bounding box for a closed articulated object?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -139,6 +139,140 @@ def get_ao_global_bb(
...
+
+def size_regularized_distance(
+    sim: habitat_sim.Simulator,
+    object_id_a: int,
+    object_id_b: int,
+    ao_link_map: Dict[int, int] = None,
+    ao_aabbs: Dict[int, mn.Range3D] = None,
+) -> float:
+    """
+    Get the heuristic surface-to-surface distance between two objects (regularized by their individual heuristic sizes).
```
akshararai, line 229: this is not surface-to-surface but more center-to-center since its the center of the bouinding box?

akshararai: Ah I see the final return which is from surface to surface.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -824,6 +964,58 @@ def ontop(
     return ontop_object_ids
 
 
+def on_floor(
```
akshararai, line 967: so to be clear, this is more implementing on_navmesh right?
Maybe we can add a comment that the floor isn't a unique receptacle, but approximated as the whole navmesh right now?
Sorry if I'm misunderstanding the code.

akshararai (approved):

akshararai (approved):
looks good overall.
Just a clarification: the floor and the navmesh are the same in this abstraction? If yes, does this affect how we initialize objects on floor? Would they be initialized anywhere on the navmesh, which might include things like unsegmented furniture?

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: This is a map of all AO bounding boxes for the scene. It includes the state of the links when computed. 
Typically, if pre-computed at initialization it will be for closed AOs (unless non-default AO states are specified in the scene instance config).
If it is computed in this function or re-computed during simulation, it will include the open state of the receptacles.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Yes, this is navmesh based solution. Happy to buff up the documentation a bit here.

aclegg3:
> Just a clarification: the floor and the navmesh are the same in this abstraction? If yes, does this affect how we initialize objects on floor? Would they be initialized anywhere on the navmesh, which might include things like unsegmented furniture?

Correct, the navmesh is our model of the valid floor space here.

It is true that an inaccurate navmesh could result in false positives. Like you say, using the entire navmesh could, for example, allow bed or sofa tops to be considered floor in some scenes.

There are controls available to prevent this. Primarily the `island_index` variable. We already assume in many other parts of the RearrangeTask/RearrangeSim/RearrangeGenerator code that we restrict the navmesh to a single island (the largest indoor island). This should also be used here to provide the best results.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Actually, this is already cleaned and tested. I'm going to merge it now given that nearby will be added soon. Less work in the long-run. Would be different if not tested.

PR #1842, opened 2024-03-09
ObjectStateSpec, ObjectStateMachine base classes

aclegg3:
## Motivation and Context

Context: we want to simulate more abstract states than physics/geometry. These should be event driven (e.g. "wash dish" action) or temporally driven (e.g. state machine logic in step())

Two identified initial states examples:
- "dish" objects have dirty/clean states and there is some state-machine logic to evolve this state 
  - e.g. "dirty" object in sink or dishwasher when "clean()" action is activated
- "appliance" objects have "on/off" states which can be set from an action

~Notes on Implementation:~
- ~If the states are semantically driven, we need something like the MetadataInterface class in habitat-lab to manage registration and set membership. Another option would be to embed the semantic classes directly into the dataset configs so they would be parsed and available from sim.~
- ~Current concrete implementations are examples and should be updated to reflect true logic better before merging~
- ~Current module locations within habitat_simulator module are best guess, open to discussion.~

Update: this PR now contains only the foundation of the feature and basic unit tests. MetadataInterface and original example logic is removed for now.

## How Has This Been Tested

added basic tests with contrived scenarios

## Types of changes

## Checklist


habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
```diff
@@ -0,0 +1,468 @@
...
+    """
+
+    user_attr = obj.user_attributes
+    obj_state_config = None
+    if "object_states" not in user_attr.get_subconfig_keys():
+        obj_state_config = (
+            habitat_sim._ext.habitat_sim_bindings.Configuration()
+        )
+    else:
+        obj_state_config = user_attr.get_subconfig("object_states")
```
jturner65, line 64: You only need this line.  It will either retrieve or create and then retrieve a config with the given name.

habitat-lab/habitat/sims/habitat_simulator/debug_visualizer.py
```diff
@@ -77,6 +78,43 @@ def save(self, output_path: str, prefix: str = "") -> str:
         return file_path
 
 
+def draw_object_highlight(
```
ASzot, line 81: Can we control the thickness of the highlight as well? I imagine this can be important for some applications based on the camera resolution.

habitat-lab/habitat/sims/habitat_simulator/metadata_interface.py
```diff
@@ -0,0 +1,258 @@
...
+# TODO: this is not a lab dependency currently, should probably remove it
+import pandas as pd
+
+import habitat.datasets.rearrange.samplers.receptacle as hab_receptacle
+import habitat.sims.habitat_simulator.sim_utilities as sutils
+import habitat_sim
+from habitat_sim.physics import ManagedArticulatedObject, ManagedRigidObject
+
+
+class MetadataInterface:
```
ASzot, line 20: Should this be in Habitat-Lab codebase? This feels very application specific to me.

habitat-lab/habitat/sims/habitat_simulator/metadata_interface.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def __init__(self, metadata_source_dict: Dict[str, str]) -> None:
+        # digest the HSSD metadata
+        self.metadata_source_dict = metadata_source_dict
+        self.metadata = self.load_metadata(self.metadata_source_dict)
+
+        self.receptacles: List[hab_receptacle.Receptacle] = None
+
+        # generate a lexicon from the metadata
+        self.hash_to_cat: Dict[str, str] = {}
```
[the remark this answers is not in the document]
aclegg3: I don't think we have any limitations as long as we can represent the desired hierarchy or grouping in the MetadataInterface logic. Currently we only pull the primary category and search a single map. I imagine we need to define these groupings or hierarchy somewhere. Could be an additional annotation field in the csv, or a dict-like map.

habitat-lab/habitat/sims/habitat_simulator/metadata_interface.py
aclegg3: Agreed. At this point it is only used map semantic classes to states. That doesn't need to be a base functionality. We could merge the machinery and use Habitat semantic_scene for tests. Then we'd extend the state classes to include what we want in downstream code.

habitat-lab/habitat/sims/habitat_simulator/debug_visualizer.py
aclegg3: I think this can be done in the debug_line_render object before it is passed into this function. https://aihabitat.org/docs/habitat-sim/habitat_sim.gfx.DebugLineRender.html#set_line_width 

So it shouldn't need to be explicitly controlled here. Rather the user application can set that for all debug renders.

habitat-lab/habitat/sims/habitat_simulator/metadata_interface.py
aclegg3: I've stripped this out for this current PR and unit tests.

habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def default_value(self) -> Any:
+        """
+        Default value for power is off.
+        """
+
+        return False
+
+
+class ObjectSateMachine:
```
rutadesai, line 249: typo.

habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+        Register a single object in the 'objects_with_states' dict by checking 'is_affordance_of_obj' for all active ObjectStateSpecs.
+        Use this when a new object is added to the scene and needs to be registered.
+
+        :param obj: The ManagedObject instance to register.
+        """
+
+        for state in self.active_states:
+            if state.is_affordance_of_obj(obj):
+                if obj.handle not in self.objects_with_states:
```
rutadesai, line 284: Not sure I understand this. Why not do:
`if obj.handle in self.objects_with_states:
      self.objects_with_states[obj.handle].append(state)`

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -13,6 +13,18 @@
 from habitat.sims.habitat_simulator.debug_visualizer import DebugVisualizer
 
 
+def object_shortname_from_handle(object_handle: str) -> str:
```
rutadesai, line 16: we have a similar function in hab-llm: https://github.com/facebookresearch/habitat-llm/blob/main/dataset_generation/benchmark_generation/generate_instructions.py#L62 Not sure if it is worth making this a utils function in one place..

rutadesai (approved):
LGTM! Looking forward to extending this to one to two instantiations pretty soon and also its integration in benchmark generation :) Thanks, Alex!

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Agreed. Once this merges, I'll refactor downstream code to use this util.

PR #1844, opened 2024-03-10
Fix agent GPS compass key in multi-agent README

gchhablani:
This PR fixes the incorrect key in docs.

`agent_0_nav_goal_sensor` is not a key in the observations, but it is `agent_0_goal_to_agent_gps_compass`  instead.


jimmytyyang (approved):
LGTM!

Issue #1804, opened 2024-02-13
`get_camera_transform` from `habitat.tasks.rearrange.utils` uses inconsistent getting method


zephirefaith:
## Habitat-Lab and Habitat-Sim versions
Habitat-Lab: v0.3.0 or master?

Habitat-Sim: v0.3.0 or master?

Habitat is under active development, and we advise users to restrict themselves to stable releases of [Habitat-Lab](https://github.com/facebookresearch/habitat-lab/releases) and [Habitat-Sim](https://github.com/facebookresearch/habitat-sim/releases). The bug you are about to report may already be fixed in the latest version.

Master branch contains 'bleeding edge' code, but we do appreciate bug reports for it!

## 🐛 Bug

The cameras attached to articulated agents are initialized using ArticulatedAgentCameraParams class [here](https://github.com/facebookresearch/habitat-lab/blob/main/habitat-lab/habitat/articulated_agents/mobile_manipulator.py#L19). The position for this sensor is set using the logic [here](https://github.com/facebookresearch/habitat-lab/blob/d4df9cc76cd3e5d66b24fbd92d004984a203196d/habitat-lab/habitat/articulated_agents/manipulator.py#L168-L183). However when `habitat.tasks.rearrange.utils.get_camera_transform` function is used to get the absolute transformation of this camera, the getter applies inconsistent logic taking only the translation component of camera's view-matrix. This leads to a camera-pose with right translation but wrong rotation wrt world frame.

## Steps to Reproduce

1. Initialize environment with Spot in it
2. Access camera pose using `get_camera_transform`
3. Access camera pose using `sim.agents[0]._sensors[<sensor-name>].render_camera.camera_matrix.inverted()`
4. Output of 2 != output of 3

See PCD built with 2 versus with 3 below.
<img width="461" alt="Screenshot 2024-02-13 at 2 43 44 PM" src="https://github.com/facebookresearch/habitat-lab/assets/9580531/590899a5-5855-449f-8ce5-50058d21fd3c">
<img width="459" alt="Screenshot 2024-02-13 at 2 43 55 PM" src="https://github.com/facebookresearch/habitat-lab/assets/9580531/58d9dda9-0336-4d76-8156-983eabf71434">

SergioArnaud:
🤯

aclegg3:
Yeah, this is a pain. @SergioArnaud happy to review a fix if you want to tackle this. :smiley:

zephirefaith:
[...]

PR #1847, opened 2024-03-11
Fixing rotational bug in `tasks.rearrange.utils.get_camera_transform`

zephirefaith:
## Motivation and Context

Fixes issue #1804 

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/utils.py
+++ b/habitat-lab/habitat/tasks/rearrange/utils.py
@@ -13,11 +13,10 @@ from functools import wraps
 from typing import List, Optional, Tuple
 
 import attr
+import habitat_sim
 import magnum as mn
 import numpy as np
 import quaternion
-
-import habitat_sim
 from habitat.articulated_agents.mobile_manipulator import MobileManipulator
 from habitat.articulated_agents.robots.spot_robot import SpotRobot
 from habitat.articulated_agents.robots.stretch_robot import StretchRobot
@@ -741,8 +740,13 @@ def get_camera_transform(cur_articulated_agent) -> mn.Matrix4:
         cam_info.attached_link_id
     ).transformation
     # Get the camera offset transformation
-    offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
-    cam_trans = link_trans @ offset_trans @ cam_info.relative_transform
+    # offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
+    cam_offset_transform = mn.Matrix4.look_at(
+        cam_info.cam_offset_pos,
+        cam_info.cam_look_at_pos,
+        mn.Vector3(0, 1, 0),
+    )
+    cam_trans = link_trans @ cam_offset_transform @ cam_info.relative_transform
     return cam_trans
 
 


habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -741,8 +745,23 @@ def get_camera_transform(cur_articulated_agent) -> mn.Matrix4:
         cam_info.attached_link_id
     ).transformation
     # Get the camera offset transformation
-    offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
-    cam_trans = link_trans @ offset_trans @ cam_info.relative_transform
+    # offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
+    cam_offset_transform = None
+    if cam_info.cam_look_at_pos == mn.Vector3(0, 0, 0):
```
aclegg3, line 750: Can you clarify or comment the two options here?
Is `cam_look_at_pos` in local space of the camera? Why not use look_at all the time? Is Vector3(0,0,0) a special case (like -1 for unused?)

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -741,8 +745,23 @@ def get_camera_transform(cur_articulated_agent) -> mn.Matrix4:
         cam_info.attached_link_id
     ).transformation
     # Get the camera offset transformation
-    offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
-    cam_trans = link_trans @ offset_trans @ cam_info.relative_transform
+    # offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
```
aclegg3, line 748: Stale comment should be removed before merge

jimmytyyang (approved):
LGTM! If would be nice if you could add a comment or flag about how this was computed before so that I can make sure my skill training code is correct and I will fix them.

zephirefaith:
Based on offline conversations:
- @aclegg3's concern that the current code uses value of `(0, 0, 0)` as an unassigned default while it is a valid setting. Should be replaced with `None` instead.
- Let's wrap `get_camera_from_link` logic as a util and use it consistently in setter + getter in: `habitat.tasks.rearrange.utils.get_camera_transform`, `kinematic_humanoid.update` and `manipulator.update`
- Add a print warning notifying change in logic and a `use_deprecated_logic` flag to make sure logic change does not break things for other users who depend on old logic (CC: @jimmytyyang )

zephirefaith:
@jimmytyyang @xavierpuigf Please take a look at this change as it directly touches your agent functions.

habitat-lab/habitat/articulated_agents/utils.py
```diff
@@ -0,0 +1,44 @@
...
+        # )
+    else:
+        link_trans = articulated_agent.sim_obj.get_link_scene_node(
+            cam_info.attached_link_id
+        ).transformation
+    # Get the camera offset transformation
+    cam_offset_transform = None
+    if cam_info.cam_look_at_pos == None:
+        # if cam_info.cam_look_at_pos is None that means the camera's pose-transform
+        # is described as a position and orientation in the world
```
aclegg3, line 27:
```suggestion
        # is described as a position and orientation in the local coordinate space of the parent link or object
```

habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -736,13 +751,21 @@ def get_camera_transform(cur_articulated_agent) -> mn.Matrix4:
...
-    # Get the camera's attached link
-    link_trans = cur_articulated_agent.sim_obj.get_link_scene_node(
-        cam_info.attached_link_id
-    ).transformation
-    # Get the camera offset transformation
-    offset_trans = mn.Matrix4.translation(cam_info.cam_offset_pos)
-    cam_trans = link_trans @ offset_trans @ cam_info.relative_transform
+    cam_trans = None
+
+    if deprecated_logic:
```
aclegg3, line 756: Interesting strategy. Do we need this? Are there active down-stream use cases we are tracking? 

IMO We can change the logic without back-compat and mark this PR as a breaking change since it is so close to the last release.

aclegg3 (commented):
This looks much better, thanks for the fix. :+1: 

I'll request that you add some typing/docs and I suggest you don't worry about backward compatibility unless there is a known down-stream users to accommodate. 
We can merge when @xavierpuigf approves.

As an aside, I think this util function could be unit tested in [test_robot_wrapper](https://github.com/facebookresearch/habitat-lab/blob/main/test/test_robot_wrapper.py).

habitat-lab/habitat/tasks/rearrange/utils.py
jimmytyyang: sounds good to me! priyam you made a call for this!

zephirefaith: Thanks! I'll take this out so we have a cleaner code-merge.

xavierpuigf (approved):
LGTM, thanks! Agreed that if we don't use the decprecated logic, it may be fine to remove it

PR #1868, opened 2024-03-18
HITL: Add client message destination mask.

0mdc:
## Motivation and Context

This is the first change that will allow the HITL framework to handle multiplayer.
It introduces a destination mask and a user container, both of which can be used to specify where to relay a specific message.

In this context, a "message" is a portion of a `gfx-replay` keyframe authored by the HITL server. The keyframes are global (they are relayed to all clients), but the message is client-specific.

Right now, by default, all messages are broadcasted. Since we only support 1 client. Therefore, this PR is essentially a noop.

## How Has This Been Tested

Tested locally and on multiplayer prototype.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
+++ b/habitat-hitl/habitat_hitl/_internal/hitl_driver.py
@@ -44,6 +44,7 @@ from habitat_hitl.core.serialize_utils import (
     save_as_pickle_gzip,
 )
 from habitat_hitl.core.text_drawer import AbstractTextDrawer
+from habitat_hitl.core.user_mask import Users
 from habitat_hitl.environment.controllers.controller_helper import (
     ControllerHelper,
 )
@@ -172,7 +173,9 @@ class HitlDriver(AppDriver):
 
         self._client_message_manager = None
         if self.network_server_enabled:
-            self._client_message_manager = ClientMessageManager()
+            # TODO: Only one user is currently supported.
+            users = Users(1)
+            self._client_message_manager = ClientMessageManager(users)
 
         gui_drawer = GuiDrawer(debug_line_drawer, self._client_message_manager)
         gui_drawer.set_line_width(self._hitl_config.debug_line_width)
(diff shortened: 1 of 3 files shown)


habitat-hitl/habitat_hitl/_internal/hitl_driver.py
```diff
@@ -538,10 +541,11 @@ def depth_to_rgb(obs):
                     if "rigUpdates" in keyframe_obj:
                         del keyframe_obj["rigUpdates"]
                 # Insert server->client message into the keyframe
-                message = self._client_message_manager.get_message_dict()
+                # TODO: Only one user is currently supported.
+                message = self._client_message_manager.get_messages()[0]
```
aclegg3, line 545: Should the `0` be a variable ("current_user" or something) instead of hard-coded to prepare for expansion?

habitat-hitl/habitat_hitl/core/user_mask.py
```diff
@@ -0,0 +1,145 @@
...
+class Users:
+    """
+    Represents a set of users with a max_user_count upper bound.
+    """
+
+    _max_user_mask: Mask
+    _max_user_count: int
+
+    def __init__(self, max_user_count: int) -> None:
+        assert max_user_count >= 0
```
aclegg3, line 60:
```suggestion
        assert max_user_count > 0
```
Should never be 0 users, right?

aclegg3 (approved):
LGTM, left some comments.

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
0mdc: The concept of `current_user` will be controlled by the application itself (user code). This specific snippet will always provide all user messages to the networking thread.

habitat-hitl/habitat_hitl/core/user_mask.py
0mdc: Right, this is a good sanity check.

While the code works for 0 users (noop), I don't currently see why it should be used as such.

PR #1871, opened 2024-03-20
"next_to" utility

aclegg3:
## Motivation and Context

This PR introduces **a utility function for checking whether two objects are "next to" one another** using bounding boxes and distance thresholds.

**Concrete criteria:**

1. threshold vertical difference between lowest keypoints on both objects. Default threshold is `10 cm`.
2. threshold surface-to-surface L2 distance as approximated by truncation of displacement between centers using bounding boxes as scaled basis spaces.  Default threshold is `50 cm`. This may be too large for some placements of smaller objects and should be adjusted for the context.

**NOTE:** we may find that refularized L2 distance is less accurate when objects have very different shapes. If so, we may consider projecting the displacement vectors into XZ plane for horizontal distance. This is more complex to regularize, so I opted not to do it for now.

This PR offers **direct bounding box functions** for use outside of a SImulator instance (e.g. with real-world derived 3D bounding boxes). Similarly, this PR refactors the `size_regularized_distance` function to also have a bounding box API available.

**Side thesis**: corrected some typos and refactored #1823 `objectA` variables to more pythonic `object_a` names. 

## How Has This Been Tested

Added a unit test asserting known in-set and out-set object pairs in `ReplicaCAD` dataset `apt_0` scene.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
+++ b/habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
@@ -935,7 +935,7 @@ def ontop(
 ) -> List[int]:
     """
     Get a list of all object ids or objects that are "ontop" of a particular object_a.
-    Concretely, 'ontop' is defined as: contact points between object_a and objectB have vertical normals "upward" relative to object_a.
+    Concretely, 'ontop' is defined as: contact points between object_a and object_b have vertical normals "upward" relative to object_a.
     This function uses collision points to determine which objects are resting on or contacting the surface of object_a.
 
     :param sim: The Simulator instance.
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -1128,92 +1156,189 @@ def set_link_normalized_joint_position(
...
+
+    return True
+
+
+def obj_next_to(
+    sim: habitat_sim.Simulator,
+    object_id_a: int,
+    object_id_b: int,
+    vertical_threshold=0.1,
+    l2_threshold=0.5,
```
jacobkrantz, line 1302: I might look into setting the `l2_threshold` relative to object size. These utilities support that well 👍

jacobkrantz (approved):
Thanks, Alex! The bb function abstractions will be useful for outside-sim applications.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
```diff
@@ -1128,92 +1156,189 @@ def set_link_normalized_joint_position(
...
+
+    if transform_a is None:
+        transform_a = mn.Matrix4.identity_init()
+    if transform_b is None:
+        transform_b = mn.Matrix4.identity_init()
+
+    keypoints_a = get_global_keypoints_from_bb(bb_a, transform_a)
+    keypoints_b = get_global_keypoints_from_bb(bb_b, transform_b)
+
+    lowest_height_a = min([p[1] for p in keypoints_a])
```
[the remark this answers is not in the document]
aclegg3: The bounding boxes are axis aligned in the local space of the object. Instead of re-constructing aabbs in global space, I transform all the local corners into global "keypoints" and then find the minimum. 

Same amount of compute I think. The global aabb would require iterative min/max to construct.

habitat-lab/habitat/sims/habitat_simulator/sim_utilities.py
aclegg3: Yeah, I already regularize the distance by object size, but it is true that our semantic notion of "nearby" or "next to" is vastly different for different object scales. I wonder if there is some nice "golden number" or ratio to capture this as a perceptual phenomena. :thinking:

PR #1872, opened 2024-03-20
[CI] Pin numpy version.

0mdc:
## Motivation and Context

`habitat-sim` requires `numpy` to be constrained as such: `numpy>=1.20.0,<1.24.0`.

This is a preventive change - An error currently shows up in CI logs, but it does not yet cause CI issues.

## How Has This Been Tested

Tested on CI.

## Types of changes

- **\[Bug Fix\]**

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,8 @@ jobs:
           name: setup
           command: |
               pip install black==23.1.0 --progress-bar off
-              pip install "isort[pyproject]" numpy --progress-bar off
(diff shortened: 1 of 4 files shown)


aclegg3 (approved):
Interesting that having the version pinned in requirements.txt is not enough. Is this an install order problem? I.e. we need to install numpy correctly before dependencies for sim?

aclegg3:
We updated habitat-sim to numpy 1.26.4, so this change is no longer needed.

PR #1900, opened 2024-04-07
HITL - Fix occasional crash in keyframe consolidation.

0mdc:
## Motivation and Context

This change fixes a rare issue in keyframe consolation where incomplete data would crash the networking thread.

## How Has This Been Tested

Tested locally, cannot reproduce the issue anymore.

## Types of changes

## Checklist


aclegg3 (approved):
My understanding here is that you expected these items in the `consolidated_keyframe`, but occasionally some dropped frame or edge case would result in hitting this code without the expected contents.
What is the edge case? Is it a network issue corrupting contents, a frame ordering issue, edge cases of some kind?

0mdc:
@aclegg3 : This happened to me when the incoming keyframe had `rigUpdates`, but not the consolidated keyframe. I'm not sure of the specific scenario that causes this yet.

The other guards are there to ensure that the handling logic is consistent.

PR #1905, opened 2024-04-11
HITL - Add an event system to track connections and disconnections.

0mdc:
## Motivation and Context

Distributed HITL applications require changing state when a client connects or disconnects.
This PR aims to add a single source of truth in the main thread to emit these events.

Here, the application is paused when nobody is connected.
While this is currently handled at the application level, in the future, the events may instead be subscribed in the `HitlDriver` class.

## How Has This Been Tested

Tested on distributed prototype.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/rearrange_v2.py
+++ b/examples/hitl/rearrange_v2/rearrange_v2.py
@@ -22,6 +22,7 @@ from habitat_hitl.core.gui_input import GuiInput
 from habitat_hitl.core.hitl_main import hitl_main
 from habitat_hitl.core.hydra_utils import register_hydra_plugins
 from habitat_hitl.core.text_drawer import TextOnScreenAlignment
+from habitat_hitl.core.types import ConnectionRecord, DisconnectionRecord
 from habitat_hitl.environment.camera_helper import CameraHelper
 from habitat_hitl.environment.controllers.gui_controller import (
     GuiHumanoidController,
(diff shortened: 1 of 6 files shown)


jturner65 (approved):

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -139,6 +140,21 @@ def __init__(self, app_service):
...
+            self._app_service.remote_client_state.on_client_disconnected.registerCallback(
+                self._on_client_disconnected
+            )
+            self._paused = True
+
+    def _on_client_connected(self, connection: ConnectionRecord):
+        self._paused = False
+
+    def _on_client_disconnected(self, disconnection: DisconnectionRecord):
+        self._paused = True
```
aclegg3, line 156: Does this apply to multi-user applications?
If so, does this pause when either client disconnects, or both?

0mdc: Kinda.

The application is always on. The pausing logic is useful to prevent the server from stepping, and to prevent the client to download the previously loaded scene upon connecting.

This iteration of `rearrange_v2` is only single user. Multiplayer support is not merged yet. In multplayer, application state will be controlled by a small state machine. The app will only be paused if nobody is connected.

PR #1909, opened 2024-04-11
Kinematic Relationship Manager

aclegg3:
## Motivation and Context

This PR introduces the `KinematicRelationshipManager` (KRM). The KRM tracks kinematic relationships between simulator objects as a forest of relationship trees. When parent objects are kinematically moved (e.g. open/close link or teleport object), the KRM can recursively apply the necessary relative transformation chain to the subtree with the moved object at the root in order to maintain the previously defined relationships.

**API details:**
- KRM maintains an internal `RelationshipGraph` with a forest of relationship trees. There is no guarantee that edges do not exist between trees, but the root of each tree is an object with no parent.
- KRM caches a snapshot of relative transforms between connected objects in the graph, as well as a snapshot of the global transformation of each root parent. Snapshots can be applied to the full transformation hierarchy as well as specific subtrees (for example: when moving only a single parent, there is no need to update the full state of all objects).
- KRM compares the root parent transforms to the cached state to check if a subtree needs to be updated. This means only the moving object's subtree will be updated, significantly improving performance over applying transforms to all tracked objects.

**Supporting changes:**
- adds `get_obj_transform_from_id` to `sim_utilities` to reduce boilerplate code necessary to extract transforms from both `ManagedObjects` and `ArticulatedLinks`.

**Demo application notes:** (removed before merge, see commit tree to recover)
- The demo uses the `ontop` util and physics simulation to setup initial kinematic relationships.
- When simulation is paused, the KRM evaluates relationships and takes a snapshot of the relative transforms.
- When an object is grabbed, KRM clears its parent relationships and as it moves, applies the cached transform chain to the object's subtree.
- _This demo is only meant as an example, there are edge cases when using a util like 'ontop' to automatically create relationships. The intent is for the task logic to manage the adding, removing, and applying relationships with KRM in practice._

**API examples from the demo app:**
Laying out the relevant call stack so reviewers don't need to digest the full example app.

_Initialize the KRM:_

```
self.krm = KinematicRelationshipManager(self.sim)
self.krm.initialize_from_dynamic_ontop()
#or from an episode:
#self.krm.initialize_from_obj_to_rec_pairs()
```

This registers all `ontop` relationships and takes initial transform snapshot.

_Apply the cached relationship snaphsot and then update the parent transform:_
Call this in the simulation update/step loop when a kinematic object is being moved. 

```
self.krm.apply_relations()
self.krm.update_snapshots(root_parent_subset=[self.mouse_grabber.settings.object_id_a]) 
```

_When an object is grasped, remove its parent relationships from the KRM:_
This object is now a 'root parent' in the `RelationshipGraph`.

```
if constraint_settings.link_id_a == -1: #links cannot be children, they are already constrained to their parent AO
  # remove the grabbed object from the relations manager so it can be manipulated
  self.krm.relationship_graph.remove_obj_relations(constraint_settings.object_id_a, parents_only=True)
```

_Open/close a link and apply its child transforms:_

```
# this is link, call the open API
if sutils.link_is_open(ao, ao_link, 0.02):
   sutils.close_link(ao, ao_link)
else:
   sutils.open_link(ao, ao_link)
self.krm.apply_relations() #apply relations
self.krm.update_snapshots(root_parent_subset=[hit_object]) #update root transform of the link
```

## How Has This Been Tested

Added a unit test.

Added `examples/viewer.py`, a demo application which utilizes the KRM. Videos below are captured from interactions in this application:

**Example: moving a spice rack full of spice jars:**

https://github.com/facebookresearch/habitat-lab/assets/1445143/17b022d2-3412-4bf6-90d8-480febbd3a5b

**Example of placing objects in a drawer and then updating the states as the draw is opened and closed:**

https://github.com/facebookresearch/habitat-lab/assets/1445143/7950e03e-4a44-46d6-8e62-a6a6da915b8c

**Example moving a table and shelf full objects and then removing subtrees of stacked objects:**

https://github.com/facebookresearch/habitat-lab/assets/1445143/5d9bc64b-e834-4ab3-a382-fbb0988ec6f9

**Example of placing objects in a container and then moving the container with the objects:**

https://github.com/facebookresearch/habitat-lab/assets/1445143/0e035615-8cdb-470f-8e8d-de5d8001f7dc

## Types of changes

## Checklist


examples/viewer.py
```diff
@@ -0,0 +1,1489 @@
+# Copyright (c) Meta Platforms, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# type: ignore
```
akshararai, line 5: can you add a docstring at the top of what this file does?

aclegg3: I'm not planning to merge this. Just recording it here for the record and if anyone wants to try the demo locally.

examples/viewer.py
```diff
@@ -0,0 +1,1489 @@
...
+                f"Toggle Region Draw from {self.semantic_region_debug_draw } to {not self.semantic_region_debug_draw}"
+            )
+            # Toggle visualize semantic bboxes. Currently only regions supported
+            self.semantic_region_debug_draw = (
+                not self.semantic_region_debug_draw
+            )
+
+        elif key == pressed.TAB:
+            # NOTE: (+ALT) - reconfigure without cycling scenes
+            # if not alt_pressed:
```
akshararai, line 599: this commenting is intentional? :)

examples/viewer.py
```diff
@@ -0,0 +1,1489 @@
...
+            return
+        Timer.prev_frame_duration = time.time() - Timer.prev_frame_time
+        Timer.prev_frame_time = time.time()
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+
```
akshararai, line 1405: could you add a comment on how to call this with a new dataset?

0mdc:
This is awesome Alex! Thank you!

examples/viewer.py
```diff
@@ -0,0 +1,1489 @@
...
+            self.replay_renderer = ReplayRenderer.create_batch_replay_renderer(
+                self.replay_renderer_cfg
+            )
+            # Pre-load composite files
+            if sim_settings["composite_files"] is not None:
+                for composite_file in sim_settings["composite_files"]:
+                    self.replay_renderer.preload_file(composite_file)
+
+        self.krm = KinematicRelationshipManager(self.sim)
+        self.krm.initialize_relationship_graph()
```
0mdc, line 508: If I understand correctly, we simply have to create a new instance upon changing episode. Is there anything else to do?

habitat-lab/habitat/sims/habitat_simulator/kinematic_relationship_manager.py
```diff
@@ -0,0 +1,426 @@
...
+                self.sim, obj_id, self.ao_link_map
+            )
+            assert parent_obj is not None, f"Object id {obj_id} is invalid."
+            # NOTE: because links are parented explicitly to their parent AO, we don't allow them to be children in the relationship manager.
+            # if parent_obj.object_id == obj_id:
+            #    #this is a ManagedObject, not a link
+            #    obj_within = sutils.within(self.sim, parent_obj)
+            #    for parent_id in obj_within:
+            #        self.relationship_graph.add_relation(parent_id, obj_id, "within")
+            # TODO: Using ontop here because it was easier to demo, this should change. Using more than one preposition makes cycles more common. Better to control this from the episode and task logic.
```
0mdc, line 179: 👍

habitat-lab/habitat/sims/habitat_simulator/kinematic_relationship_manager.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def _get_relations_recursive(
+        self, parent_id: int, wip_snapshot: Dict[int, Dict[int, mn.Matrix4]]
+    ) -> None:
+        """
+        Gather the relative transforms for the "work in progress" snapshot relations recursively to all children of the parent_id.
+
+        :param parent_id: The parent of the subtree on which to recurse.
+        :param wip_snapshot: The work-in-progress snapshot being constructed by this recursive process. default_dict(lambda: {})
+        """
```
0mdc, line 210: Is the snapshot still "wip"?

habitat-lab/habitat/sims/habitat_simulator/kinematic_relationship_manager.py
```diff
@@ -0,0 +1,426 @@
...
+                self._apply_relations_recursive(
+                    root_parent_id, snapshot, applied_to
+                )
+
+        return updated_root_parents
+
+    def apply_relations(self) -> List[int]:
+        """
+        Apply the previous relationship snapshot.
+        Call this in the kinematic sim loop when objects are updated.
```
0mdc, line 387: 👍

0mdc (approved):
Looks pretty good. I don't have much comments to add. I'm excited to try this out!

examples/viewer.py
akshararai: oh but this is so useful :)
I would love to have this somewhere to quickly study the episodes myself.

examples/viewer.py
aclegg3: TL;DR We should implement a specific RearrangeEpisode initializer for KRM where we add relationships between each added object and parent receptacle.

The current one is a fun tool for dynamic scenes, but we don't intend to use dynamics to settle the episodes, so this method won't work out of the box.

habitat-lab/habitat/sims/habitat_simulator/kinematic_relationship_manager.py
aclegg3: This parameter is `wip_snapshot` because it is being constructed during recursion, so the snapshot is not complete yet. Not that the feature itself is "wip".

examples/viewer.py
aclegg3: Probably don't want to use the app on a new dataset because there are no kinematic objects unless we load an episode. This example app won't really be interesting in that case.

examples/viewer.py
aclegg3: Yeah, I cannibalized an existing app and changed functionality. ALT+TAB won't work anyway since it is reserved by system OS or window switch. :laughing:

PR #1927, opened 2024-04-29
Integrate KRM with RearrangeSim

aclegg3:
## Motivation and Context

This PR integrates the `KinematicRelationshipManager` from #1909 into the `RearrangeSim` for `kinematic_mode`. 

When `kinematic_mode` is set:
- KRM is created and relationships initialized from the RearrangeEpisode
- KRM relationships are applied within the `update` loop
- `RearrangeGraspManager` severs kinematic relationships when a grasp is triggered.

Also adds some typing and docstrings style refactors.
Also updates the test rearrange episode datasets across the stack to "v2" following https://github.com/facebookresearch/habitat-sim/pull/2381

TODO:
- New relationships need to be established when a grasp is released and the object is snapped or placed onto a new surface. This is not done here, because the current logic does not handle snapping objects onto a surface in kinematic mode. Only the dynamic modality is implemented fully. 

## How Has This Been Tested

Not explicitly tested yet. Existing integration tests for RearrangeSim only.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
@@ -22,6 +22,9 @@ from habitat_sim.physics import (
 if TYPE_CHECKING:
     from omegaconf import DictConfig
 
+    from habitat.articulated_agents.manipulator import Manipulator
+    from habitat.tasks.rearrange.rearrange_sim import RearrangeSim
+
 
 class RearrangeGraspManager:
     """
@@ -29,14 +32,20 @@ class RearrangeGraspManager:
     """
 
     def __init__(
-        self, sim, config: "DictConfig", articulated_agent, ee_index=0
+        self,
+        sim: "RearrangeSim",
+        config: "DictConfig",
+        articulated_agent: "Manipulator",
+        ee_index=0,
     ) -> None:
         """Initialize a grasp manager for the simulator instance provided.
+
         :param sim: Pointer to the simulator where the agent is instantiated
         :param config: The task's "simulator" subconfig node. Defines grasping parameters.
         :param articulated_agent: The agent for which we want to manage grasping
         :param ee_index: The index of the end effector of the articulated_agent belonging to this grasp_manager
         """
+
         self._sim = sim
         self._snapped_obj_id: Optional[int] = None
         self._snapped_marker_id: Optional[str] = None
@@ -51,10 +60,12 @@ class RearrangeGraspManager:
 
     def reconfigure(self) -> None:
         """Removes any existing constraints managed by this structure."""
+
         self._snap_constraints.clear()
 
     def reset(self) -> None:
         """Reset the grasp manager by re-defining the collision group and dropping any grasped object."""
+
         # Setup the collision groups. UserGroup7 is the held object group, it
         # can interact with anything except for the robot.
         CollisionGroupHelper.set_mask_for_group(
@@ -70,6 +81,7 @@ class RearrangeGraspManager:
         Returns true if the object is too far away from the gripper, meaning
         the agent violated the hold constraint.
         """
+
         ee_pos = self._managed_articulated_agent.ee_transform(
             self.ee_index
         ).translation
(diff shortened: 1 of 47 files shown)


aclegg3:
CI was failing because the rearrange v1 dataset was mapping object placements to Receptacle.name instead of Receptacle.unique_name.
I re-processed this mapping and uploaded as v2. CI is currently testing against the updated version from https://github.com/facebookresearch/habitat-sim/pull/2381

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -323,6 +345,14 @@ def snap_to_obj(
         self._managed_articulated_agent.open_gripper()
 
         if self._kinematic_mode:
+            # update the KRM to sever any existing parent relationships for the newly grasped object
+            self._sim.kinematic_relationship_manager.relationship_graph.remove_obj_relations(
+                snap_obj_id, parents_only=True
+            )
+            # update root parent transforms so new parent state is registered
```
xavierpuigf, line 352: Why do we need this? I understand that the line above removes the relationship with the parent. Won't the root parent transforms stay the same when we pick an object? Why do we need to store them in prev_root_obj_state?

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -109,6 +112,7 @@ def __init__(self, config: "DictConfig"):
         self.viz_ids: Dict[Any, Any] = defaultdict(lambda: None)
         self._handle_to_object_id: Dict[str, int] = {}
         self._markers: Dict[str, MarkerInfo] = {}
+        self._targets: Dict[str, mn.Matrix4] = {}
```
xavierpuigf, line 115: What are those _targets? How are they relevant to KRM?

test/test_rearrange_task.py
```diff
@@ -232,16 +232,17 @@ def test_rearrange_tasks(test_cfg_path):
...
-        for _ in range(2):
-            env.reset()
-            for _ in range(MAX_PER_TASK_TEST_STEPS):
-                action = env.action_space.sample()
-                _, _, done, _ = env.step(  # type:ignore[assignment]
-                    action=action
-                )
-                if done:
-                    break
+    for _ in range(2):
```
xavierpuigf, line 235: Couldn't removing the `with  env` cause segfaults like in the llm codebase?

xavierpuigf (approved):
Looks good, thanks! Could you comment on how the v2 dataset and targets variable relate to the KRM?

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
aclegg3: When we disconnect an object from its parents, the object becomes a root parent. We need to cache the new root parent's global transform so it can be applied to update any children. 

This is the essential logic which allows, for example, picking up a tray with cups on it and moving the cups along with the tray. The tray is a new root parent and its transform must be tracked.

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
aclegg3: This was a clean-up refactor, not related to KRM. IMO, `self.` variables should all be declared in __init__(), even if they are initialized in an internal function. This allows any comments and typing to be co-located. More straightforward paradigm than trying to hunt around for the first declaration of the variable in some internal function.

aclegg3:
> Looks good, thanks! Could you comment on how the v2 dataset and targets variable relate to the KRM?

targets variable: I answered that [here](https://github.com/facebookresearch/habitat-lab/pull/1927#discussion_r1588105780). TL;DR: unrelated clean-up refactor

v2 dataset: details on the changes are [here](https://github.com/facebookresearch/habitat-sim/pull/2381). TL;DR: the v1 dataset did not comply to the required spec for `RearrangeEpisode.name_to_receptacle` which is expected by the KRM implementation to initialize the parent->child relationships at the start of the episode.

0mdc (approved):
LGTM!

PR #1929, opened 2024-05-01
HITL - Add S3 upload utility.

0mdc:
## Motivation and Context

This adds a S3 upload utility.

To use it, you must
* Have your [AWS credentials set up](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html).
    * These are automatically set for EC2 instances.
* Have `boto3` installed.
* Have a `S3_BUCKET` environment variable.

The toolkit also includes helper functions to generate descriptive unique names and S3-friendly file names.

Other considerations:
* The `util.py` file is included with the intent of adding more code to it soon.
* All of these files should probably be out of the example folder. This will be handled in a later refactoring pass.

## How Has This Been Tested

Tested on EC2 instances with downstream code.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/s3_upload.py
+++ b/examples/hitl/rearrange_v2/s3_upload.py
@@ -6,10 +6,12 @@
 
 
 import os
+import typing
 from typing import Dict, List
 
-from habitat_hitl.core.types import ConnectionRecord
 from util import timestamp
+
+from habitat_hitl.core.types import ConnectionRecord
 
 try:
     import boto3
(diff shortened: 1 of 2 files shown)


aclegg3 (approved):
Cool. Is the intent to be run from inside a server process or as a pre-process from a script? If the latter, then maybe we migrate to `scripts/` directory?

0mdc:
@aclegg3 This will be called from within the HITL app when a session finishes. For this reason, I would move this to a `habitat-hitl` util subpackage.

PR #1933, opened 2024-05-01
HITL - Add collaboration episode loader.

0mdc:
## Motivation and Context

This adds a "safe" collaboration episode loader into HITL.

## How Has This Been Tested

Tested with `habitat-llm`.

## Types of changes

- **\[Development\]**

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation if required.
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes if required.

PR #1936, opened 2024-05-01
New controller to work with habitat-llm Agents

zephirefaith:
## Motivation and Context

This PR creates `LLMController` based on `SingleAgentBaselineController` which houses basic `init`, `reset` and `act` code for calling habitat-llm agents.

Confused: Do I need to update `SingleAgentAccessMgr` class as well to be compatible with habitat-agents related to habitat-llm? The previously mentioned class just houses a lot of extra code which is useless in given setup. In habitat-llm setup the `action` is directly sent from policies running wherever habitat-llm code is running, without needing any policy setup or inference from `SingleAgentAccessMgr` class.

## Types of changes

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/config/lang_rearrange_llmspot_guihumanoid.yaml
+++ b/examples/hitl/rearrange_v2/config/lang_rearrange_llmspot_guihumanoid.yaml
@@ -22,7 +22,12 @@ habitat_baselines:
     should_load_ckpt: False
   rl:
     agent:
-      num_pool_agents_per_type: [1, 1]
+      type: "SingleAgentAccessMgr"
+      num_agent_types: 1
+      num_pool_agents_per_type: [1]
+      num_active_agents_per_type: [1]
+      agent_sample_interval: 20
+      force_partner_sample_idx: -1
     policy:
 
 habitat_hitl:
@@ -31,12 +36,10 @@ habitat_hitl:
     width: 1300
     height: 1000
   gui_controlled_agents:
-    - agent_index: 0
-      lin_speed: 10.0
-      ang_speed: 300
     - agent_index: 1
       lin_speed: 10.0
       ang_speed: 300
   hide_humanoid_in_gui: True
   camera:
     first_person_mode: True
+
--- a/examples/hitl/rearrange_v2/config/lang_rearrange_llmspot_guihumanoid.yaml
+++ b/examples/hitl/rearrange_v2/config/lang_rearrange_llmspot_guihumanoid.yaml
@@ -39,6 +39,10 @@ habitat_hitl:
     - agent_index: 1
       lin_speed: 10.0
       ang_speed: 300
+  llm_controller_agents:
+    - agent_index: 0
+      lin_speed: 10.0
+      ang_speed: 300
   hide_humanoid_in_gui: True
   camera:
     first_person_mode: True
--- a/examples/hitl/rearrange_v2/rearrange_v2.py
+++ b/examples/hitl/rearrange_v2/rearrange_v2.py
@@ -32,6 +32,11 @@ from habitat_hitl.environment.controllers.gui_controller import (
 from habitat_hitl.environment.hablab_utils import get_agent_art_obj_transform
 from habitat_sim.utils.common import quat_from_magnum, quat_to_coeffs
 
+try:
+    from habitat_llm.agent.env import dataset  # noqa: F401
+except ImportError:
+    print("Habitat-LLM not installed. Skipping import.")
+
 
 class DataLogger:
     def __init__(self, app_service):
--- a/habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
+++ b/habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
@@ -1,0 +1,96 @@
+from habitat.core.environments import GymHabitatEnv
+from habitat_hitl.environment.controllers.baselines_controller import (
+    SingleAgentBaselinesController,
+)
+from omegaconf import DictConfig
+from habitat_llm.utils import setup_config, fix_config
+from hydra.utils import instantiate
+from habitat_llm.agent.env import EnvironmentInterface
+from habitat_llm.agent import Agent
+
+
+class LLMController(SingleAgentBaselinesController):
+    """Controller for single LLM controlled agent."""
+
+    def __init__(
+        self,
+        agent_idx: int,
+        is_multi_agent: bool,
+        config: DictConfig,
+        gym_habitat_env: GymHabitatEnv,
+    ):
+        super().__init__(agent_idx, is_multi_agent, config, gym_habitat_env)
+        fix_config(config)
+        seed = 47668090
+        self.config = setup_config(config, seed)
+        self.planner = None
+        self.environment_interface = None
+
+        # NOTE: this is creating just one agent. Habitat-LLM has code for creating
+        # multiple processes/agents in one go. I am only prototyping single process, as
+        # I assume the onus of creating multiple processes is on the user/hitl_driver,
+        # and this code will be called once per Sim instantiation
+        self.initialize_environment_interface()
+        self.initialize_planner()
+        self.info = {}
+
+    def initialize_planner(self):
+        # NOTE: using instantiate here, but given this is planning for a single agent
+        # always will this ever be an option of Centralized vs Decentralized? Maybe DAG...?
+        self.planner = instantiate(self.config.evaluation.planner)
+        self.planner.agents = self.initialize_agents(
+            self.config.evaluation.agents
+        )
+
+    def initialize_agents(self, agent_configs):
+        agents = []
+        for _, agent_conf in agent_configs.items():
+            # Instantiate the agent
+            agent = Agent(
+                agent_conf.uid, agent_conf.config, self.environment_interface
+            )
+
+            # Make sure that its unique by adding to the set
+            agents.append(agent)
+        return agents
+
+    def initialize_environment_interface(self):
+        self.environment_interface = EnvironmentInterface(
+            self.config, gym_habitat_env=self._gym_habitat_env
+        )
+
+        # NOTE: this is to replicate initial call of  get_next_action, in
+        # run_instruction() method. I am not sure why we do this initially?
+        _low_level_actions, _planner_info, _task_done = (
+            self.planner.get_next_action(
+                self.current_instruction,
+                {},
+                self.env_interface.world_graph,
+            )
+        )
+
+    def on_environment_reset(self):
+        # NOTE: the following ONLY resets self._test_recurrent_hidden_states,
+        # self._prev_actions and self._not_done_masks
+        super().on_environment_reset()
+        self.planner.reset()
+        self.environment_interface.reset()
+
+        self.current_instruction = (
+            self.environment_interface.hab_env.current_episode.instruction
+        )
+
+    def act(self, observations):
+        # NOTE: update the world state to reflect the new observations
+        self.environment_interface.update_world_state(observations)
+
+        # NOTE: this is where the LLM magic happens, the agent is given the observations
+        # and it returns the actions for the agent
+        low_level_actions, planner_info, task_done = (
+            self.planner.get_next_action(
+                self.current_instruction,
+                observations,
+                self.envirnment_interface.world_graph,
+            )
+        )
+        return low_level_actions
--- a/examples/hitl/rearrange_v2/config/lang_rearrange_spot_humanoid.yaml
+++ b/examples/hitl/rearrange_v2/config/lang_rearrange_spot_humanoid.yaml
@@ -1,0 +1,39 @@
+# @package _global_
+
+defaults:
+  - language_rearrange_multi_agent
+  - hitl_defaults
+  - _self_
+
+habitat:
+  # various config args to ensure the episode never ends
+  environment:
+    max_episode_steps: 0
+    iterator_options:
+      # For the demo, we want to showcase the episodes in the specified order
+      shuffle: False
+
+habitat_baselines:
+  # todo: document these choices
+  eval:
+    should_load_ckpt: False
+  rl:
+    agent:
+      num_pool_agents_per_type: [1, 1]
+    policy:
+
+habitat_hitl:
+  window:
+    title: "Rearrange"
+    width: 1300
+    height: 1000
+  gui_controlled_agents:
+    - agent_index: 0
+      lin_speed: 10.0
+      ang_speed: 300
+    - agent_index: 1
+      lin_speed: 10.0
+      ang_speed: 300
+  hide_humanoid_in_gui: True
+  camera:
+    first_person_mode: True
--- a/examples/hitl/rearrange_v2/config/language_rearrange.yaml
+++ b/examples/hitl/rearrange_v2/config/language_rearrange.yaml
@@ -5,13 +5,13 @@ defaults:
   - /habitat: habitat_config_base
   - /habitat/task: task_config_base
 
-  - /habitat/simulator/sensor_setups@habitat.simulator.agents.main_agent: rgbd_head_agent
-  - /habitat/simulator/agents@habitat.simulator.agents.main_agent: human
+  - /habitat/simulator/sensor_setups@habitat.simulator.agents.agent_0: rgbd_head_agent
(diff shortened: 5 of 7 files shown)


examples/hitl/rearrange_v2/config/lang_rearrange_spot_humanoid_visual.yaml
zephirefaith, line 1: Ignore this. This needs to be deleted.

examples/hitl/rearrange_v2/config/lang_rearrange_llmspot_guihumanoid.yaml
```diff
@@ -0,0 +1,90 @@
...
+    #   gui_controlled: True
+world_model:
+  partial_obs: True
+habitat:
+  # various config args to ensure the episode never ends
+  environment:
+    max_episode_steps: 0
+    iterator_options:
+      # For the demo, we want to showcase the episodes in the specified order
+      shuffle: False
```
0mdc, line 35: We'll want to pull [those changes](https://github.com/facebookresearch/habitat-lab/pull/1935) here.

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
```diff
@@ -0,0 +1,153 @@
...
+  - /habitat_conf/habitat_agent@habitat.simulator.agents.agent_1: rgbd_head_rgbd_arm_agent_vis
+  - /habitat/simulator/agents@habitat.simulator.agents.agent_1: human
+  # following are from hab-llm entry config. Questions:
+  # 1. What is the difference between
+  #    habitat.task.actions.agent_1_humanoid_base_velocity and
+  #    habitat.task.actions.agent_1_base_velocity?
+  # 2. What is the humanoidjoint_action? Do we need it here?
+  # - /habitat/task/actions@habitat.task.actions.agent_1_humanoid_base_velocity: humanoid_base_velocity
+  # - /habitat/task/actions@habitat.task.actions.agent_1_humanoidjoint_action: humanoidjoint_action
+  # - /habitat/task/actions@habitat.task.actions.agent_1_base_velocity: base_velocity
```
0mdc, line 18: @xavierpuigf Would you know the answers to these?

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
```diff
@@ -0,0 +1,153 @@
...
+      # - agent_1_relative_resting_position
+      # - agent_1_joint
+      # - agent_1_ee_pos
+      - agent_1_is_holding
+      # - agent_1_dynamic_obj_goal_sensor
+      # - agent_1_dynamic_goal_to_agent_gps_compass
+      # - agent_1_dynamic_obj_start_sensor
+# --- habitat-llm block
+  environment:
+    max_episode_steps: 750  # this is 20000 in habitat-llm
```
0mdc, line 112: We'll want to double-check this.
1. Is this enough to complete the longest tasks?
2. What's the typical step count for a normal multiplayer run? (I'll get the data)
3. What happens when this is done? We would need to call the appropriate code in HITL to trigger the next application state.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
```diff
@@ -0,0 +1,167 @@
...
+            datefmt="%H:%M:%S",
+            level=logging.DEBUG,
+            force=True,
+        )
+        self._analysis_logger = logging.getLogger("LLMController")
+        self._analysis_logger.debug("LLMController initialized")
+
+        with habitat.config.read_write(self._config):
+            fix_config(self._config)
+        seed = 47668090
```
0mdc, line 56: Where does this seed come from?

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
```diff
@@ -0,0 +1,153 @@
...
+  simulator:
+    type: RearrangeSim-v0
+    seed: 100
```
0mdc, line 115: We could perhaps set the desired seed here.

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
```diff
@@ -0,0 +1,153 @@
...
+      - "data/objects_ovmm/train_val/hssd/configs/objects/"
+    concur_render: True
+    auto_sleep: True
+    kinematic_mode: True
+    ac_freq_ratio: 1
+    step_physics: False
+    habitat_sim_v0:
+      allow_sliding: True
+      enable_physics: True
+    # Q. What is this agents_order?
```
0mdc, line 145: This may be important to figure out.

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
```diff
@@ -0,0 +1,153 @@
...
+        joint_start_noise: 0.0
+    # --- habitat-llm block
+
+    additional_object_paths:
+      - "data/objects/ycb/configs/"
+      - "data/objects_ovmm/train_val/ai2thorhab/configs/objects/"
+      - "data/objects_ovmm/train_val/amazon_berkeley/configs/"
+      - "data/objects_ovmm/train_val/google_scanned/configs/"
+      - "data/objects_ovmm/train_val/hssd/configs/objects/"
+    concur_render: True
```
0mdc, line 137: Let me check whether we want this to be true in our context.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -44,6 +44,26 @@
...
+# -------- HABITAT-LLM SPECIFIC CODE --------
+
+
+def register_habitat_llm_extensions(config):
+    register_actions(config)
+    register_measures(config)
+    register_sensors(config)
+
+
+# -------- HABITAT-LLM SPECIFIC CODE --------
```
0mdc, line 66: We could probably move this to [this file](https://github.com/facebookresearch/habitat-lab/pull/1933), and change it from an "episode loader" to a "`habitat-llm` loader".

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
```diff
@@ -10,12 +10,14 @@
...
 import abc
 import json
+import logging
 from datetime import datetime
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
 
 import magnum as mn
 import numpy as np
+from habitat_llm.utils.analysis import CodeTimer
```
0mdc, line 20: We'll want to wrap this in a try/except, otherwise HITL examples will fail for all users. A utility class would be good to wrap this.

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
```diff
@@ -95,6 +97,16 @@ def __init__(
         self._play_episodes_filter_str = self._hitl_config.episodes_filter
         self._num_recorded_episodes = 0
         self._gui_input = gui_input
+        logging.basicConfig(
+            filename="/home/priyamp/hitl/app_timing.log",
```
0mdc, line 101: We'll want to change this default path :)

We probably also want to make this logging optional.

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
```diff
@@ -372,8 +385,8 @@ def _reset_environment(self):
             self._remote_client_state.clear_history()
 
         # todo: fix duplicate calls to self.ctrl_helper.on_environment_reset() here
-        if self.ctrl_helper:
-            self.ctrl_helper.on_environment_reset()
+        # if self.ctrl_helper:
+        #     self.ctrl_helper.on_environment_reset()
```
0mdc, line 389: We'll probably want to uncomment this.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
0mdc, line 1: The file header is missing here.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
```diff
@@ -0,0 +1,167 @@
+import logging
+import threading
+from typing import Any, Dict, Union
+
+import numpy as np
+from habitat_llm.agent import Agent
+from habitat_llm.agent.env import EnvironmentInterface
+from habitat_llm.planner.llm_planner import LLMPlanner
+from habitat_llm.utils import fix_config, setup_config
+from habitat_llm.utils.analysis import CodeTimer
```
0mdc, line 10: I suppose that this code will never be called if `habitat-llm` is not used, correct?

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
```diff
@@ -0,0 +1,167 @@
...
+        self._habitat_env = gym_habitat_env.unwrapped.habitat_env
+        self._agent_idx = agent_idx
+        # TODO: gather this from config
+        self._agent_action_length = 28
+        self._thread: Union[None, threading.Thread] = None
+        self._low_level_actions: Union[None, dict, np.ndarray] = {}
+        self._iter = 0
+        self._skip_iters = 0
+        logging.basicConfig(
+            filename="/home/priyamp/hitl/act_timing.log",
```
0mdc, line 44: Same comment about logging here :)

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
```diff
@@ -0,0 +1,167 @@
...
+    def initialize_agents(self, agent_configs):
+        agents = []
+        for _, agent_conf in agent_configs.items():
+            # Instantiate the agent
+            agent = Agent(
+                agent_conf.uid, agent_conf.config, self.environment_interface
+            )
+
+            # Make sure that its unique by adding to the set
+            agents.append(agent)
```
0mdc, line 90: > Make sure that its unique by adding to the set

The container is not a set. Typo?

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
```diff
@@ -0,0 +1,167 @@
...
+            planner_info,
+            task_done,
+        ) = self.planner.get_next_action(
+            self.current_instruction,
+            observations,
+            self.environment_interface.world_graph,
+            verbose=True,
+        )
+        if task_done:
+            print("Task Done")
```
0mdc, line 137: I suppose that this would be the point where we hook up to HITL so that the agent can signal that its part of the task is done?

Would this also be called after the timeout? (`max_episode_steps`)

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -203,28 +206,28 @@ def calculate_walk_pose(
...
+
+        # offset = 0
+        # if self.walk_mocap_frame < prev_mocap_frame:
+        #     # We looped over the motion
+        #     offset = self.walk_motion.displacement[-1]
+
+        # distance_covered = max(
+        #     0, new_cum_distance_covered + offset - prev_cum_distance_covered
+        # )
+        # dist_diff = min(distance_to_walk, distance_covered)
```
0mdc, line 230: We need those changes to set humanoid walk speed. However, this may break other stuff.

Perhaps we could make animation-driven humanoid speed optional via config? If we do that, I'm afraid that it will be hard for other developers to figure out why the speed is limited.

Ideally, animation speed should be driven by locomotion speed.

@xavierpuigf Thoughts?

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -252,7 +255,9 @@ def calculate_walk_pose(
         # The base_transform here is independent of transforms caused by the current
         # motion pose.
         obj_transform_base = look_at_path_T
-        forward_V_dist = forward_V * dist_diff * distance_multiplier
+        # HACK FROM MIKAEL
+        # forward_V_dist = forward_V * dist_diff * distance_multiplier
```
0mdc, line 259: Sorry for this abomination :)

0mdc (commented):
Thank you @zephirefaith ! I did a first pass and raised some concerns.

Let me know if I can help!

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
0mdc: Indeed, example tests are failing. Should be as simple as checking for the dependency.

0mdc:
@zephirefaith You'll probably want to have this double-checked by someone who is familiar with the LLM configs so that we don't misstep here.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
zephirefaith: correct! This controller assumes you are using habitat-llm-agent underneath. I can add comment header to explicitly call this out.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
zephirefaith: Good catch. This is cargo-cult artifact.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
zephirefaith: The two conditions this would be called on:
1. If # of planning steps > max planning steps threshold
2. LLM reasons that the task is solved/can not be solved further

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
zephirefaith: For now, enabling/disabling this through config and adding comments both at config and code-level sounds like the least effort option to still maintain "other" uses of code.

Wdyt? Also, I do not completely understand this code layer so your help to add comments here would be perfect.

habitat-hitl/habitat_hitl/_internal/hitl_driver.py
zephirefaith: I don't think we should uncomment this. There are two resets being called here, which was crashing episode initialization.

zephirefaith: Actually no, I don't think this is the 2x reset call. I think the other reset is happening on habitat-llm side that I need to disable.

habitat-hitl/habitat_hitl/environment/controllers/llm_controller.py
zephirefaith: It's the default in all our code. We can move this to config. But same as what other experiments are using.

examples/hitl/rearrange_v2/config/language_rearrange_multi_agent_llm_gui.yaml
xavierpuigf: 1. you don't need agent_1_base_velocity for the human, only agent_1_humanoid_base_velocity. The difference is that humanoid base velocity generates a human realistic motion whereas base velocity only moves the base (used to move the robot for instance).
2. humanoidjoint_action sets the human joints so that the pose is realistic. You will never be callign it yourself, the agent_1_base_velocity sets that. But you sitll need to define it here

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
xavierpuigf: Yep! The fast solution is to disable animation. I have a way to remove the speed limitation but havent had time to get into it yet. If we want to have animation, I can work on integrating it next week.

xavierpuigf: We should not merge this as is though, it will break the functionalities for any other kind of motion

PR #1942, opened 2024-05-06
allow STATIC AOs in kinematic_mode

aclegg3:
## Motivation and Context

Using `STATIC` furniture is a common optimization for kinematic simulation. The RearrangeSim should support this. By setting all AOs to `KINEMATIC`, the AO furniture is not being added to the navmesh causing issues.

## How Has This Been Tested

CI

## Types of changes

## Checklist


zephirefaith (approved):
Ah, I see. So previously every articulated object was set as Kinematic and was not added to NavMesh. Now every _dynamic_, i.e. _movable_, articulated object will still be excluded from NavMesh but _static_, i.e. _unmovable_, will be included in the initial computation.

Is that right?

PR #1947, opened 2024-05-10
add ignore ids to unoccluded snap API

aclegg3:
## Motivation and Context

The accessibility and unoccluded snap functions rely on raycasting to determine if a target object is the first hit (and thus unoccluded) from a particular navmesh point. 

This PR adds additional ignore object ids which won't trigger the success condition, but also won't result in false-positive occlusions for complex cases. For example, if the robot is already standing at a position, checking accessibility of a furniture object, the robot's own links should be ignored in the check.

## How Has This Been Tested

TODO: this API is only integration tested and needs unit tests

## Types of changes

## Checklist


rutadesai (approved):
Looks good, thanks Alex.

PR #1950, opened 2024-05-14
HITL - Make agents transparent to selection.

0mdc:
## Motivation and Context

In the HITL app, the agents controlled in first-person are hidden in the user's view. The agent may still obstruct the selection UI, e.g. the spot arm is in the way of selection. This causes selection to appear flaky.

This PR changes the following:
* Objects rejected by selection discriminator functions are "transparent" to selection, meaning that the next closest hit info will be queried.
* Agents can now be selected through.

## How Has This Been Tested

* Tested in single/multi user HITL.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/ui.py
+++ b/examples/hitl/rearrange_v2/ui.py
@@ -80,7 +80,10 @@ class UI:
         self._selections: List[Selection] = []
         # Track hovered object.
         self._hover_selection = Selection(
-            self._sim, self._gui_input, Selection.hover_fn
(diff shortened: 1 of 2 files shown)


zephirefaith (approved):
Logically makes sense. A bit concerned the discriminator can be used to create very illogical scenarios so would suggest adding caution (e.g. if we make receptacles/furniture transparent can we select an object inside it?).

Thanks for this!

PR #1958, opened 2024-05-15
HITL - Add picture-in-picture viewports.

0mdc:
## Motivation and Context

This changeset adds picture-in-picture viewports to the HITL app.

`rearrange_v2` implements this by showing the other user's view in a multiplayer session.

https://github.com/facebookresearch/habitat-lab/assets/110583667/01688fb5-0f55-42c6-b79e-b656b0560c2c

## Notes

Depends on:
* https://github.com/0mdc/siro_hitl_unity_client/pull/28

## How Has This Been Tested

Tested in multiplayer HITL application.

## Types of changes

- **\[Development\]**

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation if required.
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes if required.

PR #1964, opened 2024-05-17
HITL - Rearrange State Machine

0mdc:
## Motivation and Context

This changeset turns `rearrange_v2` into a state machine with the following states:
1. Reset (initial state)
   * Kicks all users. Intended to reset states between sessions.
   * When all users are disconnected, changes state to `lobby`.
2. Lobby
   * Blank screen. New incoming connections are only allowed in this state.
   * When `max_client_count` users are connected, changes state to `rearrange_v2`.
3. Rearrange
   * `rearrange_v2` application.
   * When both users signaled that the episode is done, changes state to `reset`.

https://github.com/facebookresearch/habitat-lab/assets/110583667/b76e4b5d-5c09-4268-969f-c8e1747ceb0a

## Notes

* The entry point is now `main.py` instead of `rearrange_v2.py`.
* `rearrange_v2.py` will be renamed later to simplify merging of concurrent changes.

**Depends on:**
* https://github.com/facebookresearch/habitat-lab/pull/1958
* https://github.com/facebookresearch/habitat-lab/pull/1959

## Limitations

* In this version, there is no session management. The same episode keeps running indefinitely - there is no reset.
* When users are kicked, no message is shown.
* Because `ClientHelper` depends on connection events, its lifetime is now coupled with `RemoteClientState`. This will be refactored in a later pass.
* `rearrange_v2` smoke test is disabled until it is stabilized.

## How Has This Been Tested

Tested on multiplayer HITL application.

## Types of changes

- **\[Development\]**

## Checklist


examples/hitl/rearrange_v2/app_state_lobby.py
```diff
@@ -0,0 +1,67 @@
...
+
+from typing import Final, Optional
+
+from app_data import AppData
+from app_state_base import AppStateBase
+from app_states import create_app_state_rearrange
+
+from habitat_hitl.app_states.app_service import AppService
+
+START_SESSION_DELAY: Final[float] = 0.5
```
aclegg3, line 15: Comment to explain the empirically chosen value? I see the comment below about disconnects, but still curious.

aclegg3 (approved):
Looks good from my context.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -306,31 +284,62 @@ def __init__(self, app_service: AppService):
...
+            return None
+
+    def on_enter(self):
+        super().on_enter()
+
+        user_index_to_agent_index_map: Dict[int, int] = {}
+        for user_index in range(len(self._user_data)):
+            user_index_to_agent_index_map[user_index] = self._user_data[
+                user_index
+            ].gui_agent_controller._agent_idx
```
zephirefaith, line 308: Confirming that user-index is networked humans who have joined and != agent-count/index?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -306,31 +284,62 @@ def __init__(self, app_service: AppService):
...
+        super().on_exit()
+
+    def _is_episode_finished(self) -> bool:
+        """
+        Determines whether all users have finished their tasks.
+        """
+        return all(
+            self._user_data[user_index].episode_finished
+            for user_index in self._users.indices(Mask.ALL)
+        )
```
zephirefaith, line 320: Interested to see how can I expose user-data for llm-agent to you?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -306,31 +284,62 @@ def __init__(self, app_service: AppService):
...
+        # TODO
+        self._app_service.remote_client_state._client_helper.activate_users()
+
         # Set the task instruction
         current_episode = self._app_service.env.current_episode
+        if hasattr(current_episode, "instruction"):
+            task_instruction = current_episode.instruction
+            # TODO: Users will have different instructions.
+            for user_index in self._users.indices(Mask.ALL):
+                self._user_data[user_index].task_instruction = task_instruction
```
zephirefaith, line 336: Hmmm, I see. I am setting such metadata/data for llm-agent directly from habitat `Env` class. It'd be good to standardize all data-setting to flow from here.


Therefore, interested to see how can I consume this `user-data` variable at init/reset time. We can just pass this on to `ctrl_helper`, right? It'll just be ignore as part of `*args, **kwargs` for other controllers which have no use for it?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -349,31 +358,29 @@ def _update_grasping_and_set_act_hints(self, user_index: int):
...
-        controls_str += "Place object: Right click (hold)\n"
+        if self._user_data[user_index].show_gui_text:
+            controls_str += "H: Toggle help\n"
+            controls_str += "Look: Middle click (drag), I, K\n"
+            controls_str += "Walk: W, S\n"
+            controls_str += "Turn: A, D\n"
+            controls_str += "Finish episode: Zero (0)\n"
+            controls_str += "Open/close: Double-click\n"
+            controls_str += "Pick object: Double-click\n"
+            controls_str += "Place object: Right click (hold)\n"
```
zephirefaith, line 374: Is there an option to see current-instruction?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -349,31 +358,29 @@ def _update_grasping_and_set_act_hints(self, user_index: int):
...
+
+        client_helper = self._app_service.remote_client_state._client_helper
+        idle_time = client_helper.get_idle_time(user_index)
+        if idle_time > 10:
+            controls_str += f"Idle time {idle_time}s\n"
```
zephirefaith, line 379: Note if you kick users/agents for `idle_time > SOME_THRESH` this may need to be tweaked for LLM-agent versus human user.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -349,31 +358,29 @@ def _update_grasping_and_set_act_hints(self, user_index: int):
...
+
         return controls_str
 
     def _get_status_text(self, user_index: int):
```
zephirefaith, line 383: Ooh! Lovely. I can pass `current_high_level_action` being executed by LLM-agent here.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -459,37 +466,46 @@ def sim_update(self, dt: float, post_sim_update_dict):
...
-def main(config):
-    hitl_main(
-        config,
-        lambda app_service: AppStateRearrangeV2(app_service),
-    )
+    def _is_any_user_active(self) -> bool:
+        return any(
+            self._user_data[user_index].gui_input.get_any_input()
+            for user_index in range(self._app_data.max_user_count)
+        )
```
zephirefaith, line 497: How does this work for LLM-agent? Or do you not care? We probably should.

0mdc: This should work without changes with LLM agents. `user_data`, `max_user_count` and `user_index` all relate to human users.

In the LLM case, `max_user_count` will be `1`, but `agent_count` will be `2`. I have to decouple these concepts within `user_data` (will most likely create an analogous `agent_data` struct.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: If time permits, I can also add a `text` field to viewports so that its rendered into the picture-in-picture view.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: The LLM agent will not have an idle counter. I still have to decouple this.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: Yes, it's currently displayed via the status message.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: That should be doable.

If you only need user events, you should be able to bind events from `rearrange_v2.on_environment_reset()` as you will have access to both variables. I'll be adding them in the next 2 PRs.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: Probably not ideal, but the easiest solution would be to query the agent controller types at initialization and handle things from there.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: Correct! This is the "decoupling" work I still have to do.

PR #1967, opened 2024-05-20
HITL - Data collection

0mdc:
## Motivation and Context

This changeset enables the `rearrange_v2` app to record data for HITL experiments.

## How it works

This adds a session recorder, which has its lifetime tied to a single HITL session.

When reaching the `End Session` state ([see this PR](https://github.com/facebookresearch/habitat-lab/pull/1965)), the session is uploaded to S3 before being destroy.

**Configuration:**

The following configuration will do the following:
* Create output file `output/session.json.gz`
* Upload it to S3 at `Placeholder/[completed/incomplete]/session.json.gz`.

The bucket is defined via the environment variable `S3_BUCKET`. 
```
rearrange_v2:
  data_collection:
    s3_path: "Placeholder/"
    output_file_name: "session"
```

## Notes

* S3 is currently the only supported data provider.
* Files are available for inspection locally. They are deleted before uploading the next session.

Depends on:
* https://github.com/facebookresearch/habitat-lab/pull/1965

## How Has This Been Tested

Tested on EC2 instances running single and multi user applications.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/app_state_end_session.py
+++ b/examples/hitl/rearrange_v2/app_state_end_session.py
@@ -4,15 +4,23 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
+import shutil
 from typing import Optional
 
 from app_data import AppData
 from app_state_base import AppStateBase
 from app_states import create_app_state_reset
+from s3_upload import (
+    generate_unique_session_id,
+    make_s3_filename,
+    upload_file_to_s3,
+)
 from session import Session
 from util import get_top_down_view
 
 from habitat_hitl.app_states.app_service import AppService
+from habitat_hitl.core.serialize_utils import save_as_json_gzip
 from habitat_hitl.core.user_mask import Mask
 
 # Duration of the end session message, before users are kicked.
@@ -55,5 +63,58 @@ class AppStateEndSession(AppStateBase):
         self._elapsed_time += dt
 
     def _end_session(self):
-        # TODO: Data collection.
-        pass
+        session = self._session
+        if session is None:
+            print("Null session. Skipping S3 upload.")
+            return
+
+        # Finalize session.
+        if self._session.error == "":
+            session.success = True
+        session.session_recorder.end_session(self._session.error)
+
+        # Get data collection parameters.
+        try:
+            config = self._app_service.config
+            data_collection_config = config.rearrange_v2.data_collection
+            s3_path = data_collection_config.s3_path
+            if s3_path[-1] != "/":
+                s3_path += "/"
+            s3_subdir = "complete" if session.success else "incomplete"
+            s3_path += s3_subdir
+
+            # Use the port as a discriminator for when there are multiple concurrent servers.
+            output_folder_suffix = str(config.habitat_hitl.networking.port)
+            output_folder = f"output_{output_folder_suffix}"
+
+            output_file_name = data_collection_config.output_file_name
+            output_file = f"{output_file_name}.json.gz"
+
+        except Exception as e:
+            print(f"Invalid data collection config. Skipping S3 upload. {e}")
+            return
+
+        # Delete previous output directory
+        if os.path.exists(output_folder):
+            shutil.rmtree(output_folder)
+
+        # Create new output directory
+        os.makedirs(output_folder)
+        json_path = os.path.join(output_folder, output_file)
+        save_as_json_gzip(session.session_recorder, json_path)
+
+        # Generate unique session ID
+        session_id = generate_unique_session_id(
+            session.episode_ids, session.connection_records
+        )
+
+        # Upload output directory
+        orig_file_names = [
+            f
+            for f in os.listdir(output_folder)
+            if os.path.isfile(os.path.join(output_folder, f))
+        ]
+        for orig_file_name in orig_file_names:
+            local_file_path = os.path.join(output_folder, orig_file_name)
+            s3_file_name = make_s3_filename(session_id, orig_file_name)
+            upload_file_to_s3(local_file_path, s3_file_name, s3_path)
--- a/examples/hitl/rearrange_v2/app_state_end_session.py
+++ b/examples/hitl/rearrange_v2/app_state_end_session.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Meta Platforms, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+from app_data import AppData
+from app_state_base import AppStateBase
+from app_states import create_app_state_reset
+from session import Session
+from util import get_top_down_view
+
+from habitat_hitl.app_states.app_service import AppService
+from habitat_hitl.core.user_mask import Mask
+
+# Duration of the end session message, before users are kicked.
+SESSION_END_DELAY = 5.0
+
+
+class AppStateEndSession(AppStateBase):
+    """
+    * Indicate users that the session is terminated.
+    * Upload collected data.
+    """
+
+    def __init__(
+        self, app_service: AppService, app_data: AppData, session: Session
+    ):
+        super().__init__(app_service, app_data)
+        self._session = session
+        self._elapsed_time = 0.0
+        self._save_keyframes = False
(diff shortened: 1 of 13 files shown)


examples/hitl/rearrange_v2/util.py
zephirefaith, line 1: seems like a splitting artifact from #1965,right?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -83,15 +91,29 @@ def get_objects_state(self):
...
-            "task_completed", task_completed
-        )
+    def record_state(
+        self, elapsed_time: float, user_data: List[UserData]
+    ) -> Dict[str, Any]:
+        data: Dict[str, Any] = {
+            "t": elapsed_time,
+            "users": [],
+            "object_states": self.get_objects_state(),
+            "agent_states": self.get_agents_state(),
```
zephirefaith, line 101: I can add a "world-graph" API here if you want to save those object-to-furniture/agent/receptacle relations here.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -83,15 +91,29 @@ def get_objects_state(self):
...
+        }
+
+        for user_index in range(len(user_data)):
+            u = user_data[user_index]
+            user_data_dict = {
+                "task_completed": u.episode_finished,
+                "task_succeeded": u.episode_success,
+                "camera_transform": u.cam_transform,
+                "held_object": u.ui._held_object_id,
```
zephirefaith, line 110: We've discussed this already but this ID if passed to grasp_mgr, in a way understood by sim, would solve what habitat-llm needs.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -145,6 +170,12 @@ def __init__(
             camera_helper=self.camera_helper,
         )
 
+        # Register UI callbacks
+        self.ui.on_pick.registerCallback(self._on_pick)
+        self.ui.on_place.registerCallback(self._on_place)
+        self.ui.on_open.registerCallback(self._on_open)
+        self.ui.on_close.registerCallback(self._on_close)
+
```
zephirefaith, line 178: <3

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -251,16 +287,56 @@ def _get_camera_lookat_pos(self) -> mn.Vector3:
...
+    def _on_close(self, e: UI.CloseEventData):
+        self.ui_events.append(
+            {
+                "type": "close",
+                "obj_handle": e.object_handle,
+                "obj_id": e.object_id,
+            }
+        )
+
```
zephirefaith, line 326: This is irrelevant to PR, but do we need open/close for single-learn data collection?

examples/hitl/rearrange_v2/rearrange_v2.py
zephirefaith: To use these, would I need to pass on `ui` object to `LLMController`?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -307,9 +395,24 @@ def on_enter(self):
...
+            episode.episode_id,
+            episode.scene_id,
+            episode.scene_dataset_config,
+            user_index_to_agent_index_map,
+        )
+
     def on_exit(self):
         super().on_exit()
 
+        episode_success = all(
```
zephirefaith, line 409: `episode_success` here means "user thinks episode was done/success", right?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -486,9 +589,12 @@ def sim_update(self, dt: float, post_sim_update_dict):
 
         #  Collect data.
         self._elapsed_time += dt
+        # TODO: Always record with non-human agent.
```
zephirefaith, line 592: What does this mean?

examples/hitl/rearrange_v2/session_recorder.py
```diff
@@ -0,0 +1,85 @@
...
+    def record_frame(
+        self,
+        frame_data: Dict[str, Any],
+    ):
+        self.data["end_timestamp"] = timestamp()
+        self.data["frame_count"] += 1
+
+        self.data["episodes"][-1]["end_timestamp"] = timestamp()
+        self.data["episodes"][-1]["frame_count"] += 1
+        self.data["episodes"][-1]["frames"].append(frame_data)
```
zephirefaith, line 85: Is this frame different from the frame you already have the `FrameRecorder` for? Why twice?

zephirefaith (approved):
Clarification questions mostly.

I skipped the files which largely looked the same as #1965. I couldn't tell the diff, LMK if you want me to look again at the diff, if any, once the previous one is merged.

examples/hitl/rearrange_v2/util.py
0mdc: The PR was based on #1965, which is now merged. The diff will be easier to follow.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: That would be perfect.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: You can probably just do it from `rearrange_v2` initialization code, which has access to both the UI and `LLMController`.

Something like this would work:
```
self._user_data[0].ui.on_pick.registerCallback(llm_controller.on_pick())
```

If we ever want to scale this to N users, we would just pass the user index in the event data. For now, this does the job.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: We don't need to do it right now. That will most likely change with the inclusion of object states.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: In the current state, success is the only outcome.

In a following PR, I'll be adding a way for users to report either success or failure via a form.

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: In single-user and multi-user modes, we can skip recording of frames when there's no user input.

With the LLM agent, we'll most likely have to record every frame.

examples/hitl/rearrange_v2/app_state_end_session.py
```diff
@@ -55,5 +63,58 @@ def sim_update(self, dt: float, post_sim_update_dict):
...
+
+        # Get data collection parameters.
+        try:
+            config = self._app_service.config
+            data_collection_config = config.rearrange_v2.data_collection
+            s3_path = data_collection_config.s3_path
+            if s3_path[-1] != "/":
+                s3_path += "/"
+            s3_subdir = "complete" if session.success else "incomplete"
+            s3_path += s3_subdir
```
zephirefaith, line 84: Can we use `os.path.join` to avoid string magic?

examples/hitl/rearrange_v2/app_state_end_session.py
```diff
@@ -55,5 +63,58 @@ def sim_update(self, dt: float, post_sim_update_dict):
...
+            output_file_name = data_collection_config.output_file_name
+            output_file = f"{output_file_name}.json.gz"
+
+        except Exception as e:
+            print(f"Invalid data collection config. Skipping S3 upload. {e}")
+            return
+
+        # Delete previous output directory
+        if os.path.exists(output_folder):
+            shutil.rmtree(output_folder)
```
zephirefaith, line 99: Just making sure this will not delete useful data?

zephirefaith (approved):
I trust you've run this before so this works :)

Code LGTM, a couple nits.

examples/hitl/rearrange_v2/app_state_end_session.py
0mdc: This only contains the `.json.gz` file. You could however change the path in the config to any directory 🤔

PR #1973, opened 2024-05-29
nan guards in snap point APIs for navmesh_utils

aclegg3:
## Motivation and Context

Some navmesh_utils APIs were allowing nan vectors to slip through. The return values seemed sound anyway, but better to control the flow precisely instead of waiting for downstream bugs to manifest.

## How Has This Been Tested

Added tests failure cases

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
+++ b/habitat-lab/habitat/datasets/rearrange/navmesh_utils.py
@@ -40,6 +40,13 @@ def snap_point_is_occluded(
 
     :return: whether or not the target is considered occluded from the snap_point.
     """
+
+    if np.any(np.isnan(target)):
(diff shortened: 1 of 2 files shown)


jimmytyyang (approved):
LGTM!

Does this mean that the user needs to catch value error when we use this functin?

Thank you for this alex!

aclegg3:
> Does this mean that the user needs to catch value error when we use this functin?

Only if the user is not sure that their inputs to the functions are not `nan`. A check on the user side in advance would prevent that just as readily as a try/except.
Internally, `nan` values are guarded and won't be passed along if generated during sampling, so no worries about that.

0mdc (approved):
LGTM!

PR #1986, opened 2024-06-19
HITL - Update rearrange_v2 configs.

0mdc:
## Motivation and Context

This updates the rearrange_v2 configs to do the following:
* Disable new connections by default.
    * In this application, new connections are only enabled in the `Lobby` state. See #1964.
* Always show the humanoid in the GUI.
    * In the past, we had to hide the humanoid in first-person. This doesn't scale multiplayer applications.
    * Now, we can set object visibility. See #1959.
* Standardize max idle duration.
* Standardize episode iterator settings.

## How Has This Been Tested

Tested on local and remote HITL applications.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/config/rearrange_v2.yaml
+++ b/examples/hitl/rearrange_v2/config/rearrange_v2.yaml
@@ -43,6 +43,6 @@ habitat_hitl:
     - agent_index: 1
       lin_speed: 10.0
       ang_speed: 300
-  hide_humanoid_in_gui: True
+  hide_humanoid_in_gui: False
   camera:
     first_person_mode: True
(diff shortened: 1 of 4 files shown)


examples/hitl/rearrange_v2/config/rearrange_v2.yaml
```diff
@@ -43,6 +43,6 @@ habitat_hitl:
     - agent_index: 1
       lin_speed: 10.0
       ang_speed: 300
-  hide_humanoid_in_gui: True
+  hide_humanoid_in_gui: False
```
zephirefaith, line 46: What does this do? Why flip it?

zephirefaith (approved):

examples/hitl/rearrange_v2/config/rearrange_v2.yaml
0mdc: This removes the humans from the GUI application entirely. This is a simple way of handling human visibility for first-person single-user applications.

This doesn't scale to multiplayer applications because _all_ humans are removed. Now, we have fine-grained control over  object visibility per-viewport. See https://github.com/facebookresearch/habitat-lab/pull/1959.

PR #1908, opened 2024-04-11
HITL - Mouse-based UI.

0mdc:
## Motivation and Context

This adds a new UI to the `rearrange_v2` app with the following features:
* Mouse-based pick and place.
* Mouse-based open/close of articulated object links.
* Color coding of goal object-receptacle pairs.
* Refactoring.

---

## Features

Here are videos showing the overall changes. Note that the instruction text was disabled for visibility, and that highlight circle billboarding is only implemented in Unity.

**Double-click to open individual articulated object links:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/2abf1b16-1182-432c-bcae-90ffe83f37c3

**Double-click to pick objects, and release right click to place it:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/5c08ee09-ae9d-4b5a-9631-0ba9ac69b7d2

**While holding right click, an indicator shows up communicating whether the object can be placed:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/1a891cb5-d43e-421b-af61-f2a19c5e2359

**Experiment objects must be placed on specific receptacles. The pairs are color-coded:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/be07b754-7834-4f65-b890-d161ffb7b1ca

**When an action is possible, the color is green. Otherwise, it is red:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/d4024df8-65fd-46ff-97d2-79cf423a6545

**All of this works in Unity:**

https://github.com/facebookresearch/habitat-lab/assets/110583667/c38b6a62-aad9-4134-a0aa-1a6276069856

---

## How Has This Been Tested

Tested locally and with remote Unity client.

## Types of changes

- **\[Development\]**

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation if required.
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes if required.

PR #1988, opened 2024-06-20
HITL - Synchronize UI with grasp manager.

0mdc:
## Motivation and Context

The `RearrangeGraspManager` is used by agents to pick up objects or identify if an object is grabbed. Therefore, the objects grabbed in `rearrange_v2` are not visible to the agents, while the objects grabbed by agents can be interacted by the UI.

This changeset changes the `rearrange_v2` GUI such as the grabbed objects are synchronized with grasp managers.

**Note:**
A hacky flag is introduced in `RearrangeGraspManager`, `_automatically_update_snapped_object`, which toggles whether objects are automatically displaced. In HITL, we displace the objects ourselves to avoid occluding the view and communicate to the user that an object is grabbed (see #1908 for examples).
Ideally, the "grasped object" data should be separated from the object displacement item. However, this is a significant refactor. Let me know if you have ideas in the thread.

## How Has This Been Tested

Tested on single-learn and multi-player evaluation.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/world.py
+++ b/examples/hitl/rearrange_v2/world.py
@@ -122,3 +122,18 @@ class World:
             agent_object_ids.add(link_object_id)
 
         return agent_object_ids
+
+    def is_any_agent_holding_object(self, object_id: int) -> bool:
+        """
+        Checks whether the specified object is being held by an agent.
+        This function looks up both the HITL world state and grasp managers.
+        """
+        sim = self._sim
+        agents_mgr = sim.agents_mgr
+
+        for agent_index in range(len(agents_mgr.agent_names)):
+            grasp_mgr = agents_mgr._all_agent_data[agent_index].grasp_mgr
+            if grasp_mgr._snapped_obj_id == object_id:
+                return True
+
+        return object_id in self._all_held_object_ids
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
@@ -56,6 +56,10 @@ class RearrangeGraspManager:
         self._managed_articulated_agent = articulated_agent
         self.ee_index = ee_index
 
+        # HACK: This flag sets whether the grasp manager handles moving the grasped object.
+        # Turn off in applications that handle grasping themselves.
+        self._automatically_update_snapped_object = True
+
         self._kinematic_mode = self._sim.habitat_config.kinematic_mode
 
     def reconfigure(self) -> None:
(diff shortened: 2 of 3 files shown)


examples/hitl/rearrange_v2/world.py
```diff
@@ -122,3 +122,18 @@ def get_agent_object_ids(self, agent_index: int) -> Set[int]:
...
+        Checks whether the specified object is being held by an agent.
+        This function looks up both the HITL world state and grasp managers.
+        """
+        sim = self._sim
+        agents_mgr = sim.agents_mgr
+
+        for agent_index in range(len(agents_mgr.agent_names)):
+            grasp_mgr = agents_mgr._all_agent_data[agent_index].grasp_mgr
+            if grasp_mgr._snapped_obj_id == object_id:
+                return True
```
zephirefaith, line 137: This generally brings HitL + habitat-llm logic closer together ✔️

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -213,15 +217,16 @@ def snap_to_marker(self, marker_name: str) -> None:
...
-            self.create_hold_constraint(
-                RigidConstraintType.PointToPoint,
-                mn.Vector3(0.0, 0.0, 0.0),
-                mn.Vector3(*marker.offset_position),
-                marker.ao_parent.object_id,
-                marker.link_id,
-            ),
-        ]
+        if self._automatically_update_snapped_object:
+            self._snap_constraints = [
```
zephirefaith, line 221: Curious what is the use-case here?

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
```diff
@@ -299,7 +304,10 @@ def update_object_to_grasp(self) -> None:
         is grasped then nothing will happen.
         """
 
-        if self._snapped_obj_id is None:
+        if (
+            self._snapped_obj_id is None
+            or not self._automatically_update_snapped_object
+        ):
```
zephirefaith, line 310: Then is automatically updating == this code handling the update?

zephirefaith (approved):
Left a few questions, otherwise looks good.
I am assuming these are tested through phase3 deployment already, so logic is OK.

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
0mdc: This is for dynamic simulations. The other code path is for kinematic simulations.

habitat-lab/habitat/tasks/rearrange/rearrange_grasp_manager.py
0mdc: Correct.

Because this class is the mechanism for communicating the grasping state of each agent, this short circuit allows for external code to handle the transform updates.

PR #2002, opened 2024-07-24
When navigating, allow to list objects for which we will ignore collisions (e.g. other agents)

xavierpuigf:
## Motivation and Context
Need to be able to navigate to palces where another agent is.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/utils.py
+++ b/habitat-lab/habitat/tasks/rearrange/utils.py
@@ -94,7 +94,7 @@ class CollisionDetails:
 
 
 def general_sim_collision(
-        sim: habitat_sim.Simulator, agent_embodiment: MobileManipulator, ignore_names:  Optional[List[str]] = None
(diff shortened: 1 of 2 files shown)


habitat-lab/habitat/tasks/rearrange/utils.py
```diff
@@ -94,7 +94,7 @@ def __add__(self, other):
 
 
 def general_sim_collision(
-    sim: habitat_sim.Simulator, agent_embodiment: MobileManipulator
+        sim: habitat_sim.Simulator, agent_embodiment: MobileManipulator, ignore_names:  Optional[List[str]] = None
```
0mdc, line 97: `ignore_names` is a list of strings and appears to be requesting object handles.

Instead, we supply `object_id` integers.

I would rename this parameter as such: `ignored_object_ids : Optional[List[int]]`.

0mdc (approved):
Looks good, aside for the comment below. Feel free to merge after this is addressed.

PR #2018, opened 2024-08-20
Refactor to support expanded Simulator reset in RearrangeTask

aclegg3:
## Motivation and Context

This PR aims to refactor RearrangeTask to support the expanded reset functionality from https://github.com/facebookresearch/habitat-sim/pull/2451
Primary change: apply auto-sleep and handle articulated_agent base setting after expanded state reset

Context: previously the Simulator.reset() only set time to 0, so RearrangeTask code was able to leave the robot in place between reset and objects would stay sleeping (e.g. for benchmarking or idle scenarios). Now the objects are set to origin or scene_instance initial state and the reset logic must also handle initial state resetting and re-sleeping (since objects wake when moved).

## How Has This Been Tested

Previously the hab2_benchmark test was failing due to regression because the robot was resetting to the origin inside a furniture and objects were awoken on initial reset.

I expect to see the CI pass to validate the change.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_task.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_task.py
@@ -243,8 +243,11 @@ class RearrangeTask(NavigationTask):
             self._is_episode_active = True
 
             if self._should_place_articulated_agent:
+                # here we are randomizing the base pos and rotation if necessary
                 for agent_idx in range(self._sim.num_articulated_agents):
                     self._set_articulated_agent_start(agent_idx)
+            # here we are setting initial joint states and such from configs, if a configured fixed base start state was provided, it is set
+            self._sim.agents_mgr.post_obj_load_reconfigure()
 
         self.prev_measures = self.measurements.get_metrics()
         self._targ_idx = 0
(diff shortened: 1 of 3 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_task.py
```diff
@@ -243,8 +243,11 @@ def reset(self, episode: Episode, fetch_observations: bool = True):
             self._is_episode_active = True
 
             if self._should_place_articulated_agent:
+                # here we are randomizing the base pos and rotation if necessary
```
0mdc, line 246: That config field name is ambiguous. It may be worth the trouble renaming (not in this PR).

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
```diff
@@ -243,8 +243,11 @@ def reset(self, episode: Episode, fetch_observations: bool = True):
...
                 for agent_idx in range(self._sim.num_articulated_agents):
                     self._set_articulated_agent_start(agent_idx)
+            # here we are setting initial joint states and such from configs, if a configured fixed base start state was provided, it is set
+            self._sim.agents_mgr.post_obj_load_reconfigure()
```
0mdc, line 250: Same with this method name.

0mdc (approved):
LGTM.

habitat-lab/habitat/tasks/rearrange/rearrange_task.py
aclegg3: Yeah, this part of the code is ripe for refactor

PR #2019, opened 2024-08-20
HITL - Networking improvements.

0mdc:
## Motivation and Context

This changeset improves on network per-user messages:
* Allow activating or deactivating users from the `Users` container.
* Avoid sending user messages to inactive users.
* Clear user messages consolidated for the previous user when a new user joins.
* Avoid consolidating messages for inactive users.

## How Has This Been Tested

Unit test + HITL server.

## Types of changes

- **\[Bug Fix\]**
- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/pick_throw_vr/pick_throw_vr.py
+++ b/examples/hitl/pick_throw_vr/pick_throw_vr.py
@@ -59,6 +59,7 @@ class AppStatePickThrowVr(AppState):
         self._can_grasp_place_threshold = (
             self._app_service.hitl_config.can_grasp_place_threshold
         )
+        app_service.users.add_user(1)
 
         self._cam_transform: Optional[mn.Matrix4] = None
         self._held_target_obj_idx: Optional[int] = None
--- a/examples/hitl/pick_throw_vr/pick_throw_vr.py
+++ b/examples/hitl/pick_throw_vr/pick_throw_vr.py
@@ -60,7 +60,7 @@ class AppStatePickThrowVr(AppState):
             self._app_service.hitl_config.can_grasp_place_threshold
         )
         # Activate the remote user.
-        app_service.users.add_user(0)
+        app_service.users.activate_user(0)
 
         self._cam_transform: Optional[mn.Matrix4] = None
         self._held_target_obj_idx: Optional[int] = None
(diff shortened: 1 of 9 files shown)


habitat-hitl/habitat_hitl/_internal/networking/networking_process.py
```diff
@@ -231,18 +238,22 @@ async def send_keyframes(self) -> None:
             inc_keyframes_and_messages = (
                 self._interprocess_record.get_queued_keyframes()
             )
-            inc_keyframes = self._interprocess_record.get_queued_keyframes()
 
             if len(inc_keyframes_and_messages) > 0:
-                # consolidate all inc keyframes into one inc_keyframe
+                # Consolidate all inc keyframes into one inc_keyframe
                 tmp_con_keyframe = inc_keyframes_and_messages[0]
                 if len(inc_keyframes_and_messages) > 1:
```
eundersander, line 245: Seems odd to be discarding these messages only for the case of more than one keyframe.

habitat-hitl/habitat_hitl/core/user_mask.py
```diff
@@ -74,13 +80,22 @@ def indices(self, user_mask: Mask) -> Generator[int, Any, None]:
...
-        bitset = user_mask & self._max_user_mask
+        bitset = user_mask & self._active_user_mask
         while bitset != 0:
             user_bit = bitset & -bitset
             user_index = log2(user_bit)
             yield int(user_index)
             bitset ^= user_bit
 
+    def add_user(self, user_index: int) -> None:
+        """Activate an user."""
```
eundersander, line 91: Why not just call the function `activate_user`? (and `deactivate_user`).

examples/hitl/pick_throw_vr/pick_throw_vr.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, app_service: AppService):
         self._can_grasp_place_threshold = (
             self._app_service.hitl_config.can_grasp_place_threshold
         )
+        # Activate the remote user.
+        app_service.users.add_user(0)
```
eundersander, line 63: Again, it seems like you want to call this function `activate_user` based on this comment.

Also is there still a notion of a non-network mode for this PickThrowVR app? Does this line break something in that case?

eundersander (approved):
I left some comments including a possible bug, otherwise LGTM!

examples/hitl/pick_throw_vr/pick_throw_vr.py
0mdc: >Also is there still a notion of a non-network mode for this PickThrowVR app? Does this line break something in that case?

The `pick_throw_vr` application works just as before in local mode.

PR #2020, opened 2024-08-20
HITL - Fix choppy continuous remote input.

0mdc:
## Motivation and Context

This improves continuous inputs (scrolling, dragging) coming from remote clients.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f21c8c4c-9258-45f5-8168-e872326f6a19"> | <video src="https://github.com/user-attachments/assets/6acef5d6-3689-4dc1-bb7a-0c8ef31b7569"> |

## Context

Unlike most remote real-time applications, the HITL application is entirely written in the server, in Python. The client simply renders what it's given, and relays back the input. This is a design decision from the standpoint where having the code self-contained within Habitat is more important than optimizing input latency.

## Explanation

In local applications, inputs are updated at the simulator frequency. At the end of each step, we reset the input container (`GuiInput`) and controls behave as expected.

In remote applications, user input arrives at a different frequency and is not guaranteed.
1. Discrete input must be reset every frame. Otherwise, discrete events like `mouse_down` are activated across multiple frames.
2. Continuous input must be reset before a new input packet arrives. Otherwise, the continuous actions like mouse dragging abruptly stop on frames where input is not received.

**Previous State:**

| Frame | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
| - | - | - | - | - | - | - | - | - |
| Incoming |  |  | Scroll(2) |  | No Input |  | Scroll(3) |  |
| Scroll Value | 0 | 0 | 2 | 0 | 0 | 0 | 3 | 0 |

**Proposed State:**

| Frame | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
| - | - | - | - | - | - | - | - | - |
| Incoming |  |  | Scroll(2) |  | No Input |  | Scroll(3) |  |
| Scroll Value | 0 | 0 | 2 | 2 | 0 | 0 | 3 | 3 |

**Caveat:**
While this change improves controls over previous state, continuous inputs overshoot. This can be mitigated with client-side control or dampening.

## How Has This Been Tested

Tested on local and remote HITL application.

## Types of changes

## Checklist

### DIFF
--- a/habitat-hitl/habitat_hitl/_internal/gui_application.py
+++ b/habitat-hitl/habitat_hitl/_internal/gui_application.py
@@ -179,7 +179,12 @@ class GuiApplication(InputHandlerApplication):
 
         for _ in range(num_sim_updates):
             post_sim_update_dict = self._driver.sim_update(sim_dt)
-            self._gui_input.on_frame_end()
+
+            gui_input = self._gui_input
+            gui_input.on_frame_end()
+            gui_input._relative_mouse_position = [0, 0]
+            gui_input._mouse_scroll_offset = 0.0
+
             self._post_sim_update(post_sim_update_dict)
             if "application_exit" in post_sim_update_dict:
                 return
--- a/habitat-hitl/habitat_hitl/_internal/gui_application.py
+++ b/habitat-hitl/habitat_hitl/_internal/gui_application.py
@@ -179,12 +179,7 @@ class GuiApplication(InputHandlerApplication):
 
         for _ in range(num_sim_updates):
             post_sim_update_dict = self._driver.sim_update(sim_dt)
-
-            gui_input = self._gui_input
-            gui_input.on_frame_end()
-            gui_input._relative_mouse_position = [0, 0]
-            gui_input._mouse_scroll_offset = 0.0
-
+            self._gui_input.reset()
             self._post_sim_update(post_sim_update_dict)
             if "application_exit" in post_sim_update_dict:
                 return
--- a/habitat-hitl/habitat_hitl/core/gui_input.py
+++ b/habitat-hitl/habitat_hitl/core/gui_input.py
@@ -99,8 +99,6 @@ class GuiInput:
         self._key_up.clear()
         self._mouse_button_down.clear()
         self._mouse_button_up.clear()
-        self._relative_mouse_position = [0, 0]
-        self._mouse_scroll_offset = 0.0
 
     def copy_from(self, other: GuiInput):
         self._key_down = set(other._key_down)
--- a/habitat-hitl/habitat_hitl/core/gui_input.py
+++ b/habitat-hitl/habitat_hitl/core/gui_input.py
@@ -96,11 +96,10 @@ class GuiInput:
         """
         Reset the input states. To be called at the end of a frame.
 
-        reset_continuous_input: controls whether to reset non-discrete input like scrolling or dragging.
-        For local applications, this must be reset at the end of the frame.
-
-        Remote clients are out of sync with the HITL frame cadence and have their own input cycle.
-        Their continuous inputs must be reset before consolidating a new set of input packets.
+        `reset_continuous_input`: controls whether to reset continuous input like scrolling or dragging.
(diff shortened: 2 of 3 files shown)


habitat-hitl/habitat_hitl/_internal/gui_application.py
```diff
@@ -179,7 +179,12 @@ def draw_event(self):
 
         for _ in range(num_sim_updates):
             post_sim_update_dict = self._driver.sim_update(sim_dt)
-            self._gui_input.on_frame_end()
+
+            gui_input = self._gui_input
+            gui_input.on_frame_end()
+            gui_input._relative_mouse_position = [0, 0]
```
eundersander, line 185: This looks like a hack. Can you encapsulate this inside the GuiInput class?

eundersander (commented):
Can you explain what's going on here? So you're no longer clearing relative mouse position and mouse_scroll_offset from remote users. So why don't these values just keep accumulating to very large values? I must have misread the code somewhere.

habitat-hitl/habitat_hitl/core/gui_input.py
```diff
@@ -92,14 +92,25 @@ def mouse_scroll_offset(self):
...
-    def on_frame_end(self):
+    def reset(self, reset_continuous_input: bool = True):
+        """
+        Reset the input states. To be called at the end of a frame.
+
+        reset_continuous_input: controls whether to reset non-discrete input like scrolling or dragging.
+        For local applications, this must be reset at the end of the frame.
+
+        Remote clients are out of sync with the HITL frame cadence and have their own input cycle.
+        Their continuous inputs must be reset before consolidating a new set of input packets.
```
eundersander, line 103: I'm not sure I agree with this logic. I messaged you privately on Slack.

habitat-hitl/habitat_hitl/core/remote_client_state.py
```diff
@@ -246,6 +246,9 @@ def _update_input_state(
             if len(client_states) == 0:
                 continue
             gui_input = self._gui_inputs[user_index]
+            mouse_scroll_offset: float = 0.0
+            relative_mouse_position: List[int] = [0, 0]
```
zephirefaith, line 250: OK, so this is so continuous inputs do not accumulate but continue like a platform-shaped step-change.

habitat-hitl/habitat_hitl/core/remote_client_state.py
```diff
@@ -491,7 +495,7 @@ def get_new_connection_records(self) -> List[ConnectionRecord]:
 
     def on_frame_end(self) -> None:
         for user_index in self._users.indices(Mask.ALL):
-            self._gui_inputs[user_index].on_frame_end()
+            self._gui_inputs[user_index].reset(reset_continuous_input=False)
```
zephirefaith, line 498: Why is the `reset_continuous_input` hardcoded here? Move it to config or set at class init-time?

zephirefaith (approved):
Thanks, this makes sense. The video is great!
Let's not hard-code the config but otherwise approved.

habitat-hitl/habitat_hitl/core/remote_client_state.py
0mdc: What's going on here is the following:
* When the server steps (simulator set), multiple client messages may have been received.
* If we just pick the last received value, we lose discrete information like clicks.
* While not as critical, it follows that the correct aggregation strategy for continuous input is accumulating them.

habitat-hitl/habitat_hitl/core/remote_client_state.py
0mdc: I'm leaning towards avoiding to introduce an additional state to `GuiInput`.

The main upside of passing the parameter here is that it exposes the underlying complexity at the relevant line of code, as opposed to passing the argument via constructor.

I don't think that this should be driven by config either because there's no case where you would want to deviate from this implementation.

PR #2024, opened 2024-08-20
HITL - Handle task metrics.

0mdc:
## Motivation and Context

This changeset adds handling of task metrics:
* Add a thin wrapper around common metrics to access them safely.
* Use `task_percent_complete` for defining episode success.
* Log `task_percent_complete` and `task_explanation`.

## How Has This Been Tested

Tested on HITL applications.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/metrics.py
+++ b/examples/hitl/rearrange_v2/metrics.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from habitat_hitl.app_states.app_service import AppService
 
@@ -11,7 +11,7 @@ class Metrics:
     def __init__(self, app_service: AppService):
         self._get_metrics = app_service.get_metrics
 
-    def get_task_success(self) -> Optional[float]:
+    def get_task_percent_complete(self) -> Optional[float]:
         """
         Get the current progression of the task, from 0.0 to 1.0.
         Returns None if the metric is unavailable.
@@ -26,3 +26,9 @@ class Metrics:
         """
         metrics = self._get_metrics()
         return metrics.get("task_explanation", None)
+
+    def get_all_metrics(self) -> Dict[str, Any]:
+        """
+        Gets all available metrics.
+        """
+        return self._get_metrics()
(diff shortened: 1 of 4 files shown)


examples/hitl/rearrange_v2/metrics.py
```diff
@@ -0,0 +1,29 @@
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from habitat_hitl.app_states.app_service import AppService
+
+
+class Metrics:
```
eundersander, line 7: IMO a plural word is a bad choice for a class name. E.g. what if one day you need a list of these, what would you call that variable? `metricses`?

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -788,10 +804,22 @@ def _is_episode_finished(self) -> bool:
 
     def _is_episode_successful(self) -> bool:
         """
-        Returns true if all agents finished the episode successfully.
+        Returns true if:
+        * Task success is >99%.
```
eundersander, line 808: Why >99%? Looks like a magic number. How about a comment.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -788,10 +804,22 @@ def _is_episode_finished(self) -> bool:
...
+        * All agents finished the episode without reporting an error.
         """
-        return all(
+
+        task_percent_complete = self._metrics.get_task_percent_complete()
+        task_successful = (
```
eundersander, line 813: What is the real situation for `rearrange_v2` that this metric is None? How about add a comment explaining why sometimes that metric is missing and why we want to consider the task successful in that case.

eundersander (approved):
I left some comments otherwise LGTM!

examples/hitl/rearrange_v2/metrics.py
0mdc: We can rename the class.

I will do this separately as a refactor pass.

PR #2027, opened 2024-08-21
HITL - Combine `set_camera_transform()` and `show_viewport()` APIs.

0mdc:
## Motivation and Context

This changeset allows specifying the viewport when changing a camera transform.
The default `viewport_id` of `set_camera_transform()` is the main viewport, therefore has the same behavior as before.

This more straightforward approach avoids consolidation logic for viewport properties.

Depends on:
* https://github.com/0mdc/siro_hitl_unity_client/pull/36

## How Has This Been Tested

Tested on remote HITL application.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/habitat-hitl/habitat_hitl/core/client_message_manager.py
+++ b/habitat-hitl/habitat_hitl/core/client_message_manager.py
@@ -14,6 +14,7 @@ from habitat_hitl.core.user_mask import Mask, Users
 
 DEFAULT_NORMAL: Final[List[float]] = [0.0, 1.0, 0.0]
 DEFAULT_VIEWPORT_SIZE: Final[List[float]] = [0.0, 0.0, 1.0, 1.0]
+MAIN_VIEWPORT: Final[int] = -1
 
 
 # TODO: Move to another file.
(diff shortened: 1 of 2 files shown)


aclegg3 (approved):

eundersander (commented):
LGTM!

habitat-hitl/habitat_hitl/core/client_message_manager.py
```diff
@@ -335,10 +313,15 @@ def update_navmesh_triangles(
             ]
 
     def update_camera_transform(
```
eundersander, line 315: Maybe `set_camera_transform` would be better? I notice in the Unity PR that if you don't update on a given frame, then the viewport doesn't get rendered at all, is that right? I can imagine a user setting up a PIP to show a static overhead view, and then wondering why it isn't showing up.

0mdc: Good point.

I'm unsure whether this is self-explanatory due to the fact that most messages are immediate-mode. It's definitely worth a docstring update.

PR #2035, opened 2024-08-22
HITL - Add object states and categories.

0mdc:
## Motivation and Context

This changeset adds the following:

* API to query the region that contains most of an object.
* Placeholder mouse hover GUI.

https://github.com/user-attachments/assets/3cc2cec8-1388-4a56-b69c-b6e270e25d7a

## How Has This Been Tested

Tested on HSSD in HITL applications.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/ui.py
+++ b/examples/hitl/rearrange_v2/ui.py
@@ -512,11 +512,12 @@ class UI:
                 if primary_region is not None:
                     primary_region_name = primary_region.category.name()
 
-        if primary_region_name is not None:
+        if primary_region_name is not None and object_category is not None:
             # TODO: Draw UI
-            print(f"Region: {primary_region_name}")
-            print(f"Category: {object_category}")
-            print(f"States: {object_states}")
+            # print(f"Region: {primary_region_name}")
+            # print(f"Category: {object_category}")
+            # print(f"States: {object_states}")
+            pass
 
     def _draw_aabb(
         self, aabb: mn.Range3D, transform: mn.Matrix4, color: mn.Color3
--- a/examples/hitl/rearrange_v2/ui.py
+++ b/examples/hitl/rearrange_v2/ui.py
@@ -8,12 +8,15 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from datetime import datetime, timedelta
-from typing import TYPE_CHECKING, Callable, List, Optional, Tuple
+from typing import TYPE_CHECKING, Callable, List, Optional, Tuple, cast
 
 import magnum as mn
 from world import World
 
 from habitat.sims.habitat_simulator import sim_utilities
+from habitat.sims.habitat_simulator.object_state_machine import (
+    BooleanObjectState,
+)
 from habitat.tasks.rearrange.rearrange_sim import RearrangeSim
 from habitat_hitl.core.event import Event
 from habitat_hitl.core.gui_drawer import GuiDrawer
@@ -477,6 +480,8 @@ class UI:
         """Draw a UI when hovering an object with the cursor."""
         object_id = self._hover_selection.object_id
 
+        object_category: Optional[str] = None
+        object_states: List[Tuple[str, str]] = []
         primary_region_name: Optional[str] = None
 
         if object_id is not None:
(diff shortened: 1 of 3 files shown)


examples/hitl/rearrange_v2/world.py
```diff
@@ -165,3 +288,47 @@ def is_any_agent_holding_object(self, object_id: int) -> bool:
...
+                MetadataInterface,
+                get_metadata_dict_from_config,
+            )
+
+            metadata_dict = get_metadata_dict_from_config(
+                dataset_config=config.habitat.dataset
+            )
+            metadata_interface = MetadataInterface(metadata_dict)
+            metadata_interface.refresh_scene_caches(
+                sim, filter_receptacles=False
```
aclegg3, line 313: I think we are filtering everywhere else for the world_graph, so agent's won't try to set states on filtered Furniture. :thinking: 
Let's double check that or HitL folks could set states that agents wouldn't.

aclegg3 (approved):
This looks good from an initial pass. Let's check on the filter bool and maybe @MatthewChang and/or @zephirefaith should scan this as well for alignment with task details.

examples/hitl/rearrange_v2/ui.py
```diff
@@ -486,13 +491,32 @@ def _update_hovered_object_ui(self):
...
                 if primary_region is not None:
                     primary_region_name = primary_region.category.name()
 
-        if primary_region_name is not None:
+        if primary_region_name is not None and object_category is not None:
             # TODO: Draw UI
-            # print(primary_region_name)
+            # print(f"Region: {primary_region_name}")
+            # print(f"Category: {object_category}")
+            # print(f"States: {object_states}")
```
zephirefaith, line 519: remove these prints altogether or put them under debug/verbose flag?

examples/hitl/rearrange_v2/world.py
```diff
@@ -18,6 +27,12 @@
 from habitat_sim.scene import SemanticRegion
 
 
+@dataclass
+class ObjectStateInfo:
+    state_spec: ObjectStateSpec
+    value: Any
```
zephirefaith, line 33: Out of curiosity, why is the value `Any` here?

examples/hitl/rearrange_v2/world.py
```diff
@@ -97,6 +124,102 @@ def reset(self) -> None:
...
+        state_snapshot = self.get_state_snapshot_dict()
+
+        for state_id, handles in state_snapshot.items():
+            if object_handle in handles:
+                object_states[state_id] = handles[object_handle]
+
+        state_infos: List[ObjectStateInfo] = []
+        all_object_states = self.get_all_states()
+        for state, value in object_states.items():
+            state_spec = all_object_states[state]
```
zephirefaith, line 192: Is this like a semantically understandable string to name the state?

examples/hitl/rearrange_v2/world.py
zephirefaith: Agreed this should probably be true. In hab-llm `main` `CollaborationSim.reconfigure()` sets `filter_receptacles=True` before the start of every episode.

Let's keep this consistent.

zephirefaith (approved):
This looks good to me. +1 to @aclegg3's call out regarding receptacle filtering. I am not as plugged into the OSM as @MatthewChang but unless there are huge caveats this PR makes sense.

examples/hitl/rearrange_v2/ui.py
0mdc: This is going to be replaced by a GUI in a follow-up PR.

examples/hitl/rearrange_v2/world.py
0mdc: Object states can hold any type of value. We are currently only using booleans, but that may not be the case in the future.

examples/hitl/rearrange_v2/world.py
0mdc: This contains [this information](https://github.com/facebookresearch/habitat-lab/blob/main/habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py#L70).

PR #2052, opened 2024-08-29
[BE] - reactivate flake8 checks in rearrange_sim and correct issues

aclegg3:
## Motivation and Context

flake8 was turned off in rearrange_sim, this PR reactivates it and fixes issues for compliance

## How Has This Been Tested

pre-commit locally validated the formatting, CI should verify and unit tests should pass

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
@@ -21,12 +21,8 @@ from typing import (
 
 import magnum as mn
 import numpy as np
-import numpy.typing as npt
 
 import habitat_sim
-
-# flake8: noqa
-from habitat.articulated_agents.robots import FetchRobot, FetchRobotNoWheels
 from habitat.config import read_write
 from habitat.core.registry import registry
 from habitat.core.simulator import AgentState, Observations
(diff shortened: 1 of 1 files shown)


habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
```diff
@@ -279,7 +274,11 @@ def _update_markers(self) -> None:
             m.update()
 
     @add_perf_timing_func()
-    def reset(self):
+    def reset(self) -> None:
+        """
+        Reset the Simulator instance.
+        NOTE: this override does not return an observation.
```
xavierpuigf, line 280: Note something that this PR introduces, but wondering if here we should return some empty observation or a format that is compatible with the rest of resets.

xavierpuigf (approved):

0mdc (approved):

habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
aclegg3: Good question, by default python handles None, so I'm assuming the code is already robust against that given this code is fairly old. Honestly, why should reset return an observation anyway? Seems like something we should be asking for when we need it, not something provided without choice.
The top level reset or the Env reset can provide an observation if it is so important to have one for RL library compatibility.

PR #2061, opened 2024-09-01
Allow disabling agent sensors auto-update.

0mdc:
## Motivation and Context

In the current state, agent sensor transforms are automatically updated every frame based on the agent position.

In HITL and other GUI-controlled apps, the camera is controlled via GUI (e.g. first person camera). Here, it is desirable to synchronize the sensor transforms with the user view. Because the `Manipulator` objects override the transforms, there is a discrepancy between the _real_ camera and the sensors.

This changeset introduces a configuration field that disables this behavior so that HITL can control the sensor transforms.

**Note:** This is hacky, but I did not find a cleaner solution that doesn't involve refactoring. This is nonetheless better than carrying cherry-picks in experiments. Suggestions are welcome.

## How Has This Been Tested

Tested on HITL single-learn application.

## Types of changes

- **\[Bug Fix\]**
- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/config/language_rearrange.yaml
+++ b/examples/hitl/rearrange_v2/config/language_rearrange.yaml
@@ -62,6 +62,10 @@ habitat:
     agents_order:
       - agent_0
 
+    agents:
+      agent_0:
+        auto_update_sensor_transform: False
+
     kinematic_mode: True
     ac_freq_ratio: 1
     step_physics: False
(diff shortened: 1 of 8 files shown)


zephirefaith (approved):
Offline conversation: this is one of 2 PRs to implement config and then the logic. Approving.

Thanks for this @0mdc !

aclegg3 (approved):
This hack is fine for now to unblock deployment, but it will need to be refactored once the agent/sensor redesign makes the parenting between robot nodes and camera sensors implicit by embedding them in the SceneGraph.

Request: please add some TODOs to the docstrings and relevant code to mark this as a patch scheduled for deprecation.

0mdc:
>This hack is fine for now to unblock deployment, but it will need to be refactored once the agent/sensor redesign makes the parenting between robot nodes and camera sensors implicit by embedding them in the SceneGraph.

@aclegg3 Agreed! Let's consider this as a use case for sensor redesign.

PR #2064, opened 2024-09-03
[bugfix] - re-initialize the KRM from episode contents in rearrange_sim 

aclegg3:
## Motivation and Context

re-initialize the KRM from episode contents in rearrange_sim any time add_objs is called to add new objects or initialize a new scene

## How Has This Been Tested
CI? Mikael will validate against test cases

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
+++ b/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py
@@ -706,15 +706,15 @@ class RearrangeSim(HabitatSim):
                     for motor_id in ao.existing_joint_motor_ids:
                         ao.remove_joint_motor(motor_id)
                 self.art_objs.append(ao)
-            if self._kinematic_mode:
-                # initialize KRM with parent->child relationships from the RearrangeEpisode
-                self.kinematic_relationship_manager = (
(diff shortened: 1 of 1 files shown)


0mdc (approved):
LGTM! Works on my side.

[Here](https://github.com/facebookresearch/habitat-lab/blob/f529be1f5037c147c53bcfbe4128b3fec5ba48ba/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py#L320), we also re-create the KRM on the `new_scene` and `should_add_objects` conditions. Does this cover another edge case I'm not aware of?

aclegg3:
> [Here](https://github.com/facebookresearch/habitat-lab/blob/f529be1f5037c147c53bcfbe4128b3fec5ba48ba/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py#L320), we also re-create the KRM on the `new_scene` and `should_add_objects` conditions. Does this cover another edge case I'm not aware of?

KRM is destroyed when a new scene is loaded because it will definitely be outdated but no objects are available to be added, so we don't want to init. Otherwise, once new objects are added to a stale scene, or a fresh scene is loaded with new objects, a simple re-initialization should be enough.

PR #2067, opened 2024-09-05
New function to generate humanoid walking motion

xavierpuigf:
## Motivation and Context
New method to select humanoid walking motion without changing humanoid pace. This change allows to better control humanoid using speed, thus providing better support for HITL and a controller more similar to the robot one.

## How Has This Been Tested

Ran Base Velocity with it.

## Types of changes

## Checklist

- [ x] My code follows the code style of this project.
- [ x] I have updated the documentation if required.
- [ x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ x] I have completed my CLA (see **CONTRIBUTING**)
- [ x] I have added tests to cover my changes if required.

### DIFF
--- a/habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
+++ b/habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
@@ -32,6 +32,23 @@ DIST_TO_STOP: float = (
 from typing import List, Tuple
 
 
+def bin_search(v, a, b, target_value):
+    left = a
+    right = b
+
+    while left <= right:
+        mid = (left + right) // 2
+
+        if v[mid] == target_value:
+            return mid
+        elif v[mid] < target_value:
+            left = mid + 1
+        else:
+            right = mid - 1
+
+    return min(left, len(v) - 1)
+
+
 class HumanoidRearrangeController(HumanoidBaseController):
     """
     Humanoid Controller, converts high level actions such as walk, or reach into joints positions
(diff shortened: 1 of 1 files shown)


habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -32,6 +32,23 @@
 from typing import List, Tuple
 
 
+def bin_search(v, a, b, target_value):
```
akshararai, line 35: docstring?

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -279,6 +296,80 @@ def calculate_walk_pose(
         self.obj_transform_base = obj_transform_base @ rot_offset
         self.joint_pose = joint_pose
 
+    def translate_and_rotate_with_gait(
```
0mdc, line 299: This function is not being called anywhere. It probably should be the default from now on.

It would be worth replacing all calls to `calculate_walk_pose_directional` and deprecating it.

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -279,6 +296,80 @@ def calculate_walk_pose(
...
+        self, forward_delta: float, rot_delta: float
```
0mdc, line 300: Because agent velocity is driven from config, we should consider having `forward_delta` a value between `-1.0` and `1.0`, and having this function multiply it with the agent's `lin_speed`.

If we do this, there would be a single source of truth to determine the humanoid agent's velocity.
This would eliminate any discrepancy caused by the implementer failing to read the right config field, and would ease agent config refactoring.

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
```diff
@@ -32,6 +32,23 @@
...
+        mid = (left + right) // 2
+
+        if v[mid] == target_value:
+            return mid
+        elif v[mid] < target_value:
+            left = mid + 1
+        else:
+            right = mid - 1
+
+    return min(left, len(v) - 1)
```
0mdc, line 49: It would be worthwhile unit testing this one.

We could also use [bisect](https://docs.python.org/3.11/library/bisect.html) to solve this.

0mdc (commented):
Some more comments

0mdc (approved):
I think that this is good to go, because there is no side-effect.

Let's get this merged to get you unblocked, and revisit my concerns as we deploy the code across the codebase.

habitat-lab/habitat/articulated_agent_controllers/humanoid_rearrange_controller.py
xavierpuigf: This multiplication happens in the BaseVelocityAction, this way we have parity with how the robot moves around.
Here is where we calculate how much to move: https://github.com/facebookresearch/habitat-llm/blob/main_improve_planner/habitat_llm/agent/env/actions.py#L116


This is similar to the robot action, which computes delta and then transforms in here:
https://github.com/facebookresearch/habitat-lab/blob/main/habitat-lab/habitat/tasks/rearrange/actions/actions.py#L489

PR #2058, opened 2024-08-31
HITL - Add selection panel and object state manipulation UI.

0mdc:
## Motivation and Context

This changeset adds a selection panel and object state manipulation controls.
The object state change event is recorded to HITL output.

The selection panel shows:
* Object name.
* Region name.
* Contextual information.
    * Contextual controls.
    * Why an action (pick/place/open/close) cannot be done.
* Object state toggles.

Builds on top of the following recent changes:
* https://github.com/facebookresearch/habitat-lab/pull/2031
* https://github.com/facebookresearch/habitat-lab/pull/2035
* https://github.com/facebookresearch/habitat-lab/pull/2041
* https://github.com/facebookresearch/habitat-lab/pull/2042

**Held object contextual information:**

https://github.com/user-attachments/assets/b69c3980-d484-4ac2-8459-40182999d895

**Object state manipulation:**

https://github.com/user-attachments/assets/ea93de0c-d50d-43f4-aa1e-1b4b987be448

## How Has This Been Tested

Tested on multiplayer HITL server with Unity clients.

## Types of changes

- **\[Development\]**

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation if required.
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes if required.

PR #2069, opened 2024-09-05
HITL - Change object state check function signature.

0mdc:
## Motivation and Context

This changes the object state check function signature.
A result struct is used instead of a tuple.

## How Has This Been Tested

Tested locally.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/object_state_manipulator.py
+++ b/examples/hitl/rearrange_v2/object_state_manipulator.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional, cast
 
 from habitat_llm.sims.collaboration_sim import CollaborationSim
 from habitat_llm.tools.motor_skills.object_states.oracle_clean_skills import (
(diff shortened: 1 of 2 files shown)


examples/hitl/rearrange_v2/ui.py
zephirefaith, line 1: Do we already have the UI knobs locally which show the available object-states and the ability to click on them?

examples/hitl/rearrange_v2/object_state_manipulator.py
```diff
@@ -122,36 +128,41 @@ def get_action(
...
         """
-        success, error_message = self.can_execute_action(
+        result = self.can_execute_action(
             state_name, state_value, object_handle
         )
-        if success:
+        if result.success:
             self.set_object_state(object_handle, state_name, state_value)
-        return (success, error_message)
+        return ObjectStateChangeResult(result.success, result.error_message)
```
zephirefaith, line 165: If result is not a success, is the error communicated anywhere to the user?

zephirefaith (approved):
Looks good! Thanks! Left a comment.

examples/hitl/rearrange_v2/object_state_manipulator.py
0mdc: No, in this case, it's just failing silently.

The UI calls `can_execute_action` every frame to check whether the user can execute the action. The message is passed here to be reactive to user's actions.

examples/hitl/rearrange_v2/ui.py
0mdc: Yes!

* https://github.com/facebookresearch/habitat-lab/pull/2058

PR #2070, opened 2024-09-06
HITL - Configure camera height based on the agent height.

0mdc:
## Motivation and Context

This changeset sets the HITL first person camera height from the agent type they are controlling.

This is hacky - This will be moved into a purposeful API in the HITL controllers.

## How Has This Been Tested

Tested locally.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/rearrange_v2.py
+++ b/examples/hitl/rearrange_v2/rearrange_v2.py
@@ -26,6 +26,9 @@ from ui import UI
 from util import UP
 from world import World
 
+from habitat.articulated_agents.humanoids.kinematic_humanoid import (
+    HUMANOID_CAMERA_HEIGHT_OFFSET,
+)
 from habitat_hitl._internal.networking.average_rate_tracker import (
     AverageRateTracker,
 )
@@ -269,10 +272,6 @@ class UserData:
         # TODO: Communicate to the controller via action hints.
         gui_agent_controller._gui_input = self.gui_input
 
-    def reset(self):
-        self.camera_helper.update(self._get_camera_lookat_pos(), dt=0)
-        self.ui.reset()
-
         # If networking is enabled...
         if self.app_service.client_message_manager:
             # Assign user agent objects to their own layer.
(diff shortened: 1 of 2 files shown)


examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -365,11 +367,32 @@ def pop_ui_events(self) -> List[Dict[str, Any]]:
         return events
 
     def _get_camera_lookat_pos(self) -> mn.Vector3:
+        # HACK: Estimate camera height.
+        # TODO: GuiController should have knowledge on the agent type it controls, and should provide an API to get the camera height.
+        gui_agent_controller = self.gui_agent_controller
+        sim = self.app_service.sim
+        agent_data = self.agent_data
+        camera_height_offset = 1.0
+        if isinstance(gui_agent_controller, GuiRobotController):
```
0mdc, line 376: @zephirefaith : Beware that until this code is handled at the right place, you'll have to check for `LLMController`.

examples/hitl/rearrange_v2/rearrange_v2.py
```diff
@@ -365,11 +367,32 @@ def pop_ui_events(self) -> List[Dict[str, Any]]:
...
+            # The robot camera height is defined in `spot_robot.py`.
+            # It's directly attached to the 6th articulated object link, without a height offset.
+            agent = sim.agents_mgr[agent_data.agent_index].articulated_agent
+            root = agent.sim_obj
+            head_link = root.get_link_scene_node(6)
+            head_link_height = head_link.absolute_translation.y
+            base_link = root.get_link_scene_node(-1)
+            base_link_height = base_link.absolute_translation.y
+            camera_height_offset = head_link_height - base_link_height
+        elif isinstance(gui_agent_controller, GuiHumanoidController):
```
zephirefaith, line 386: I think I can just expand this condition to trigger when `gui_agent_controller` is `LLMController`

zephirefaith (approved):
Makes sense. Thanks for figuring this out!

examples/hitl/rearrange_v2/rearrange_v2.py
0mdc: Correct!
Eventually, the controller will know about the agent it's controlling and will be able to get the correct camera height.
Better have this hack in and expose the complexity for now.

PR #2072, opened 2024-09-06
[do not merge] - update default_link util 

aclegg3:
## Motivation and Context

Current implementation chooses the link with the lowest global bounding box corner y value. When an AO has multiple links which should be equivalent, there is potential for floating point error to determine which one gets picked. 

This change considers 1cm difference in y value or less as equal and then breaks ties using lowest link index which should result in more deterministic behavior.

**NOTE: This is a breaking change since some previously detected default_links may have changed with the index tie breaker.**

TODO: all HSSD receptacle filter files with automatically computed active sets using default_link will need to be updated.

## How Has This Been Tested

Locally

## Types of changes

## Checklist


0mdc (approved):
LGTM!

aclegg3:
Do not merge.
This change introduced too much thrash in existing episodes, postponing or closing at a later time.

aclegg3:
closing this change since it doesn't seem relevant anymore

PR #2075, opened 2024-09-07
HITL - Highlight default receptacles

0mdc:
## Motivation and Context

This changeset adds a contextual UI that shows the default receptacle.

The following changes are included:
* Add support for multiple contextual info labels.
* Add default receptacle highlight and contextual info.
* Add feature flag.

https://github.com/user-attachments/assets/ac71dd3e-a523-4d37-a05b-8254c4ebf06a

## How Has This Been Tested

Tested on HITL application.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/examples/hitl/rearrange_v2/ui.py
+++ b/examples/hitl/rearrange_v2/ui.py
@@ -798,7 +798,7 @@ class UI:
                     if self._ui_settings.highlight_default_receptacles:
                         contextual_info.append(
                             (
-                                "Highlighted receptacle\nmay contain objects.",
+                                "Yellow container\nmay have objects.",
                                 color_ui_object_info,
                             )
                         )
--- a/examples/hitl/rearrange_v2/ui.py
+++ b/examples/hitl/rearrange_v2/ui.py
@@ -66,6 +66,16 @@ COLOR_HIGHLIGHT = mn.Color4(_HI, _HI, _LO, _HI)  # Yellow
 COLOR_SELECTION = mn.Color4(_LO, _HI, _HI, 1.0)  # Cyan
 
 
+@dataclass
+class UISettings:
+    """
+    Configuration of the UI.
+    """
+
+    can_change_object_states: bool
+    highlight_default_receptacles: bool
+
+
 class UI:
     """
     User interface for the rearrange_v2 app.
@@ -83,7 +93,7 @@ class UI:
         gui_input: GuiInput,
         gui_drawer: GuiDrawer,
         camera_helper: CameraHelper,
-        can_change_object_states: bool,
+        ui_settings: UISettings,
     ):
         self._user_index = user_index
         self._dest_mask = Mask.from_index(self._user_index)
@@ -93,7 +103,7 @@ class UI:
         self._gui_input = gui_input
         self._gui_drawer = gui_drawer
         self._camera_helper = camera_helper
-        self._can_change_object_states = can_change_object_states
+        self._ui_settings = ui_settings
 
         self._can_grasp_place_threshold = hitl_config.can_grasp_place_threshold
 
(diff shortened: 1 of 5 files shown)


examples/hitl/rearrange_v2/ui.py
```diff
@@ -696,20 +705,29 @@ def _update_selected_object_ui(self):
...
+                                )
                             )
-                            contextual_color = color_ui_invalid
                         else:
                             if self._is_within_reach(obj.translation):
-                                contextual_info = "Double-click to pick up."
-                                contextual_color = color_ui_valid
+                                contextual_info.append(
+                                    (
+                                        "Double-click to pick up.",
```
akshararai, line 723: should this be color_ui_info?

examples/hitl/rearrange_v2/ui.py
```diff
@@ -696,20 +705,29 @@ def _update_selected_object_ui(self):
                 # Get the contextual information.
                 color_ui_valid = [0.2, 1.0, 0.2, 1.0]
                 color_ui_invalid = [1.0, 0.2, 0.2, 1.0]
+                color_ui_info = None
+                color_ui_object_info = [0.8, 0.8, 0.2, 1.0]
```
akshararai, line 709: hm maybe we should only have ` color_ui_info` and use it in all places? rather than have ` color_ui_info` and `color_ui_object_info`?

aclegg3 (approved):

examples/hitl/rearrange_v2/ui.py
akshararai: Also let;s make these the same as the colors defined at the top of this file, rather than re-define them here?

examples/hitl/rearrange_v2/ui.py
```diff
@@ -1019,6 +1063,36 @@ def _draw_selected_object_highlights(self) -> None:
...
+                ):
+                    default_link_node = obj.get_link_scene_node(
+                        default_link_index
+                    )
+                    if (
+                        default_link_node.object_semantic_id
+                        not in self._world._opened_link_set
+                    ):
+                        color_default_link = self._to_color_array(
+                            COLOR_HIGHLIGHT
```
akshararai, line 1085: I think ok for now, but maybe in the future, should be a config variable?

akshararai (approved):
looks good, left minor comments. I think in the future, we should make these features behind flags that can be set in configs, so that its easy to turn them on or off, but looks good for now.

examples/hitl/rearrange_v2/ui.py
0mdc: The color theme is the following:
* White: Generic information.
* Yellow: Anything related to a pickable object.
* Cyan: Selection.
* Green: Anything related to a possible action.
* Red: Anything related to an impossible action.

I put this label green to match with the green highlight that indicates that an object can be picked up.

https://github.com/user-attachments/assets/ba312739-3802-4d27-b1d7-9c2c03671260

examples/hitl/rearrange_v2/ui.py
0mdc: >Also let;s make these the same as the colors defined at the top of this file, rather than re-define them here?

I'll refactor that in a later pass.
1. Those 2 colors are in different formats. This will be unified.
2. The contextual information duplicates the real checks. This will be unified.

0mdc: >hm maybe we should only have  color_ui_info and use it in all places? rather than have  color_ui_info and color_ui_object_info?

I think that it's preferable to have contextual information related to pickable objects yellow, like the pickable object highlights.

examples/hitl/rearrange_v2/ui.py
0mdc: Agreed!
I introduced UI configuration in this PR. I intend to make it fully configurable in the following weeks.

PR #2078, opened 2024-09-13
Filter deactivated sensors.

0mdc:
## Motivation and Context

This change ensures that only "active" sensors are rendered.

In lab, only the sensors listed in `gym.obs_keys` can be read. However, all other sensors are still rendered.
Therefore, when pulling agents from configs, **all their sensors are rendered**.

For example, if a human and spot are instantiated, the following sensors are rendered:
image (6)

## Performance

Here's the difference on a single-learn HITL session with 3 visual sensors.

| Before  | After |
| ------------- | ------------- |
| sensor_filter_before2 | sensor_filter_after |

## How Has This Been Tested

Tested locally and profiled.
**_Running unit tests on CI._**

## Types of changes

- **\[Bug Fix\]**

## Checklist

### DIFF
--- a/habitat-lab/habitat/core/env.py
+++ b/habitat-lab/habitat/core/env.py
@@ -111,6 +111,18 @@ class Env:
         else:
             self.number_of_episodes = None
 
+        # Filter out inactive sensors from habitat-sim config.
+        active_sensors = config.gym.obs_keys
+        agent_configs = config.simulator.agents
+        with read_write(agent_configs):
+            for agent_name, agent_config in agent_configs.items():
+                active_sim_sensors = {}
+                for sensor_key, sensor_config in agent_config.sim_sensors.items():
+                    sensor_uuid = f"{agent_name}_{sensor_config.uuid}"
+                    if sensor_uuid in active_sensors:
+                        active_sim_sensors[sensor_key] = sensor_config
+                agent_config.sim_sensors = active_sim_sensors
+
         self._sim = make_sim(
             id_sim=self._config.simulator.type, config=self._config.simulator
         )


habitat-lab/habitat/config/default_structured_configs.py
```diff
@@ -1887,6 +1887,10 @@ class GymConfig(HabitatBaseConfig):
     action_keys: Optional[List[str]] = None
     achieved_goal_keys: List = field(default_factory=list)
     desired_goal_keys: List[str] = field(default_factory=list)
+    cull_visual_sensors: Optional[bool] = False
```
eundersander, line 1890: How about `cull_unused_visual_sensors`

habitat-lab/habitat/core/env.py
```diff
@@ -111,6 +111,22 @@ def __init__(
         else:
             self.number_of_episodes = None
 
+        # Filter out inactive sensors from habitat-sim config.
```
eundersander, line 114: How about `Filter out visual sensors from the habitat-sim config that aren't used by this Gym env.`

habitat-lab/habitat/config/default_structured_configs.py
eundersander: How about a warning comment that explains the dependency issue. You can mention humanoid_detector_sensor and spot_head_stereo_depth_sensor.

eundersander: Also I would expect the comment for this flag to go above it, not below it.

habitat-lab/habitat/core/env.py
eundersander: It would be nice to use the terms "active" and "inactive" consistently, or not introduce these terms at all. Does inactive mean not used or not rendered? What do you call a visual sensor that is unused but still rendered (i.e. if you don't have this cull flag set)?

habitat-lab/habitat/core/env.py
```diff
@@ -111,6 +111,22 @@ def __init__(
...
+        if config.gym.cull_visual_sensors:
```
eundersander, line 115: Doing this cull logic here in `Env.__init__`... I wonder if this is too late and not ideal. For example, what about our call to `envs.initialize_batch_renderer` which also reads sim_sensors? Will it see the updated sim_sensors? For reference, when I accidentally re-implemented this recently, I put the culling in `default.py patch_config`, which is much earlier in our loading path. I don't know if that's better.

eundersander (commented):
I left some comments.

habitat-lab/habitat/config/default_structured_configs.py
0mdc: >Also I would expect the comment for this flag to go above it, not below it.

You right that in this file, only class docstrings are under the symbol, so I'll change accordingly.
However, comments above the attributes aren't recognized as docstrings.

attribute_docstrings

habitat-lab/habitat/core/env.py
0mdc: I agree. There is some unwrangling to do here.

This pattern of changing configs on-the-fly is causing us headaches with sensors. @aclegg3  and @jturner65 are currently working on refactoring them.

For example:

`rearrange_sim` changes the UUIDs here: https://github.com/facebookresearch/habitat-lab/blob/f41187690ac6607dede2bd50cf50f2360f40b223/habitat-lab/habitat/tasks/rearrange/rearrange_sim.py#L68

UUIDs get changed again when they are passed to sim here: https://github.com/facebookresearch/habitat-lab/blob/f41187690ac6607dede2bd50cf50f2360f40b223/habitat-lab/habitat/sims/habitat_simulator/habitat_simulator.py#L371

@aclegg3 let's consider this as we refactor agents and sensors.

eundersander: So what's your plan here? Are you going to make a change here before merging this PR?

aclegg3: Sensor refactor won't be "done" for a few weeks. I suggest you move forward with a stop-gap if you need this change to be around soon. If this is just living on a branch or not really needed in the short term then you could wait. A similar mechanism is planned to be included in the sensor refactor.

0mdc: @eundersander I would then go and merge this as-is.

PR #2079, opened 2024-09-15
Add a cache to ObjectStateMachine.

0mdc:
## Motivation and Context

This adds a cache to `ObjectStateMachine` so that `get_snapshot_dict()` doesn't re-compute the same snapshot over and over.

This increases the current single-learn framerate from `11.2` to `12.3` FPS.

## How Has This Been Tested

Tested locally + unit tests.

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
+++ b/habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
@@ -269,6 +269,10 @@ class ObjectStateMachine:
         self.objects_with_states: Dict[
             str, List[ObjectStateSpec]
         ] = defaultdict(lambda: [])
+        # Whether the snapshot cache is dirty.
+        self._dirty = False
+        # Snapshot cache.
+        self._snapshot_cache: Dict[str, Dict[str, Any]] = {}
 
     def initialize_object_state_map(self, sim: habitat_sim.Simulator) -> None:
         """
@@ -299,6 +303,8 @@ class ObjectStateMachine:
                     f"registered state {state} for object {obj.handle}"
                 )
 
+        self._dirty = True
+
     def update_states(self, sim: habitat_sim.Simulator, dt: float) -> None:
         """
         Update all tracked object states for a simulation step.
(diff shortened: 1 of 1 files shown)


habitat-lab/habitat/sims/habitat_simulator/object_state_machine.py
```diff
@@ -350,4 +361,7 @@ def get_snapshot_dict(
                     if obj_state is not None
                     else state.default_value()
                 )
-        return dict(snapshot)
+
+        self._snapshot_cache = dict(snapshot)
+        self._dirty = False
+        return dict(self._snapshot_cache)
```
aclegg3, line 367:
```suggestion
        return self._snapshot_cache
```
Already converted to dict at line 365, right?

aclegg3 (changes requested):
As it stands, this isn't going to catch cases where the states are updated externally to the state machine, such as action updates. I don't know that caching is the best solution unless the `ObjectStateSpec` singletons also map back to the parent `ObjectStateMachine` in order modify the dirty flag themselves. Even then, most of the updates use `set_state_of_obj` directly which doesn't hit either singleton class. For example: https://github.com/facebookresearch/habitat-lab/blob/fe8bfda97eb5f5072c3a7138a64b81ac3467a661/examples/hitl/rearrange_v2/object_state_manipulator.py#L95

None of these will be caught and the update() function is called in the simulation loop, so you're really only catching that "step" has been called since last query I think.

examples/hitl/rearrange_v2/world.py
```diff
@@ -144,27 +142,18 @@ def update(self, dt: float) -> None:
...
-        }
         """
-        return self._state_snapshot_dict
+        sim = self._sim
+        osm = self._object_state_machine
+
+        if osm is not None:
+            return osm.get_snapshot_dict(sim)
+        else:
+            return {}
```
0mdc, line 156: This cache is no longer necessary.

0mdc (commented):
Added some comments for reviewers.

PR #2080, opened 2024-09-15
Add simulator physics collision cache.

0mdc:
## Motivation and Context

This adds a cache for the `habitat-sim` `get_physics_contact_points()` function.

This function is called multiple times a frame and is quite expensive.

This increases the current single-learn HITL framerate from `11.3` to `13.8` FPS.

## How Has This Been Tested

Tested locally + unit tests.

## Types of changes

- **\[Optimization\]**

## Checklist

### DIFF
--- a/habitat-lab/habitat/sims/habitat_simulator/habitat_simulator.py
+++ b/habitat-lab/habitat/sims/habitat_simulator/habitat_simulator.py
@@ -44,6 +44,7 @@ from habitat.core.simulator import (
     VisualObservation,
 )
 from habitat.core.spaces import Space
+from habitat_sim.physics import ContactPointData
 
 if TYPE_CHECKING:
     from torch import Tensor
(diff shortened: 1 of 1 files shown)


aclegg3 (commented):
My other concern here is that moving objects will invalidate the cache and any calls to [Simulator.perform_discrete_collision_detection](https://aihabitat.org/docs/habitat-sim/habitat_sim.simulator.Simulator.html#perform_discrete_collision_detection) will be missed.
Perhaps you want to override that function also to modify the dirty flag.
I think at that point the likelihood of erroneous contact point data should be minimized.

0mdc:
>My other concern here is that moving objects will invalidate the cache and any calls to [Simulator.perform_discrete_collision_detection](https://aihabitat.org/docs/habitat-sim/habitat_sim.simulator.Simulator.html#perform_discrete_collision_detection) will be missed.

Shouldn't this function change the world time?

PR #2083, opened 2024-09-19
[bugfix] in support surface selection and KRM hard failure 

aclegg3:
## Motivation and Context

Fixes a bug causing the last hit to be used as the default support surface for rec match. Usually this is the stage, so no matches were being found with furniture.

Also prevents program crash when Episode contains a non-active object->receptacle match. Logs an error and skips the kinematic relationship creation.

## How Has This Been Tested

Locally.

## Types of changes

## Checklist


0mdc (approved):
LGTM! Thank you!

PR #2084, opened 2024-09-23
[BE] - add humnaoid data to CI download to activate the test_humanoid.py pytest

aclegg3:
## Motivation and Context

I realized that test_humanoids.py pytest was being skipped on CI because the humanoid data was missing from the download. Added this back and now expecting the tests to run.

## How Has This Been Tested

Should run the tests on CI now.
TODO: I noticed this fails locally due to missing config fields. Fix incoming.

## Types of changes

- **\[Refactoring\]** Large changes to the code that improve its functionality or performance
- **\[Bug Fix\]** (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation if required.
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes if required.

PR #2085, opened 2024-09-23
[BE] - fix inconsistent base_rot property 

aclegg3:
## Motivation and Context

This PR addresses a long-standing bug in the `articulated_agent.base_rot` property API.

**The bug:** :bug: 
Rotations are stored internally as `Magnum.Quaternion` objects. The `base_rot` property operates in scalar space, namely the angle of rotation about the Y axis. This means that every call to the setter creates a Quaternion from a rotation angle and Y axis, while the getter deconstructs a Quaternion into an angle [using 2arccos](https://magnum.graphics/doc/magnum/classMagnum_1_1Math_1_1Quaternion.html#ae8e09c06b590ce0c8676f83ed771d395). 
While the result is correct in some sense, it does not account for:
1. Inversion of the axis
2. redundancies in SO3
This results in the potential for several nasty bugs. The most obvious being that getting an angle after settings does not return the same scalar value (it could be a correct, but different angle) and could return an erroneous value (e.g. if the quaternion axis is inverted during normalization).

**The solution:**
This PR modifies the getter to:
1. check for inverted Y axis and negate the angle if detected. This prevents cases where an erroneous angle would be returned due to unknown inversion of the axis.
2. clamping the output range to [-2pi, 2pi] by adding or subtracting 2pi overflows.

Together, these changes result in a few nice guarantees from the function:
1. Testable consistency between the getter and setter within the range (-pi, pi). Note the end points are excluded because `-pi ==pi`. This means, e.g. `base_rot = pi/2` guarantees `base_rot == pi/2`.
2. The getter can be used in the setter to increment the rotation like: `base_rot = base_rot + 0.1` with the expected outcome of continual rotation.
4. The getter will always return a value in the range [-2pi, 2pi] and it will always be an equivalent rotation to that which was set (though the scalar value may be different outside the range (-pi,pi))

## How Has This Been Tested

Added unit tests for robot wrapper and humanoid wrapper.
This includes changes from #2084. Note that CI was failing there and is now passing. 

https://github.com/user-attachments/assets/68aa6a40-3d74-4635-bcf0-5645f757c483

## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
+++ b/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
@@ -184,10 +184,13 @@ class ArticulatedAgentBase(ArticulatedAgentInterface):
         # NOTE: if the quaternion axis is inverted (-Y) then the angle will be negated
         if self.sim_obj.rotation.axis()[1] < 0:
             angle = -1 * angle
-        while angle > mn.math.pi:
+        # NOTE: This offsetting gives us guarantees of consistency in the (-pi, pi) range.
+        if angle > mn.math.pi:
             angle -= mn.math.pi * 2
-        while angle < -mn.math.pi:
+        elif angle < -mn.math.pi:
             angle += mn.math.pi * 2
+        # NOTE: This final fmod ensures that large angles are mapped back into the -2pi, 2pi range.
+        angle = mn.math.fmod(angle, 2 * mn.math.pi)
         return angle
 
     @base_rot.setter
--- a/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
+++ b/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
@@ -187,7 +187,7 @@ class ArticulatedAgentBase(ArticulatedAgentInterface):
         # NOTE: This offsetting gives us guarantees of consistency in the (-pi, pi) range.
         if angle > mn.math.pi:
             angle -= mn.math.pi * 2
-        elif angle < -mn.math.pi:
+        elif angle <= -mn.math.pi:
             angle += mn.math.pi * 2
         # NOTE: This final fmod ensures that large angles are mapped back into the -2pi, 2pi range.
         angle = mn.math.fmod(angle, 2 * mn.math.pi)
--- a/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
+++ b/habitat-lab/habitat/articulated_agents/articulated_agent_base.py
@@ -176,10 +176,25 @@ class ArticulatedAgentBase(ArticulatedAgentInterface):
 
     @property
     def base_rot(self) -> float:
-        return float(self.sim_obj.rotation.angle())
+        """
+        Returns scalar rotation angle of the agent around the Y axis.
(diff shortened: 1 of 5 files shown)


test/test_humanoid.py
```diff
@@ -202,14 +203,17 @@ def test_humanoid_controller(humanoid_name):
...
         )
         if not osp.exists(humanoid_path):
             pytest.skip(f"No humanoid file {humanoid_path}")
         kin_humanoid = kinematic_humanoid.KinematicHumanoid(agent_config, sim)
         kin_humanoid.reconfigure()
         kin_humanoid.update()
-        assert kin_humanoid.get_robot_sim_id() == 1  # 0 is the ground plane
+        assert (
+            kin_humanoid.get_robot_sim_id() == 2
+        )  # 0 is the stage and 1 is the ground plane
```
0mdc, line 216: We should use constants here for clarity (enum?).

test/test_humanoid.py
```diff
@@ -272,6 +276,89 @@ def test_humanoid_controller(humanoid_name):
...
+@pytest.mark.skipif(
+    not habitat_sim.built_with_bullet,
+    reason="Robot wrapper API requires Bullet physics.",
+)
+def test_base_rot():
+    # set this to output test results as video for easy investigation
+    produce_debug_video = False
+    observations = []
+    cfg_settings = default_sim_settings.copy()
+    cfg_settings["scene"] = "NONE"
```
0mdc, line 292: Good to know :)

test/test_humanoid.py
```diff
@@ -272,6 +276,89 @@ def test_humanoid_controller(humanoid_name):
...
+        if produce_debug_video:
+            from habitat_sim.utils import viz_utils as vut
+
+            vut.make_video(
+                observations,
+                "color_sensor",
+                "color",
+                "test_base_rot",
+                open_vid=True,
+            )
```
0mdc, line 359: Great idea!

habitat-lab/habitat/articulated_agents/articulated_agent_base.py
```diff
@@ -176,10 +176,28 @@ def base_pos(self, position: mn.Vector3):
...
+        if self.sim_obj.rotation.axis()[1] < 0:
+            angle = -1 * angle
+        # NOTE: This offsetting gives us guarantees of consistency in the (-pi, pi) range.
+        if angle > mn.math.pi:
+            angle -= mn.math.pi * 2
+        elif angle <= -mn.math.pi:
+            angle += mn.math.pi * 2
+        # NOTE: This final fmod ensures that large angles are mapped back into the -2pi, 2pi range.
+        angle = mn.math.fmod(angle, 2 * mn.math.pi)
+        return angle
```
0mdc, line 194: In general, it would be more robust to force usage of forward vectors or rotation quaternions.

0mdc (approved):
LGTM! Tested on the HITL stack. Left some general comments.

habitat-lab/habitat/articulated_agents/articulated_agent_base.py
aclegg3: I agree that we are baking in lots of assumptions using a scalar angle about Y axis with implicitly defined origin.
Happy to re-design this part of the system, but pretty deeply embedded, so just trying to get it to behave consistently for now.

PR #2113, opened 2024-12-03
Add convert glb to usd objects.  Issac Lab repo installation required.

danieltmeta:
## Types of changes

## Checklist

### DIFF
--- a/habitat-lab/habitat/isaacsim/remote_content/converter.py
+++ b/habitat-lab/habitat/isaacsim/remote_content/converter.py
@@ -1,0 +1,54 @@
+''' 
+This module takes converts glb files to usd. 
+
+The system that runs this folder needs have the Issac Lab repo installed.
+The converter function from isaac lab can import only after the Isaac Lab stage
+is launched. asyncio library is used because the source code for 
+MeshConverter._convert_mesh_to_usd is also an async function. 
+
+'''
+import asyncio
+import argparse
+from omni.isaac.lab.app import AppLauncher
+import os
+import timer
+
+#Run laucher through Isaac Lab
+
+
+async def convert_glb_to_usd(glb_folder_path: str, usd_folder_path: str) -> None:
+    """
+    This function converts glb files to usd. This function launches an empty Isaac Lab stage,
+    Imports the respective Isaac Lab converter functions, then converts with another 
+    MeshConverter._convert_mesh_to_usd async function. The simulation app then closes, and
+    the converted functions are dumped onto the respective usd_folder_path.
+    """
+
+    # create argparser
+    parser = argparse.ArgumentParser(description="Tutorial on creating an empty stage.")
+    # append AppLauncher cli args
+    AppLauncher.add_app_launcher_args(parser)
+    # parse the arguments
+    args_cli = parser.parse_args()
+    # launch omniverse app
+    app_launcher = AppLauncher(args_cli)
+    simulation_app = app_launcher.app
+
+    from omni.isaac.lab.sim.converters import MeshConverter #NOTE: Import isaac lab converter extension after launcher runs
+
+    glb_folder_list = [glb for glb in os.listdir(glb_folder_path) if os.path.isfile(os.path.join(glb_folder_path, glb))]
+
+    loop = asyncio.get_event_loop()
+    for glb_filename in glb_folder_list:
+        glb_filepath = glb_folder_path + glb_filename
+        usd_filename = f"{glb_filename.replace('.glb', '.usd')}"
+        usd_filepath = usd_folder_path + usd_filename
+        loop.run_until_complete(MeshConverter._convert_mesh_to_usd(in_file=glb_filepath, out_file=usd_filepath))
+    loop.close()
+
+    simulation_app.close()
+
+if __name__ == "__main__":
+    glb_folder_path = '/path/object_glb/'
+    usd_folder_path = '/path/object_usd/'
+    asyncio.run(convert_glb_to_usd(glb_folder_path, usd_folder_path))
--- a/habitat-lab/habitat/isaacsim/ssh_conversion.py
+++ b/habitat-lab/habitat/isaacsim/ssh_conversion.py
@@ -1,0 +1,62 @@
+"""
+This module helps a local computer interact with a remote server.
+The remote server has the machine specs required for Issac Lab installation
+to run MeshConverter._convert_mesh_to_usd(). The local machine does not.
+See this url for documentation:  
+https://isaac-sim.github.io/IsaacLab/main/_modules/omni/isaac/lab/sim/converters/mesh_converter.html#MeshConverter
+"""
+
+from scp import SCPClient
+import os
+import subprocess
+from multiprocessing import Barrier
+
+#NOTE: USER SHOULD INPUT REMOTE SERVER VALUES HERE
+username = 'USERNAME'
+hostname = 'IP_ADDRESS' #NOTE: THIS IS A SECURITY ISSUE
+port = 23456
+REMOTE_GLB_FOLDER = '/path/object_glb/'
+REMOTE_USD_FOLDER = '/path/object_usd'
+
+
+def upload_object_glb_to_server(object_glb_path: str) -> None:
+    """
+    Upload a local folder with .glb files and upload it to remote server. 
+    """
+
+    attempts = 3 #NOTE: The attempts and for loop ensure uploads complete if upload task is dumped from queue.
+    for attempt in range(attempts):
+        try:
+            remote_file_path = REMOTE_GLB_FOLDER + os.path.basename(object_glb_path)
+            scp_command = f'scp {object_glb_path} {hostname}:{remote_file_path}'
+            subprocess.run(scp_command, shell=True)
+            break
+        except Exception as e:
+            print(f"Error uploading {filepath} (attempt {i+1}/{attempts}): {e}")
+            if attempt == attempts - 1:
+                raise
+
+        barrier.wait() 
+
+def run_remote_converter() -> None:
+    """
+    This function runs the glb to usd conversion.py that sits on the given remote computer.
+    This sends bash commands to active Isaac Lab library in conda and run converter.py with a local subprocess library.
+    """
+
+    #Check remote pathing.
+    scp_command = f"ssh {username}@{hostname} 'source /home/guest/miniforge3/etc/profile.d/conda.sh; conda activate isaaclab2; python /home/guest/dev/isaac_lab_converter/converter.py' "
+    subprocess.run(scp_command, shell=True)
+
+
+def download_remote_usd_folder(local_download_file_path: str) -> None:
+    """
+    This function downloads folder contents from a remote computer.
+    This function uses the bash script with the help of the subprocess library.
+    """
+    scp_command = f'scp -r -C {hostname}:{REMOTE_USD_FOLDER} {local_download_file_path}'
+    subprocess.run(scp_command, shell=True)
+
+
+if __name__ == "__main__":
+    run_remote_converter()
--- a/habitat-lab/habitat/isaacsim/usd_conversion_scene.py
+++ b/habitat-lab/habitat/isaacsim/usd_conversion_scene.py
@@ -1,0 +1,296 @@
+"""This file converts a <EPISODE>.scene_instance.json to a .usd file, 
+using various .glb scene objects to buld the .usd scene instance..
+"""
+
+from pxr import Usd, UsdGeom, Sdf, Gf
+import json
+from pathlib import Path
+import os
+from ssh_conversion import *
+from multiprocessing import Pool, Barrier
+from typing import List, Dict, Union, Tuple
+
+
+# CONSTANTS
+STAGE_INSTANCE = "stage_instance"
+OBJECT_INSTANCES = "object_instances"
+TEMPLATE_NAME = "template_name"
+TRANSLATION = "translation"
+ROTATION = "rotation"
+SCALE = "non_uniform_scale"
+
+
+def find_filepath(filename, parent_directory) -> str:
+    file_found = False
+    for root, dirs, files in os.walk(parent_directory):
+        if filename in files:
+            output_filepath = str(os.path.join(root, filename))
+            file_found = True
+            break
+
+    if not file_found:
+        print(f"{filename} not found in the parent directoy {parent_directory}")
+        return None
+
+    return output_filepath
+
+
+def get_unique_object_names(scene_data) -> list[str]:
+    """
+    This looks at the scene_instance.json by looking at the "template_name" key.
+    HSSD keeps all these objects as .glb. The scene json could use the same type of
+    object multiple times, but we only need the unique list of objects
+    """
+    object_list = scene_data[OBJECT_INSTANCES]  
+
+    object_name_list = []
+    for object in object_list:
+        object_name_list.append(object[TEMPLATE_NAME])
+
+    return list(set(object_name_list))
+
+
+def make_object_filepath_list(
+    glb_data_folder: str,
+    scene_instance_filepath: str,
+    scene_data,   #NOTE: This should be typing for scene_instance.json Dict[str, Union[str, Dict[str, str] , List[ Dict[str, List[float]] ]]],
+    file_extension: str = ".glb",
+) -> List[str]:
+
+    object_name_list = get_unique_object_names(scene_data)
+
+    # Search through the object glb folder to make a list of object filepaths from scene_data
+    object_filepath_list = []
+    for object_name in object_name_list:
+        object_glb = object_name + file_extension
+        object_path = find_filepath(object_glb, glb_data_folder)
+        if object_path:
+            object_filepath_list.append(object_path)
+
+    if not object_filepath_list:
+        # Stop the entire script if no .glb objects were found
+        print("No .glb objects found. Halting the program.")
+        sys.exit(1)
+
+    return object_filepath_list
+
+
+def make_collider_glb_filepath(object_name: str, scene_instance_path: str) -> str:
+
+    object_collider_glb = object_name + ".collider.glb"
+    scene_folder_path = os.path.dirname(os.path.abspath(scene_instance_path))
+    hssd_folder_path = os.path.dirname(scene_folder_path)
+    object_folder_path = hssd_folder_path + "/objects"
+
+    object_usd_path = object_folder_path + "/" + object_name + ".collider.usd"
+    return object_usd_path
+
+
+def initialize_stage(stage_usd_filepath: str, scene_instance_usd_filepath: str) -> Usd.Stage:
+    """Initialize Usd.Stage"""
+
+
+    stage = Usd.Stage.CreateNew(scene_instance_usd_filepath) # NOTE Might not work starting with nummber
+    xform_prim = UsdGeom.Xform.Define(stage, "/Scene")
+
+    #probably don't need this
+    #prim = xform_prim.GetPrim()
+    #prim.GetReferences().AddReference(stage_usd_filepath)
+
+    return stage
+
+
+def find_object_usd(object_data, object_usd_folder: str, file_extension=".collider.usd") -> Tuple[bool, str, str]:
+    """
+    Find the object .usd file from the object usd folder.
+    """
+    object_exists = True
+
+    object_name = object_data[TEMPLATE_NAME]
+    usd_filename = object_name + file_extension
+    usd_filepath = object_usd_folder + usd_filename
+
+    if not os.path.exists(usd_filepath):
+        print(f"File {usd_filepath} not found. .usd file may not have been coverted or downloaded")
+        object_exists = False
+
+    return object_exists, usd_filepath, object_name
+
+
+def add_object_to_stage(
+    stage: Usd.Stage,
+    object_data,
+    object_filepath: str,
+    count: str,
+    position: List[float]=[0.0, 0.0, 0.0],
+    rotation: List[float]=[0.0, 0.0, 0.0],
+    scale: List[float]=[1.0, 1.0, 1.0],
+) -> None:
+    """
+    This module adds an UsdGeom.Xform to an initialized Usd.Stage.
+    """
+    position = object_data[TRANSLATION]
+    rotation = object_data[ROTATION]
+    scale = object_data[SCALE]
+    object_name = object_data[TEMPLATE_NAME]
+
+    # PSEUDOCODE # Given template name from scene_instance.json from above
+    # From root: /home/trandaniel/dev/habitat-sim/data/hssd-hab/scenes/<MY_SCENE>.scene_instance.json
+    # Object folder: /home/trandaniel/dev/habitat-sim/data/hssd-hab/objects
+
+    # Navigate from scene instance folder to object folder through script
+    #
+
+    # Recursively search through object folder for object json (Structure vary between object data sets. Write for HSSD for now)
+    # Open JSON, and then use the "render_asset" key and "collision_asset" key to get .glb file
+
+    # Convert .collision.glb to USD with isaacsim omniverse
+    # Output .usd file in the object folder that contains collision.glb
+
+    # CONVERSION
+
+    # Turn object .glb to xform
+
+    # Add count to string if object used more than once
+    if count > 0:
+        object_name = object_name + f"_{count}"
+
+    object_stage_name = "/OBJECT_" + object_name # NOTE: Object state name must start with a char letter
+
+    object_xform = UsdGeom.Xform.Define(
+        stage, object_stage_name
+    )  
+    object_xform.GetPrim().GetReferences().AddReference(object_filepath)
+
+    # Set Position
+    translate_op = next(
+        (
+            op
+            for op in object_xform.GetOrderedXformOps()
+            if op.GetName() == "xformOp:translate"
+        ),
+        None,
+    )
+    if translate_op is None:
+        translate_op = object_xform.AddTranslateOp()
+    translate_op.Set(value=Gf.Vec3d(*position))
+
+    # Set Rotation
+    orient_op = next(
+        (
+            op
+            for op in object_xform.GetOrderedXformOps()
+            if op.GetName() == "xformOp:orient"
+        ),
+        None,
+    )
+    if orient_op is None:
+        orient_op = object_xform.AddOrientOp()
+    orient_op.Set(value=Gf.Quatd(*rotation))
+
+    # Set Scale
+    scale_op = next(
+        (
+            op
+            for op in object_xform.GetOrderedXformOps()
+            if op.GetName() == "xformOp:scale"
+        ),
+        None,
+    )
+    if scale_op is None:
+        scale_op = object_xform.AddScaleOp()
+    scale_op.Set(value=Gf.Vec3d(*scale))
+
+
+def convert_scene_instance(
+    scene_instance_filepath: str, stage_glb_folder: str, local_download_folder: str, glb_data_folder: str
+) -> None:
+    """
+    This function takes a folder of .glb objects, converts them remotely to .usd,
+    downloads the converted files, and then makes a .usda scene. The idea is to
+    convert a scene_instance json to a .usd file by aggregating all the defined 
+    objects within the scene instance json.
+    """
+    
+    # Open scnece_instance JSON
+    with open(scene_instance_filepath, "r") as file:
+        scene_data = json.load(file)
+ 
+    
+    stage_name: str = scene_data[STAGE_INSTANCE][TEMPLATE_NAME].split('/')[1
+    ]
+    
+    # Form a list of existing object glb file paths 
+    object_glb_filepath_list = make_object_filepath_list(
+        glb_data_folder,
+        scene_instance_filepath,
+        scene_data,
+        file_extension=".collider.glb",
+    )
+
+    # # Append stage glb to piggy back on conversion.py. #Note, stage glb not necessary probably
+    #stage_glb_filepath = stage_glb_folder + stage_name + '.glb'
+    #object_glb_filepath_list.append(stage_glb_filepath)
+
+    # Upload glb objects to remote server
+    num_processes = 10
+    barrier = Barrier(num_processes)
+    with Pool(processes=num_processes) as pool:
+        pool.map(upload_object_glb_to_server, object_glb_filepath_list)
+
+    # Process objects remotely
+    run_remote_converter()
+
+    # Download objects
+    download_remote_usd_folder(local_download_folder)
+
+    
+    # TODO: Delete objects on remote server after download
+
+    # Populate USDA file. USDA is the human readable version of a usd file.\
+
+    stage_usd_filepath = local_download_folder + 'object_usd/' + stage_name + '.usd' # NOTE: Might not need this
+    scene_instance_usd_filepath = './' + stage_name + '.scene_instance.usda'
+
+    stage = initialize_stage(stage_usd_filepath, scene_instance_usd_filepath)
+
+    object_usd_folder = local_download_folder + "/object_usd/" #The string is the name of the folder from remote server
+    
+    # This is to ensure the same object used multiple times have a unique ID number
+    previous_object_name = "" 
+    count = 0
+
+    for object_data in scene_data[OBJECT_INSTANCES]:
+        object_exists, usd_filepath, object_name = find_object_usd(
+            object_data, object_usd_folder, file_extension=".collider.usd"
+        )
+
+        # This if/else is to ensure multiple copies of the same type of object have unique names.
+        if previous_object_name == object_name:
+            count += 1
+        else:
+            previous_object_name = object_name
+            count = 0
+
+        if object_exists:
+            add_object_to_stage(stage, object_data, usd_filepath, count)
+
+    stage.GetRootLayer().Save()
+
+    print(f"Finished rendering convertin {scene_instance_filepath} to .usd")
+
+
+if __name__ == "__main__":
+
+    scene_instance_filepath = "/home/trandaniel/dev/habitat-sim/data/hssd-hab/scenes/102343992.scene_instance.json"
+    stage_glb_folder = '/home/trandaniel/dev/IsaacSimConversion/'
+    local_download_folder = "/home/trandaniel/dev/IsaacSimConversion/usd_scene/"
+    glb_data_folder = "/home/trandaniel/dev/habitat-sim/data/hssd-hab/objects"
+
+    convert_scene_instance(
+        scene_instance_filepath, stage_glb_folder, local_download_folder, glb_data_folder
+    )
+
+    # TODO: Test with /home/trandaniel/dev/habitat-sim/data/hssd-hab/scenes/<MY_SCENE>.scene_instance.json
+    # TODO; Open resulting scene output with blender, open it in viewer.py
+    # Some things might be missing, only rigid objects render for now, articulated later


habitat-lab/habitat/isaacsim/remote_content/converter.py
```diff
@@ -0,0 +1,54 @@
...
+    from omni.isaac.lab.sim.converters import MeshConverter #NOTE: Import isaac lab converter extension after launcher runs
+
+    glb_folder_list = [glb for glb in os.listdir(glb_folder_path) if os.path.isfile(os.path.join(glb_folder_path, glb))]
+
+    loop = asyncio.get_event_loop()
+    for glb_filename in glb_folder_list:
+        glb_filepath = glb_folder_path + glb_filename
+        usd_filename = f"{glb_filename.replace('.glb', '.usd')}"
+        usd_filepath = usd_folder_path + usd_filename
+        loop.run_until_complete(MeshConverter._convert_mesh_to_usd(in_file=glb_filepath, out_file=usd_filepath))
```
eundersander, line 46: `_convert_mesh_to_usd` is private as indicated by the leading underscore, so we shouldn't be calling it directly. Can you try something like this?
```
        print(f"Converting {collision_asset_filepath}...")
        converter_config = MeshConverterCfg(
            asset_path=collision_asset_filepath,
            usd_dir="./data/usd/objects/",
            usd_file_name=f"{object_config_filename}.usda",
            force_usd_conversion=False,
            collision_approximation="convexDecomposition",
        )
        converter = MeshConverter(converter_config)
        print(f"Wrote {converter.usd_path}")
```

eundersander: Assuming the above API works, you shouldn't need to do anything with asyncio to use MeshConverter.

habitat-lab/habitat/isaacsim/remote_content/object_glb/.gitignore
eundersander, line 1: We shouldn't need this file or this folder in source control.

habitat-lab/habitat/isaacsim/remote_content/object_usd/.gitignore
eundersander, line 1: We shouldn't need this file or this folder in source control.

habitat-lab/habitat/isaacsim/ssh_conversion.py
```diff
@@ -0,0 +1,62 @@
+"""
```
eundersander, line 1: This looks like a nice utility for your SSH-based workflow, but this doesn't need to be added to the habitat-lab repo. Most folks won't be using this workflow. This file shouldn't be part of this PR.

habitat-lab/habitat/isaacsim/urdf_conversion.py
```diff
@@ -0,0 +1,48 @@
...
+    # parse the arguments
+    args_cli = parser.parse_args()
+    # launch omniverse app
+    app_launcher = AppLauncher(args_cli)
+    simulation_app = app_launcher.app
+
+    #NOTE: Import isaac lab converter extension after launcher runs
+    from omni.isaac.lab.sim.converters import UrdfConverter, UrdfConverterCfg
+    
+    # urdf_filename = 'sink.urdf'
```
eundersander, line 30: Commented-out test code shouldn't merge to `main`.

habitat-lab/habitat/isaacsim/urdf_conversion.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Add config information
+    config = UrdfConverterCfg()
+    config.asset_path = source_urdf_filepath # Source file
+    config.usd_dir = usd_folder_path # Output file
+    config.fix_base = True
+
+    # Convert asset
+    converter = UrdfConverter(cfg=config)
+    converter._convert_asset(cfg=config)
```
eundersander, line 41: We shouldn't call `_convert_asset`. This is a private method as indicated by the leading underscore. Something like this worked for me:
```
def convert_urdf(urdf_filepath):

    from omni.isaac.lab.sim.converters import UrdfConverter, UrdfConverterCfg

    # Define the configuration for the URDF conversion
    config = UrdfConverterCfg(
        asset_path=urdf_filepath,  # Path to the input URDF file
        usd_dir="./data/usd/",         # Directory to save the converted USD file
        usd_file_name="converted_robot.usda",         # Name of the output USD file
        force_usd_conversion=True,                   # Force conversion even if USD already exists
        make_instanceable=True,                      # Make the USD asset instanceable
        link_density=1000.0,                         # Default density for links
        import_inertia_tensor=True,                  # Import inertia tensor from URDF
        convex_decompose_mesh=False,                 # Decompose convex meshes
        fix_base=False,                              # Fix the base link
        merge_fixed_joints=False,                    # Merge links connected by fixed joints
        self_collision=False,                        # Enable self-collision
        default_drive_type="position",               # Default drive type for joints
        override_joint_dynamics=False                # Override joint dynamics
    )

    # Create the UrdfConverter with the specified configuration
    converter = UrdfConverter(config)

    # Get the path to the generated USD file
    usd_path = converter.usd_path
    print(f"USD file generated at: {usd_path}")  
````

File: habitat-lab/habitat/isaacsim/remote_content/converter.py [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+import asyncio
+import argparse
+from omni.isaac.lab.app import AppLauncher
+import os
+import timer
+
+#Run laucher through Isaac Lab
+
+
+async def convert_glb_to_usd(glb_folder_path: str, usd_folder_path: str) -> None:
```
eundersander, line 19: Let's not create AppLauncher inside the function, because we'll likely call this function several times inside a larger application in the future. You can create the AppLauncher in your `if __name__ == "__main__":` block so that the script still works standalone.

habitat-lab/habitat/isaacsim/urdf_conversion.py
```diff
@@ -0,0 +1,48 @@
...
+    This function converts a .urdf file into a .usd file. NOTE: Please be aware of 
+    pathing issues and references to other .glb files within the urdf file.
+    
+    :param source_urdf_filepath: The absolute string path of the urdf file
+    :param usd_folder_path: The absolute directory of the desired directory location of .usd file. The output will generatate a Props-> instanceable_meshes.usd, .asset_hash, config.yaml, and the desired .usd.
+
+    """
+
+    # create argparser
+    parser = argparse.ArgumentParser(description="Create an empty Issac Sim stage.")
```
eundersander, line 18: Let's not create AppLauncher inside the function, because we'll likely call this function several times inside a larger application in the future. You can create the AppLauncher in your `if __name__ == "__main__":` block so that the script still works standalone.

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
```diff
@@ -0,0 +1,296 @@
...
+    scene_data,   #NOTE: This should be typing for scene_instance.json Dict[str, Union[str, Dict[str, str] , List[ Dict[str, List[float]] ]]],
+    file_extension: str = ".glb",
+) -> List[str]:
+
+    object_name_list = get_unique_object_names(scene_data)
+
+    # Search through the object glb folder to make a list of object filepaths from scene_data
+    object_filepath_list = []
+    for object_name in object_name_list:
+        object_glb = object_name + file_extension
```
eundersander, line 65: This is not the right way to find the collision GLB for an object. You need to find the `.object_config.json` file, open that, and extract the `collision_asset` string, which is a filepath to a GLB. We talked about this earlier, and in fact it is mentioned in your pseudocode below around line 144. Here's a half-finished version of scene-instance conversion that I started writing (untested):
```
def find_file(folder, filename):
    result = None
    for root, dirs, files in os.walk(folder):
        if filename in files:
            assert not result
            result = os.path.join(root, filename)
    return result

def convert_hab_scene(scene_filepath):

    from omni.isaac.lab.sim.converters import MeshConverterCfg, MeshConverter

    scenes_folder = os.path.dirname(scene_filepath)
    objects_folder = scenes_folder + "/../objects"
    assert os.path.exists(objects_folder) and os.path.isdir(objects_folder)

    temp_symlink_filename = None # todo "./data/usd/temp_symlink.
    assert False

    import json
    with open(scene_filepath, 'r') as file:
        scene_json_data = json.load(file)    

    assert "object_instances" in scene_json_data
    for obj_instance_json in scene_json_data["object_instances"]:

        object_config_filename = f"{obj_instance_json['template_name']}.object_config.json"
        object_config_filepath = find_file(objects_folder, object_config_filename)
        assert object_config_filepath

        object_config_json_data = None
        with open(object_config_filepath, 'r') as file:
            object_config_json_data = json.load(file)  
        collision_asset_filename = object_config_json_data["collision_asset"]
        collision_asset_filepath = find_file(objects_folder, collision_asset_filename)

        print(f"Converting {collision_asset_filepath}...")
        converter_config = MeshConverterCfg(
            asset_path=collision_asset_filepath,
            usd_dir="./data/usd/objects/",
            usd_file_name=f"{object_config_filename}.usda",
            force_usd_conversion=False,
            collision_approximation="convexDecomposition",
        )
        converter = MeshConverter(converter_config)
        print(f"Wrote {converter.usd_path}")
```

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
```diff
@@ -0,0 +1,296 @@
...
+
+    # # Append stage glb to piggy back on conversion.py. #Note, stage glb not necessary probably
+    #stage_glb_filepath = stage_glb_folder + stage_name + '.glb'
+    #object_glb_filepath_list.append(stage_glb_filepath)
+
+    # Upload glb objects to remote server
+    num_processes = 10
+    barrier = Barrier(num_processes)
+    with Pool(processes=num_processes) as pool:
+        pool.map(upload_object_glb_to_server, object_glb_filepath_list)
```
eundersander, line 239: Let's not use SSH for anything in these scene-conversion scripts. I understand you are using SSH for your development, but in practice, this script is going to run locally on a machine on which HSSD and other assets are present, and the converted USD files should be written to the local filesystem.

I suspect the reason you took this approach is that you didn't have a practical way to upload HSSD to my Lambda machine? I have it on that machine, just not under the guest account. I can add a copy to the guest account so you can do all your conversion testing remotely on the Lambda machine, without having to upload or download any files to/from your local machine. Let's chat more if you're blocked here!

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
```diff
@@ -0,0 +1,296 @@
...
+            add_object_to_stage(stage, object_data, usd_filepath, count)
+
+    stage.GetRootLayer().Save()
+
+    print(f"Finished rendering convertin {scene_instance_filepath} to .usd")
+
+
+if __name__ == "__main__":
+
+    scene_instance_filepath = "/home/trandaniel/dev/habitat-sim/data/hssd-hab/scenes/102343992.scene_instance.json"
```
eundersander, line 285: This kind of test code with hard-coded paths shouldn't be in this file.

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
```diff
@@ -0,0 +1,296 @@
...
+    stage_glb_folder = '/home/trandaniel/dev/IsaacSimConversion/'
+    local_download_folder = "/home/trandaniel/dev/IsaacSimConversion/usd_scene/"
+    glb_data_folder = "/home/trandaniel/dev/habitat-sim/data/hssd-hab/objects"
+
+    convert_scene_instance(
```
eundersander, line 290: Let's add `test/test_isaac_sim.py` to habitat-lab in this PR. Look at our other test files in that folder for reference. Let's have a test that tries to convert a toy habitat scene instance (this is something you suggested a while back). The test can assume the source assets are present on the machine (or maybe check for them and skip the test if they're missing). The test should inspect the output `usda` file and check for expected contents.

eundersander (commented):
Some good progress here! I left some comments.

My biggest request is to add a test file (the last comment in `usd_conversion_scene.py`).

Also, I see that pre-commit and python lint CI tests are failing (see below "Some checks are failing"). Try to get those passing. Have you installed pre-commit locally? See [instructions](https://github.com/facebookresearch/habitat-lab/blob/main/CONTRIBUTING.md#coding-style).

Let me know when you've addressed my comments and I'll take another look!

habitat-lab/habitat/isaacsim/remote_content/converter.py
danieltmeta: Asyncio library removed and MeshCoverterCfg added. Private functions removed and committed.

habitat-lab/habitat/isaacsim/remote_content/object_glb/.gitignore
danieltmeta: remote_content folder removed.

habitat-lab/habitat/isaacsim/urdf_conversion.py
danieltmeta: Extraneous comment removed.

habitat-lab/habitat/isaacsim/urdf_conversion.py
danieltmeta: Private function removed. UrdfConverterCfg added.

habitat-lab/habitat/isaacsim/remote_content/converter.py
danieltmeta: Applauncher class moved.

habitat-lab/habitat/isaacsim/urdf_conversion.py
danieltmeta: Applauncher class moved.

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
danieltmeta: My initial draft was to make sure the individual classes imported from isaac lab worked. But you are right, the comments near line 144 were notes from our earlier meeting. The current draft was merely for expediency. Will update this file according to this draft code with object_config.json.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+        assert os.path.exists(objects_folder)
+    except AssertionError:
+        raise FileNotFoundError(
+            f"Object glb folder {objects_folder} does not exist"
+        )
+
+    # A scene may have multiple instances of the same object mesh. Xform usda needs
+    # to have one unique name per mesh.
+    object_counts: Dict[str, int] = {}
+    max_count = -1  # 50  #TODO: temp only convert the first N objects
```
eundersander, line 471: This is debug code; we don't need to merge this.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+        (
+            object_config_filename,
+            out_usd_path,
+            object_counts,
+            unique_object_name,
+        ) = object_usd_info(obj_instance_json, object_counts)
+
+        # TODO: gather these up and do them later with multiprocessing
+        # NOTE: This will be answered when trying to proccess multiple scene_instance_jsons.
+        # if not os.path.exists(out_usd_path):
```
eundersander, line 505: We don't need to merge old commented-out code like this.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+    project_root_folder: str,
+    objects_folder: str = "",
+    scene_usd_filepath: str = "./data/test_scene_instance.usda",
+) -> None:
+    """
+    This is the main function takes data from a scene_instance.json file to .usda.
+    NOTE, folder directories for objects are speific to the hssd object dataset.
+    """
+
+    # TODO: DANIEL Not sure why check is here, when json library for scene_json_data variable does same check?
```
eundersander, line 438: Resolve this TODO before merging. It sounds like this path.exists check is redundant and can be removed.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+def convert_hab_scene(
+    scene_filepath: str,
+    project_root_folder: str,
+    objects_folder: str = "",
+    scene_usd_filepath: str = "./data/test_scene_instance.usda",
+) -> None:
+    """
+    This is the main function takes data from a scene_instance.json file to .usda.
```
eundersander, line 434: Let's give a proper docstring here to document this function. Also, "This is the main function takes data from a scene_instance.json file to .usda." isn't very clear and the grammar is sloppy.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    Convert a position from Habitat (Y-up) to USD (Z-up) coordinate system.
+
+    Habitat (-x, z, y) -> Isaac (x, y, z)
+    """
+    x, y, z = position
+    return [-x, z, y]
+
+
+############################################################################################
```
eundersander, line 422: This decoration doesn't really fit our style and seems unnecessary.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+    asset_scale_attr.Set(Gf.Vec3f(*render_asset_scale))
+
+    # Save the updated USD file
+    stage.GetRootLayer().Save()
+    print(
+        f"Added habitat visual metadata to RigidBody prim in: {usd_filepath}"
+    )
+
+
+############################################################################################
```
eundersander, line 296: Again, this decoration looks strange and unnecessary to me.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+            # set purpose to "guide" so that this mesh doesn't render in Isaac?
+            if True:
+                mesh_geom = UsdGeom.Imageable(prim)
+                mesh_geom.CreatePurposeAttr(UsdGeom.Tokens.guide)
+
+
+def add_habitat_visual_to_usd_root(
+    usd_filepath, render_asset_filepath, render_asset_scale
+):
+    """TODO: Add function description, param description, and output description"""
```
eundersander, line 268: Let's resolve this todo before merging.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+    convert_meshes_to_static_colliders(stage, "/World")
+
+    stage.GetRootLayer().Export(out_file)
+
+
+async def async_convert_mesh_to_usd(
+    in_file: str, out_file: str, load_materials: bool = True
+) -> bool:
+    """ported from IsaacLab mesh_converter.py _convert_mesh_to_usd
+    TODO: Add param descriptions, and return description.
```
eundersander, line 188: resolve TODO before merging

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+def convert_mesh_to_usd(
+    in_file: str, out_file: str, load_materials: bool = True
+) -> None:
+    """Convert mesh to usd
+
+    :param in_file: string filepath of input mesh
+    :param out_file: string of output usda file
+    :param load_materials: TODO: Add description
+
+    :return: TODO: Add description
```
eundersander, line 166: resolve TODO before merging

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+############################################################################################
+
+
+def convert_object_to_usd(
+    objects_folder: str,
+    object_config_filename: str,
+    out_usd_path: str,
+    project_root_folder: str,
+) -> None:
+    """This converts an hssd object to a usda file.
```
eundersander, line 88: We shouldn't be specific to HSSD or name HSSD. (Although I'm okay with parameter default values being HSSD-centric.)

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+        if "collision_asset" in object_config_json_data
+        else object_config_json_data["render_asset"]
+    )
+    collision_asset_filepath = find_file(
+        objects_folder, collision_asset_filename
+    )
+
+    print(f"Converting {collision_asset_filepath}...")
+
+    convert_mesh_to_usd(
```
eundersander, line 113: We should have logic to skip converting the mesh if the output file already exists.

habitat-lab/habitat/isaacsim/urdf_to_usd.py
```diff
@@ -0,0 +1,209 @@
...
+            asset_scale_attr.Set(Gf.Vec3f(*metadata["assetScale"]))
+        else:
+            print(f"Warning: Prim not found for link: {link_name}")
+
+    # Save the updated USD to the output file
+    stage.GetRootLayer().Export(out_usd_filepath)
+    print(f"Updated USD file written to: {out_usd_filepath}")
+
+
+def convert_urdf_test():
```
eundersander, line 168: This test doesn't belong in this file.

habitat-lab/habitat/isaacsim/urdf_to_usd.py
```diff
@@ -0,0 +1,209 @@
...
+    convert_urdf(clean_urdf_filepath, temp_usd_filepath)
+    add_habitat_visual_metadata_for_articulation(
+        temp_usd_filepath,
+        source_urdf_filepath,
+        out_usd_filepath,
+        project_root_folder="./",
+    )
+
+
+if __name__ == "__main__":
```
eundersander, line 188: Let's clean up the main for this script by making it a CLI to convert a URDF (including cleaning it).

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+
+        count += 1
+        if count == max_count:
+            break
+
+    stage.GetRootLayer().Save()
+    print(f"wrote scene {scene_usd_filepath}")
+
+
+if __name__ == "__main__":
```
eundersander, line 539: Let's clean up this main by giving this file a CLI to convert a single scene instance file.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
```diff
@@ -0,0 +1,562 @@
...
+    scene_instance_filepath = "/home/trandaniel/dev/habitat-sim/data/hssd-hab/scenes/102343992.scene_instance.json"
```
eundersander, line 540: In general, we never want to merge hard-coded filepaths that won't make sense for all users.

eundersander (changes requested):
I left some comments. A few top-level comments:
* Let's add nice CLIs for both of these files. This means using argument-parsing and offering a friendly command-line interface.
* You need to fill out the PR description. What's autopopulated there is a template for you to edit. This is an important part of documenting a code change.
* As I requested earlier, let's add a minimal test file to this PR.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: max_count removed.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Deleted.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Comment removed along with assert statement.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Removed.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: parameters added.  Doctring updated.  Resolved.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Added if condition prior to convert_mesh_to_usd

habitat-lab/habitat/isaacsim/urdf_to_usd.py
danieltmeta: Added CLI

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Remove hard-coded filepaths.

habitat-lab/habitat/isaacsim/urdf_to_usd.py
danieltmeta: Removed

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
danieltmeta: Added CLI

habitat-lab/habitat/isaacsim/usd_conversion_scene.py
danieltmeta: Migrating unit tests and example data from [PR ](https://github.com/facebookresearch/habitat-lab/pull/2123)

danieltmeta: Need to delete PR #2123 after unit tests are reviewed.

test/test_isaac_sim.py
```diff
@@ -0,0 +1,317 @@
...
+                    visual_metadata[safe_link_name] = {
+                        "assetPath": asset_path,
+                        "assetScale": scale,
+                    }
+
+    # Open the USD file
+    stage_before_metadata = Usd.Stage.Open(converted_clean_usda)
+    stage_output_metadata = Usd.Stage.Open(out_usd_filepath)
+
+    # Add the habitatVisual metadata to each relevant prim
```
eundersander, line 299: This comment looks wrong.

test/test_isaac_sim.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def test_add_habitat_visual_metadata_for_articulation():
+    urdf_dir = repo_dir + "/test/data/usd_conversion_data/"
+    converted_clean_usda = urdf_dir + "hab_spot_arm_test_convert_urdf.usda"
+    reference_urdf_filepath = urdf_dir + "hab_spot_arm_EXAMPLE.urdf"
+    out_usd_filepath = urdf_dir + "hab_spot_arm_with_hab_metadata.usda"
+    project_root_folder = repo_dir
+
+    subprocess.run(
```
eundersander, line 242: Your test shouldn't run the script in a subprocess. Just import and run the required functionality directly.

test/test_isaac_sim.py
```diff
@@ -0,0 +1,317 @@
...
+
+def test_convert_urdf():
+    urdf_dir = repo_dir + "/test/data/usd_conversion_data/"
+    clean_urdf_filepath = urdf_dir + "hab_spot_arm_clean_CORRECT.urdf"
+    output_usd = urdf_dir + "hab_spot_arm_test_convert_urdf.usda"
+
+    convert_urdf(clean_urdf_filepath, output_usd)
+    # TODO: Issac sim drops a prim. "Warning: Prim not found for link: fl_uleg"
+
+    assert True
```
eundersander, line 232: You should do a real test here that checks something. At a minimum, check that the output file exists.

test/data/usd_conversion_data/config.yaml
```diff
@@ -0,0 +1,18 @@
+asset_path: /home/trandaniel/dev/habitat-lab/test/data/usd_conversion_data/hab_spot_arm_clean_CORRECT.urdf
```
eundersander, line 1: Is this needed in this PR? It looks like debug output from the conversion process.

test/test_isaac_sim.py
```diff
@@ -0,0 +1,317 @@
...
+    # -1 values?  Don't know what hab space scaling is.
+
+
+## urdf_to_usd.py unit tests
+
+
+def test_clean_urdf():
+    urdf_dir = repo_dir + "/test/data/usd_conversion_data/"
+    input_file = urdf_dir + "hab_spot_arm_EXAMPLE.urdf"
+    output_file = urdf_dir + "hab_spot_arm_clean.urdf"
```
eundersander, line 180: If hab_spot_arm_clean.urdf is generated by the test, then I guess hab_spot_arm_clean.urdf doesn't need to be included in this PR. It isn't a source file needed to run the test.

habitat-lab/habitat/isaacsim/scene_instance_json_to_usd.py
eundersander: Just bumping this comment. Let's not name HSSD in comments.

eundersander (changes requested):
Good progress here!

Please flesh out your PR description a bit. You don't have to explain what Isaac Sim is, but give some background for why we need these conversion utils. Mention that we convert scene instances including rigid objects (but not any included articulated objects or the stage yet). Mention the URDF conversion for converting robot URDF files. For Type of change, specify [Development]. 

Let's move `meshColored` inside `data/usd_conversion_data` so all the isaac-related test data is in one subfolder.

It looks like maybe you're not using `test/data/usd_conversion_data/102343992.scene_instance.json` for testing, so you can remove it from this PR.

I suspect you don't need to add `test/data/usd_conversion_data/objects_EXAMPLE2/1efdc3d37dfab1eb9f99117bb84c59003d684811.glb` to this repo, because it's a render GLB and it won't be used/loaded during the conversion (or am I missing something?). In general, we want to be really careful and selective about adding binary files to a git repo, since they are usually large.

test/data/usd_conversion_data/.asset_hash
```diff
@@ -0,0 +1 @@
+24f1b688baf20921994b95e47f42b7ee
```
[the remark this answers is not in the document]
danieltmeta: That is a result of a .urdf conversion. I will use .gitignore so this doesn't get pushed

habitat-lab/habitat/isaacsim/urdf_to_usd.py
```diff
@@ -0,0 +1,245 @@
+"""This module converts urdf files into usda files."""
```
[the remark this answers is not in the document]
danieltmeta: Files moved

test/test_isaac_sim.py
danieltmeta: Comment fixed. The for loop after this line is checking if the habitatVisual data was saved onto the respective usda file being unit tested.

test/data/usd_conversion_data/config.yaml
danieltmeta: This is an artifact of the clean_urdf function. I will put a .gitignore so this doesn't get pushed.

test/test_isaac_sim.py
danieltmeta: I wasn't sure what to test here because convert_urdf uses the UrdfConverter class from issac lab.  I guess I can use getters to grab class property values?  I can add file output existence.

test/test_isaac_sim.py
danieltmeta: I can't run the conversion files as imports because assuming you run a unit test file, the other conversion files cannot use the Issac Lab USD library imported inside the unit test file. Instead, you have to run another instance of Applauncher on the conversion file.

An alternative to subprocess would be running the below, but at this point I don't know how to run the conversion file from Applauncher environment within the unit test file.

```
command = f"python script.py {arg1} {arg2}"
os.system(command)
```

test/test_isaac_sim.py
danieltmeta: I adjusted the code to remove output files generated by the unit test

test/test_isaac_sim.py
danieltmeta: Added test to check output exists.

test/test_isaac_sim.py
danieltmeta: Subprocess removed.  Imports need to added to external functions to solve the issue.

.gitignore
```diff
@@ -106,3 +106,11 @@ data
...
 # Autogenerated navmesh files
 *.navmesh
+
+# urdf conversion artifact
+.asset_hash
+test/data/usd_conversion_data/config.yaml
+
+# Dependency Repos
+IssacLab/
+habitat-sim/
```
eundersander, line 116: This isn't needed. `.gitignore` entries are generally for files that many or most folks will encounter, where we know 100% it shouldn't be added to source control. Folks generally aren't cloning habitat-sim at the root of habitat-lab. I recommend cloning habitat-sim outside habitat-lab. If you happen to have this cloned here on your local system and you're trying to clean up your branch, just leave this folder as un-staged. No need to add it to .gitignore.

.gitignore
```diff
@@ -106,3 +106,11 @@ data
 
 # Autogenerated navmesh files
 *.navmesh
+
+# urdf conversion artifact
+.asset_hash
+test/data/usd_conversion_data/config.yaml
+
+# Dependency Repos
+IssacLab/
```
eundersander, line 115: Likewise, this isn't needed. Folks generally shouldn't be cloning IsaacLab at the root of habitat-lab.

.gitignore
```diff
@@ -106,3 +106,11 @@ data
 
 # Autogenerated navmesh files
 *.navmesh
+
+# urdf conversion artifact
+.asset_hash
```
eundersander, line 111: Can you qualify this as test/data/usd_conversion_data/.asset_hash?

test/data/usd_conversion_data/config.yaml
```diff
@@ -14,5 +14,5 @@ default_drive_damping: 0.0
 link_density: 1000.0
```
eundersander, line 14: This file looks like output from your test. It doesn't need to be added to version control.

habitat-lab/habitat/isaacsim/urdf_to_usd.py
eundersander: I'd still like to see the CLI for `urdf_to_usd.py` simplified as described [in this comment](https://github.com/facebookresearch/habitat-lab/pull/2113#discussion_r1920550571). However, if you're eager to wrap up this PR, you can add a todo in the code here and we'll address this later.

test/test_isaac_sim.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+def test_convert_urdf():
+    urdf_dir = repo_dir + "/test/data/usd_conversion_data/"
+    clean_urdf_filepath = urdf_dir + "hab_spot_arm_clean_CORRECT.urdf"
+    output_usd = urdf_dir + "hab_spot_arm_test_convert_urdf.usda"
+
+    convert_urdf(clean_urdf_filepath, output_usd)
+
+    if os.path.exists(output_usd):
```
eundersander, line 251: You should assert that the file exists. I.e. the test should fail if the file doesn't exist.

eundersander (commented):
I left a few more comments. Also, your test is failing in CI:
```
==================================== ERRORS ====================================
___________________ ERROR collecting test/test_isaac_sim.py ____________________
ImportError while importing test module '/home/runner/work/habitat-lab/habitat-lab/habitat-lab/test/test_isaac_sim.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../../miniconda3/envs/habitat/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/test_isaac_sim.py:7: in <module>
    from lxml import etree as ET
E   ModuleNotFoundError: No module named 'lxml'
```

I believe the fix here is to add lxml to `habitat-lab/requirements.txt`.

.gitignore
danieltmeta: Dependency Repos section removed.

.gitignore
danieltmeta: Removed Dependency Repos section

test/test_isaac_sim.py
danieltmeta: Fixed two of these assert statements. I don't know what I was thinking, haha.

.gitignore
danieltmeta: Qualified this .asset_hash

danieltmeta:
Oh I see, you wanted the CLI to be consolidated into a single command with all of the functions you mentioned for urdf to usd.

test/data/usd_conversion_data/config.yaml
danieltmeta: config.yaml file removed.

habitat-lab/habitat/isaac_sim/asset_conversion/urdf_to_usd.py
```diff
@@ -181,62 +181,55 @@ def add_habitat_visual_metadata_for_articulation(
...
-    parser_add_hab_visual_metadata.add_argument("reference_urdf_filepath")
-    parser_add_hab_visual_metadata.add_argument("out_usd_filepath")
-    parser_add_hab_visual_metadata.add_argument("project_root_folder")
-    parser_add_hab_visual_metadata.set_defaults(
-        func=add_habitat_visual_metadata_for_articulation
-    )
+    parser.add_argument("input_urdf_file")
+    parser.add_argument("output_usda_file")
+    parser.add_argument("project_root_folder")
+    parser.add_argument("--remove-visual")
```
eundersander, line 229: You should hook this up to your call to `convert_urdf_to_usd`

eundersander (commented):
Your test is still failing, due to Isaac not installed on the CI machine.

Some thoughts:
* Can the CI machine install Isaac? Needs to be Ubuntu 22 and have an Nvidia GPU.
* If yes, how should we install it? Since Isaac is optional for Habitat-lab, let's not add it to requirements.txt. Instead, it can be a separate pip install in the CI script.
* If no, you should configure your test to skip if Isaac isn't installed (if it can't be imported). This means CI won't run your test, but your test is still valuable; it can be run locally by folks that want to understand these conversion scripts.

habitat-lab/habitat/isaacsim/urdf_to_usd.py
danieltmeta: This was updated.

danieltmeta:
Pausing work here. Trying to run test/test_isaac_sim.py for unit testing, but it requires Isaac Sim and Issac lab to be installed on github actions CI. The github action file related to this unit test is at .github/actions/install_isaac_deps/action.yml.  Basically, ./isaac_modified.sh is copied into the IsaacLab repo, and then installed instead of isaaclab.sh.  Isaaclab.sh is not run because there is a pathing issue with the global constant ISAACLAB_PATH and extract_isaacsim_path().  The below is the error message related to missing files that somehow were not cloned into the IsaacLab repo.

https://github.com/facebookresearch/habitat-lab/actions/runs/13173253221/job/36767237376

PR #2124, opened 2024-12-20
add pure region matching to the object->rec mapping util

aclegg3:
## Motivation and Context

This PR adds additional return options for the object->Receptacle matching util. It now produces rec unique names, "floor,<region.id>" or "<region.id>". This will help with mapping objects to graph nodes in world model graph structures.

Additionally splits the 3D point -> Receptacle matching code to its own util.

Refactors the hitl use example to consider the new output.

## How Has This Been Tested

TODO, should add unit test

## Types of changes

## Checklist


zephirefaith (approved):
Thanks for this addition! Tests on counterpart PR are passing, so approving this. Let's merge that PR once this is on `main`.

PR #2132, opened 2025-01-16
Bulk usd conversion **\[Development\]**

danieltmeta:
## Types of changes

## Checklist


danieltmeta:
The current state of the code takes more than two hours to bulk convert scene instance jsons.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
...
+        )
+
+        assert True
+
+    if os.path.exists(scene_output_dir):
+        shutil.rmtree(scene_output_dir)
+        print(f"Directory '{scene_output_dir}' deleted successfully.")
+
+
+if __name__ == "__main__":
```
eundersander, line 393: This main is not needed. If you ever want to run this test directly, I believe there's a way to invoke pytest and request this specific file or test.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
...
+    )
+
+    if os.path.exists(scene_output_dir):
+        shutil.rmtree(scene_output_dir)
+        print(f"Directory '{scene_output_dir}' deleted successfully.")
+
+    if not os.path.exists(scene_output_dir):
+        os.mkdir(scene_output_dir)
+        gitignore_filepath = scene_output_dir + "/.gitignore"
+        with open(gitignore_filepath, "w"):
```
eundersander, line 365: I don't understand why you're creating a `.gitignore` file here.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
...
+            pass
+        print(f"File '{gitignore_filepath}' created successfully.")
+
+    scene_instance_json_list = find_scene_instance_json_files(scene_directory)
+
+    for scene_filepath in scene_instance_json_list:
```
eundersander, line 371: Two points:
1. If the source folder doesn't exist, you should skip the test. See `test/test_sim_utils.py`'s use of `@pytest.mark.skipif` as a reference.
2. If you're not skipping the test, you should assert that you found at least one scene to convert.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
...
+        scene_usd_filepath = make_scene_instance_usd_path(
+            scene_filepath, scene_output_dir
+        )
+
+        subprocess.run(
```
eundersander, line 376: Don't use `subprocess.run`. Just import the functions you need.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
...
+            [
+                "python",
+                scene_instance_conversion_script,
+                scene_filepath,
+                project_root_folder,
+                scene_usd_filepath,
+            ]
+        )
+
+        assert True
```
eundersander, line 386: Let's do a real test here, e.g. assert that all the scenes you found got converted, i.e. test for existence of expected files.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
                     "habitatVisual:assetScale"
                 ).Get()
             )
+
+
+def test_bulk_scene_instance_conversion():
+    project_root_folder = repo_dir
+    scene_directory = repo_dir + "/../" + "habitat-sim/data/hssd-hab/scenes"
```
eundersander, line 353: This is a weird location to look for data. Other tests assume `./data` (at the root of habitat-lab). You should follow that convention.

On your local machine, you should probably symlink or move your HSSD install from habitat-sim/data to habitat-lab/data.

test/test_isaac_sim.py
```diff
@@ -315,3 +346,49 @@ def test_add_habitat_visual_metadata_for_articulation():
                     "habitatVisual:assetScale"
                 ).Get()
             )
+
+
+def test_bulk_scene_instance_conversion():
```
eundersander, line 351: You can name this test_bulk_hssd_scene_instance_conversion, since we are specifically using HSSD.

eundersander (commented):
Some good progress! I left comments. Beware I also changed the PR base to `add_conversion_utils` to make it easier to inspect this PR's changes in isolation.

test/test_isaac_sim.py
danieltmeta: Removed.

test/test_isaac_sim.py
danieltmeta: You're right. I was afraid of a bunch of scene instance usdas being left in the repo, but makes more sense to just delete the output folder when the bulk conversion is done

test/test_isaac_sim.py
danieltmeta: `@pytest.mark.skipif` was added, try/except was added to catch if directory is empty.

test/test_isaac_sim.py
danieltmeta: renamed.

test/test_isaac_sim.py
danieltmeta: I changed the scene_directory path. I assume the user will move the hssd-hab folder into habitat-lab/data

danieltmeta:
This branch will not finish until https://github.com/facebookresearch/habitat-lab/pull/2113 is closed.

PR #2148, opened 2025-01-30
Revert "fix bug in region_center"

aclegg3:
Reverts facebookresearch/habitat-lab#2147

This change worked because an older version of habitat-sim was installed on the test platform and should be reverted.


akshararai (approved):
lgtm

PR #2176, opened 2025-04-16
Dataset Processing: Transform collada files to match Habitat coordinate system.

0mdc:
## Motivation and Context

Habitat [ignores the up direction of Collada files](https://github.com/facebookresearch/habitat-sim/blob/main/src/esp/assets/ResourceManager.cpp#L145).

```
#ifdef ESP_BUILD_ASSIMP_SUPPORT
  if (Cr::PluginManager::PluginMetadata* const assimpmetadata = importerManager_.metadata("AssimpImporter")) {
    assimpmetadata->configuration().setValue("ImportColladaIgnoreUpDirection", "true");
  }
#endif
```

Therefore, other engines cannot accurately determine the coordinate system of Habitat Collada files.

This changeset forces a Collada up vector so that third party engines can interpret the Collada coordinate systems like Habitat.

## How Has This Been Tested

Tested with Unity.

## Types of changes

- **\[Development\]**

## Checklist

### DIFF
--- a/scripts/habitat_dataset_processing/habitat_dataset_processing/asset_processor.py
+++ b/scripts/habitat_dataset_processing/habitat_dataset_processing/asset_processor.py
@@ -9,9 +9,11 @@ from pathlib import Path
 from threading import Lock
 from typing import Any
 
+import collada
 from habitat_dataset_processing import magnum_decimation
 from habitat_dataset_processing.configs import Config, Operation
 from habitat_dataset_processing.job import Job
+from habitat_dataset_processing.util import is_file_collada
 
 METADATA_FILE_VERSION = 1
 OMIT_BLACK_LIST = False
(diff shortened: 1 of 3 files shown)


aclegg3 (commented):
It makes sense to me that we would add coordinate system reading to the importer for parity with other engines before we overwrite the existing coordinate system.
Also worth noting that Habitat's Y-up is something we want to change, so any assets we convert this way would need to be reconverted later if we switch.

That said, I understand the need for parity now and this does solve the problem without modifying Habitat's behavior which is desireable for the scope of the fix. My remaining nit would be to have a non-destructive option to provide a separate output path instead of overwriting the original.

0mdc:
>Also worth noting that Habitat's Y-up is something we want to change, so any assets we convert this way would need to be reconverted later if we switch.

This fix can be removed if we read the Y-up direction of Collada files in Assimp.

>My remaining nit would be to have a non-destructive option to provide a separate output path instead of overwriting the original.

This operation is done when copying processed assets into a new directory - it ensures that the generated files are all compliant. The original file remain intact.

PR #2189, opened 2025-06-28
[bugfix] use field(default_factory) for dataclass defaults

xiaosq2000:
## Motivation and Context

This change is required to support Python 3.11 and newer. In recent Python versions, using mutable default arguments (like `dataclass` instances) directly in `dataclass` field definitions is disallowed, as it leads to all instances of the parent class sharing the same default object. This can cause unexpected behavior where modifying a configuration for one object inadvertently affects others.

This pull request fixes this issue by using `field(default_factory=...)`, which is the correct pattern for initializing mutable default values in `dataclasses`.

## How Has This Been Tested

This change applies a standard Python pattern for dataclass initialization. The correctness of this approach is established by Python's own documentation and behavior. 

## Types of changes

## Checklist


xiaosq2000:
It has been patched in main branch.

PR #2224, opened 2026-04-11
fix: use int32 bounds for semantic sensor observation space

xiaosq2000:
## Summary

- Relaxes numpy version requirement to `>=1.26.4`
- Fixes `OverflowError` in `HabitatSimSemanticSensor._get_observation_space` when running with NumPy 2.0+

The semantic sensor declared `dtype=np.int32` but used `np.iinfo(np.uint32)` for `low`/`high` bounds. NumPy <2.0 silently wrapped `uint32` max (4294967295) to `-1`; NumPy 2.0+ enforces strict integer bounds checking and raises:

```
OverflowError: Python integer 4294967295 out of bounds for int32
```

## Test plan

- [x] Run any config with `semantic_sensor` enabled under NumPy 2.0+
- [x] Verify observation space dtype is `int32` and bounds are within range

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wadeKeith (commented):
Withdrawn; this review was too cursory.

